PR #11, opened 2021-07-08
Feature/auto refresh

### LINKED ISSUE #9: Auto refresh countdown feature for Swap view
reload transfer routes after x seeconds from the server to guarantee best swapping routes at all times. 

similar to 

https://app.1inch.io/#/1/swap/USDC/DAI

image

Addminus:
Feedback wanted

closes #9

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "ethers": "^5.2.0",
     "paraswap": "^4.19.1",
     "react": "^17.0.2",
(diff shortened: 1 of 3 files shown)


maxklenk (approved):
Nice, works as expected. I just moved it inline, to avoid to much scrolling on small screens, fixed the tooltip and stoped refreshing once the swap modal opens.

PR #18, opened 2021-08-24
Deploy Connext UI

maxklenk:
Prepare new xpollinate.io
- [x] host/deploy connext.li.finance (+ xpollinate.io)
- [x] Customize UI (header, footer, background)
- [x] store configuration in URL query (partly - wait for mainnets)
- [x] redirect old xpollinate query parameters (partly - wait for mainnets)
- [x] parse big numbers
- [x] reduced UI of popup
- [x] history
- [x] hide scroll bars
- [x] closable info
- [x] embed logo text in li.finance logo
- [x] make "No route" retry
- [x] disable optimism
- [x] social media preview
- [x] account change (delete messaging token)
- [x] Add BSC testnet
- [x] fix fee calculation for history
- [x] mobile logos
- [x] favicon
- [x] host testnet version (testnet.xpollinate.io / testnet-connext.li.finance)
- [x] link testnets to other domain
- [x] Nice Info text
- [x] check if token approval is needed or not
- [x] direct xpollinate domain to new nameservers

Mainnet:
- [x] add mainnet token and chains
- [x] store configuration in URL query
- [x] redirect old xpollinate query parameters

Launch:
- [x] CNAME settings for xpollinate.io
- [x] Caching for xpollinate.io


Tommy-R8:
[Minor UI] The active and historical transaction deployable menus only work when you click on the text and not the arrow drop down icon

andyranitzsch:
Moderately important suggestion:

- [ ] Showing a warning prompt when a user decides to set the 'Receiving Address' under Advanced Options to a different one than his wallet's. This is to avoid loss of funds for inexperienced users that get confused about the labels of things, which has happened in the past. It could say something like "Warning. You are sending funds to a different address than the one you are transferring with."

Some less important UI suggestions:

- [ ] Adding Dark Mode
- [ ] Informing the user that he has to approve the token before clicking 'swap'. A good implementation of this is 1inch's approach: A yellow-glowing 'lock' indicator next to the token that only shows when the user has to approve that token (also considering the user may not want Infinite Approval).

maxklenk:
Thanks @Tommy-R8 and @andyranitzsch for your feedback. I added the tasks.
The "less important" ones are good to keep in mind, but we will for noch focus on getting the first release done and then take care of the optional ones.

PR #19, opened 2021-08-26
Separate Steps

maxklenk:
In order to make make each step interchangeable we separate them and execute them one after the other.

Backend support required: https://github.com/lifinance/lifi-backend/pull/9

ToDo:
- [x] comment in honeyswap swaps in backend
- [x] execute honeyswap swap
- [x] update status while executing swap (paraswap.execute.ts)
- [x] store received amount
- [x] execute any uniswap v2 clone swap
- [x] ensure to use the right chain before triggering the swap
- [x] parse error messages and display them in UI
- [x] allow to retry steps and contiune execution from there


maxklenk:
Closed because it is included in #25

PR #29, opened 2021-09-21
Feature: Add hop as cross-chain transfer exchange

Addminus:
- [x] backend: merge develop
- [x] on start "Amount must be greater than bonder fee"
- [x] wait/find for receiving transaction
- [x] add more chains


maxklenk:
Looks good. The code is very clean and matches our structure well. We my just split the approve and send step. So we can handle and track them separately.
It was quite difficult to test, on the testnets the right coins didn't show up and on mainnet I had to disable nxtp to find a route with hop. And the UI is still waiting for my transfer 30 minutes ago...
We could merge it and then add other chains in a second step so it can really be used.

PR #30, opened 2021-09-22
General UI enhancements

Addminus:
- Add local notifications for successful and failed transactions
  - only show/ring when not on the page
- Fix the "retrigger step" Button glitch in swapping modal
- general styling and animations
- refresh balances on successful swap
- show final balance in modal after a swap
- "no routes found" information display
- error handling
  - fixed error-parsing in nxtp process


maxklenk:
Pretty cool, looks nice.

Feedback:
- [ ] The mobile layout of the swap timeline should maybe not switch between the left and right side
- [ ] The desktop modal is very big. I think it does not have to be that tall. Also the footer section looks quite empty most often. Perhaps it would help to vertically center the loading animation to use the space
- [ ] I like that we show the new balance at the end. Perhaps also add on which chain we have this balance, add a coin icon and make the balance/coin clickable with a hint, to add the coin to metamask
- [ ] We may allow to extend the process with an optional footer message, that would allow to keep the message in the timeline short and add a longer more explaining message for the footer section
- [ ] We need a warning/info box on top of the swap to add explaination, a beta waring and that users should use metamask. Maybe you can think about a design for this, and make the image smaler...
- [ ] The "No Route Found" section could have a background so it better readable

I have to test the notifications more, so far I didn't see any.

maxklenk (changes requested):
Nice, I was now able to test the notifications in firefox.
 
I found some improvements (and fixed some already):
- [x] no hover on warning, because it is not clickable
- [x] more footer space, and centered
- [x] loading annimation for timeline icon used for uniswap, also use to cross transfer
- [ ] add token, does only work right when the user is on the right chain - check and change chain if nessesarry
- [ ] show on which chain the user has to finalBalance "You now have 33 USDC on Binance Smart Chain"

PR #37, opened 2021-09-27
Fix liquidity query

rhlsthrm:


maxklenk (approved):
Great. I removed the logging and disabled the failing rpcs as suggested.

PR #55, opened 2021-10-23
LF-50: Integrate LiFi Contract with nxtp on staging

maxklenk:
When activated by specifying supported chains in `REACT_APP_LIFI_CONTRACT_ENABLED_CHAINS_JSON`, the returned routes will be checked if they can be executed using our contract. The user then only has to submit a single transaction to perform swaps and a nxtp based transfer.
Screenshot 2021-10-23 at 8 02 51 PM
Screenshot 2021-10-23 at 8 07 07 PM
.


src/components/Swap.tsx
```diff
@@ -66,12 +67,14 @@ const Swap = ({
 
   // Options
   const [optionSlippage, setOptionSlippage] = useState<number>(3)
+  // const [optionInfiniteApproval, setOptionInfiniteApproval] = useState<boolean>(true)
+  const [optionUseLifiContract, setOptionUseLifiContract] = useState<boolean>(true)
 
   // Routes
-  const [routes, setRoutes] = useState<Array<Array<TransferStep>>>([])
+  const [routes, setRoutes] = useState<Array<Array<Step>>>([])
```
mathiasmoeller, line 74: Shouldn't we maybe use routes here?
```suggestion
  const [routes, setRoutes] = useState<Array<Route>>([])
```

src/components/Swap.tsx
```diff
@@ -66,12 +67,14 @@ const Swap = ({
...
   const [routesLoading, setRoutesLoading] = useState<boolean>(false)
   const [noRoutesAvailable, setNoRoutesAvailable] = useState<boolean>(false)
-  const [selectedRoute, setselectedRoute] = useState<Array<TransferStep>>([])
+  const [selectedRoute, setselectedRoute] = useState<Array<Step>>([])
```
mathiasmoeller, line 77: And here
```suggestion
  const [selectedRoute, setselectedRoute] = useState<Route>([])
```

src/components/Swap.tsx
```diff
@@ -209,7 +212,8 @@ const Swap = ({
         try {
           const result = await axios.post<any>(process.env.REACT_APP_API_URL + 'transfer', { deposit, withdraw }, config)
           // filter if needed
-          const routes: Array<Array<TransferStep>> = result.data
+          const routesRaw: Array<Array<Step>> = result.data
```
mathiasmoeller, line 215: I mean everywhere where we have a list of steps 😁

src/components/Swapping.tsx
```diff
@@ -314,49 +317,37 @@ const Swapping = ({ route, updateRoute, onSwapDone }: SwappingProps) => {
     setIsSwapping(true)
   }
 
-  const getRecevingInfo = (step: TransferStep) => {
-    let toChain: Chain, toToken: Token;
-    switch(step.action.type) {
+  const getRecevingInfo = (step: Step) => {
+    let toChain: Chain, toToken: Token
+    switch (step.action.type) {
```
mathiasmoeller, line 322: This is missing the default case now, isn't it?

src/services/1Inch.ts
```diff
@@ -93,6 +93,8 @@ const getTransaction = async (chainId: number, fromTokenAddress: string, toToken
   const path: Array<any> = result.data.protocols ? result.data.protocols[0].map((step: Array<any>) => step[0]) : []
   const tx = result.data.tx
 
+  console.info('1inch data', result.data)
```
mathiasmoeller, line 96: ✂️

src/services/nxtp.tsx
```diff
@@ -75,7 +75,7 @@ export const getTransferQuote = async (
   initiator?: string,
 ): Promise<AuctionResponse | undefined> => {
   // Create txid
-  const transactionId = getRandomBytes32();
+  const transactionId = getRandomBytes32()
```
mathiasmoeller, line 78: Maybe we use that for our step ids as well?

maxklenk: Yes, that would be good.

src/components/Swap.tsx
maxklenk: This PR has been build before we implemented Routes.

maxklenk:
Merged in new implementation

PR #63, opened 2021-11-04
LF-44: get balances from chain instead of debank

maxklenk:
The DeBank API token balances with too many decimals, which cases errors if users try to use all the tokens we think they have. We have now integrated multicall, which allows us to query balances directly from chain. I added this as the default way to get balances for both main- and testnets. If multicall is not available on a chain we instead fall back to query the balances of every token in a single call. The DeBank API is still used for the dashboard since we aim to show all the token a user have, in the swap we are only interested in token the user can actually swap or transfer (based on the dexs token lists).


Addminus (approved):
Top!

mathiasmoeller (approved):

PR #64, opened 2021-11-04
LF-70: do not try to add native gas token to wallet

maxklenk:
After the users did a transfer or swap we show him his new token balance. When hovering the token name the user can add this token to his metamask wallet, because the wallet does not automatically show all tokens a user has. The native gas token on the other hand is always shown at the very top in the wallet, trying to add it to the wallet is not needed and doesn't work. So we should not show this option.


Addminus (approved):
TOP

mathiasmoeller (approved):

PR #67, opened 2021-11-08
Lf-86: general resume fixes

Addminus:
- FIX state display in Tx-Table
- FIX Active TX-Collapse
- FIX wait for nxtp fulfilltransaction 
- FIX overwriting execution on resume
- FIX toChain display error In Transaction Table
- ADD Table pagination for Tx-Table
- ADD Remove Tx-Button for active transaction table


maxklenk (changes requested):
Nice, works very well now. The problems I previously had have been fixed.
While testing I was able to find and reproduce two problems:

1.  NXTP send transaction:
If I have a route with nxtp transfer and pause the process after the quote was confirmed. When I resume the graphs are not synced yet, similar to the problem we had with getQuote before, but in this case sentTransaction is the first thing we do.

2. NXTP Sign Message to Claim Funds
I start a nxtp transfer, sent the transaction and then leave/reload the page. I wait a few minutes until the receiver should be prepared. When resuming the process now, I can sign the  claim but the UI stops in "Sign Message to Claim Funds" and ever shows the next step.

maxklenk (approved):
Awesome. Works great now.

PR #71, opened 2021-11-09
feat(@lifinance/types): added

michaelpautov:


maxklenk (approved):
Great, good idea to just replace the reference in `src/types/index.ts` instead of updating it everywhere in the project. Since our `@lifinance/types` package does not only contain types used by typescript while developing, but also list of chains, coins, exchanges which we access when running the application I think it would be better to store it in `dependencies` instead of `devDependencies`.

PR #74, opened 2021-11-10
[DNM] Explorer integration

jakekidd:
- removed Historical Transactions
- added Transaction History button (takes you to user's tx history on explorer
- removed Liquidity (explorer shows this now)
- added Explorer button (kinda redundant)
- Split the interface into two columns, one with Active Transactions, one with the sending interface
     - on mobile, they go vertical arrangement, and active txs is at the top
- no longer using a table to display tx info, instead just use these card components. looks a bit basic, could be cleaned up, but imo is a bit more readable than that table

Left TODO:
- [x] split up fees and impact
- [ ] newly added buttons get scrunched up weird on mobile: should use a hamburger menu, possibly a second bar (footer)?
- [ ] have ActiveTransactions be drawer that pops out
- [ ] fix ActiveTransactions and Crosschain Swap placement
     - [x] former on right, latter on left on desktop
     - [ ] Use flexbox to have ActiveTransactions be at the top on mobile
- [ ] handle receiving amount being incorrect; should come from the bid?
    - from Max:
        > It works later in the process as you see when "Sign to claim" is shown. I think tx.txData.receiving.amount  is not filled in the beginning. You need to parse the bid to add the information. 
- [ ] Add some horizontal space between the header buttons on mobile
- [ ] on mobile there is some strange padding/margin on the right side
- [ ] set a width limit on the form (again) and give more space for the active transactions


sanchaymittal:
leaving the request for UI hamburger Menu on mobile interface to lifi team!

PR #79, opened 2021-11-11
LF-58: SDK wrap backend calls

Addminus:
- ADD LIFI class
- ADD `findRoutes(deposit, withdraw)` method
- ADD tests for `findRoutes(deposit, withdraw)`
- MODIFY `Swap.tsx` to work with the additions mentioned above

### DIFF
--- a/src/components/Swap.tsx
+++ b/src/components/Swap.tsx
@@ -24,6 +24,7 @@ import {
   defaultTokens,
   DepositAction,
   getChainByKey,
+  Step,
   Token,
   TransferStep,
   WithdrawAction,
@@ -95,7 +96,7 @@ const Swap = ({ transferChains }: SwapProps) => {
   const [balances, setBalances] = useState<{ [ChainKey: string]: Array<ChainPortfolio> }>()
   const [refreshBalances, setRefreshBalances] = useState<boolean>(true)
   const [currentRouteCallId, setCurrentRouteCallId] = useState<string>()
-  const [routeCallResult, setRouteCallResult] = useState<any>()
+  const [routeCallResult, setRouteCallResult] = useState<{ result: Step[][]; id: string }>()
 
   // Options
   const [optionSlippage, setOptionSlippage] = useState<number>(3)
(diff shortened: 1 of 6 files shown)


src/services/LIFI/Lifi.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+    if (!isWithdraw(withdraw)) {
+      throw Error('Invalid Withdraw Type')
+    }
+    const result = await axios.post<any>(process.env.REACT_APP_API_URL + 'transfer', {
+      deposit,
+      withdraw,
+    })
+    return result.data as Step[][]
+  }
```
maxklenk, line 19: We don't need to use `<any>` here if we know what result we expect. You can define `Step[][]` instead.

src/components/Swap.tsx
```diff
@@ -239,38 +244,14 @@ const Swap = ({ transferChains }: SwapProps) => {
...
-          setRoutes(routes)
-          fadeInAnimation(routeCards)
-          setHighlightedIndex(routes.length === 0 ? -1 : 0)
-          setNoRoutesAvailable(routes.length === 0)
+          const id = uuid()
+          setCurrentRouteCallId(id)
+          const result = await LIFI.findRoutes(deposit, withdraw)
+          setRouteCallResult({ result, id })
+        } catch {
+          setNoRoutesAvailable(true)
```
maxklenk, line 253: We should check if the catched error here matches the `currentRouteCallId´, because if not we can ignore it and should not overwrite a perhaps valid result of the current call.

src/components/Swap.tsx
```diff
@@ -92,6 +94,8 @@ const Swap = ({ transferChains }: SwapProps) => {
   const [refreshTokens, setRefreshTokens] = useState<boolean>(true)
   const [balances, setBalances] = useState<{ [ChainKey: string]: Array<ChainPortfolio> }>()
   const [refreshBalances, setRefreshBalances] = useState<boolean>(true)
+  const [currentRouteCallId, setCurrentRouteCallId] = useState<string>()
+  const [routeCallResult, setRouteCallResult] = useState<any>()
```
maxklenk, line 98: Please avoid `any` types, I think we know what will be in here.

maxklenk (changes requested):

maxklenk (approved):
Well done.

PR #86, opened 2021-11-12
LF-94: Use new routes endpoint to fetch possible routes

mathiasmoeller:
[LF-94](https://lifi.atlassian.net/jira/software/projects/LF/boards/1?selectedIssue=LF-94)

We now use new endpoint to fetch possible routes. This required _some_ refactoring since the shape of the returned types changed.


maxklenk (approved):
ready to merge

PR #91, opened 2021-11-16
Add a warning alert when subgraphs are out of sync.

jakekidd:


jakekidd:
pls note this has changes from #88

PR #96, opened 2021-11-16
Add support for opt tokens

jakekidd:


maxklenk:
tokens will be added when optimism contracts are deployed

PR #100, opened 2021-11-17
LF-100: Use getStepTransaction endpoint to fetch swap transactions

mathiasmoeller:
[LF-100](https://lifi.atlassian.net/jira/software/projects/LF/boards/1?selectedIssue=LF-100)


src/services/1inch.execute.tsx
```diff
@@ -15,28 +16,19 @@ export class OneInchExecutionManager {
...
+    const { status, update } = initStatus(updateStatus, execution)
     if (!this.shouldContinue) return status
     if (action.fromToken.id !== constants.AddressZero) {
       const contractAddress = await oneInch.getContractAddress()
       await checkAllowance(
         signer,
         fromChain,
         action.fromToken,
-        srcAmount.toString(),
+        action.fromAmount,
```
maxklenk, line 31: the estimate contains an `approvalAddress` address you can use instead of `contractAddress` here.

src/services/1inch.execute.tsx
```diff
@@ -55,17 +47,9 @@ export class OneInchExecutionManager {
...
-          action.toToken.id,
-          srcAmount.toString(),
-          userAddress,
-          destAddress,
-          action.slippage,
-        )
-        tx = await signer.sendTransaction(call)
+        const personalizedStep = await personalizeStep(signer, step)
+        const { tx: transaction } = await Lifi.getStepTransaction(personalizedStep)
+        tx = await signer.sendTransaction(transaction)
```
maxklenk, line 52: I know this was one done in one process before. But since `getStepTransaction` is a backend call, where the user has to wait, and `sendTransaction` requires the user to confirm a message in his wallet I would split this up. So we first get the transaction (`swapProcess.status = 'PENDING'`) and then ask the user to sign the message (`swapProcess.status = 'ACTION_REQUIRED'`)

src/services/LIFI/typeguards.ts
```diff
@@ -17,6 +27,68 @@ export const isRoutesRequest = (routesRequest: RoutesRequest): routesRequest is
...
-  return !routeOptions?.slippage || typeof routeOptions.slippage === 'number'
+const isRoutesOptions = (routeOptions: RouteOptions): routeOptions is RouteOptions =>
+  !routeOptions?.slippage || typeof routeOptions.slippage === 'number'
+
+export const isStep = (step: Step): step is Step => {
+  const { id, type, tool, action, estimate } = step
+
+  return (
+    typeof id === 'string' &&
+    ['swap', 'cross', 'lifi'].includes(type) &&
```
maxklenk, line 38: I think this should reference the possible value of `StepType`

src/services/utils.ts
```diff
@@ -41,6 +41,23 @@ export const sleep = (mills: number) => {
...
+  const toAddress = step.action.toAddress || address
+
+  return {
+    ...step,
+    action: {
+      fromAddress,
+      toAddress,
+      ...step.action,
+    },
+  }
```
maxklenk, line 58: `...step.action` will overwrite `fromAddress` and `toAddress` because of the order in the new object. It should work here if `fromAddress` and `toAddress` are not present in `action` before, but if the are defined with a falsey value, they would overwrite the addresse we want to set. So we should change the order.

src/services/uniswaps.execute.tsx
```diff
@@ -55,8 +46,9 @@ export class UniswapExecutionManager {
       if (swapProcess.txHash) {
         tx = await signer.provider.getTransaction(swapProcess.txHash)
       } else {
-        const call = await uniswap.getSwapCall(action, estimate, srcAddress, destAddress)
-        tx = await signer.sendTransaction(call)
+        const personalizedStep = await personalizeStep(signer, step)
+        const { tx: transaction } = await Lifi.getStepTransaction(personalizedStep)
+        tx = await signer.sendTransaction(transaction)
```
maxklenk, line 51: We should have the same handling as in 1inch here: update the status in between `getStepTransaction` and `sendTransaction`.

maxklenk (changes requested):
Our changes are aiming to have the same handling for all swap steps. I think it would be easy to merge uniswap/paraswap/1inch.execute into one swap.execute instead. Only the `parseReceipt(tx, receipt)` call should differ (handle this in the frontend for now, but we should consider moving it to the backend as well)

src/services/utils.ts
mathiasmoeller: äh ja totally right. The spread has to come first of course 👍

maxklenk (approved):
Good to go.
I just added a few comments, so I am able to understand the code better and changed the displayed messages, because I would like them to fit in one line and the tool-names are not that nice formatted.

PR #102, opened 2021-11-18
🐕  Fix husky configuration

mathiasmoeller:
To enable husky you have to run `yarn install` once. There is a `prepare` step which will set it up.

**Important**: This works only for yarn < v2. For yarn v2 we need to do some changes

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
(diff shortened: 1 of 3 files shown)


maxklenk (approved):
Works and looks good. I just changed, that the husky pre-push script calls `yarn pre-push` to follow the same schema as pre-commit.

mathiasmoeller:
I guess we want the same for backend then as well?

PR #103, opened 2021-11-18
LF-108: Disable MAX amount if no deposit token selected

mathiasmoeller:
[LF-108](https://lifi.atlassian.net/jira/software/projects/LF/boards/1?selectedIssue=LF-108)


maxklenk (changes requested):
Please also check if a balance is loaded or not, I was able to create some errors by selecting a token quickly and then pressing MAX.

maxklenk:
I will merge the fix already. We can add the balance check in a second one.

PR #111, opened 2021-11-19
LF-107: Add new nxtp executor that uses getStepTransaction endpoint

mathiasmoeller:
[LF-107](https://lifi.atlassian.net/jira/software/projects/LF/boards/1?selectedIssue=LF-107)

# TODO:
- [x] Implement continue logic in `nxtp_new.execute.ts`
- [x] add cbridge icon

Later
- [x] Check which parts of the old implementation are used by Xpollinate and remove the rest


src/components/Swapping.tsx
```diff
@@ -397,6 +419,9 @@ const Swapping = ({ route, updateRoute, onSwapDone }: SwappingProps) => {
           case 'horizon':
             avatar = horizonAvatar
             break
+          case 'cbridge1':
+            avatar = 'cBridge1'
```
mathiasmoeller, line 423: Do we have a proper avatar for this?

src/components/Swapping.tsx
```diff
@@ -526,7 +566,7 @@ const Swapping = ({ route, updateRoute, onSwapDone }: SwappingProps) => {
       // lifi supported?
       if (isLifiSupported(route)) {
         if (!steps[0].execution) {
-          triggerLifi()
+          // triggerLifi()
```
mathiasmoeller, line 569: Is this on purpose?

src/services/hop_new.execute.ts
```diff
@@ -0,0 +1,126 @@
+import { TransactionResponse } from '@ethersproject/abstract-provider'
```
mathiasmoeller, line 1: Can't we replace the old hop implementation? (`hop.execute.ts`)

src/services/cbridge1.execute.ts
```diff
@@ -0,0 +1,121 @@
...
+    const personalizedStep = await personalizeStep(signer, step)
+    const { tx: transactionRequest } = await Lifi.getStepTransaction(personalizedStep)
+
+    // STEP 3: Send Transaction ///////////////////////////////////////////////
+    crossProcess.status = 'ACTION_REQUIRED'
+    crossProcess.message = 'Sign Transaction'
+    update(status)
+    if (!this.shouldContinue) return status
+
+    const tx = await signer.sendTransaction(transactionRequest)
```
mathiasmoeller, line 52: Here the check if the transaction was already sent (`crossProcess.txHash` exists) is missing, right?

Addminus (approved):

PR #134, opened 2021-12-01
Update Connext SDK

rhlsthrm:


maxklenk:
Please do not merge before we have updated our backend!

Is the sdk update important? I would like to test our change on staging for a moment.

rhlsthrm:
> Please do not merge before we have updated our backend!
> 
> Is the sdk update important? I would like to test our change on staging for a moment.

It has some fixes we want. We have had our users test on staging for a while. Did you test your front end yet?

maxklenk (approved):
Backend is deployed. Ready to merge

PR #141, opened 2021-12-06
Lf 95 sdk execution files

Addminus:
TODOs:

- [x] stop execution
- [x] resume execution
- [x] handling chain switches


src/components/Swapping.tsx
```diff
@@ -330,33 +330,17 @@ const Swapping = ({ route, updateRoute, onSwapDone }: SwappingProps) => {
...
-      }
-      resumeExecution()
-    }
-  }
-
   const startCrossChainSwap = async () => {
     if (!web3.account || !web3.library) return
+    const settings = {
+      updateCallback: updateCallback,
+      switchChainHook: switchChainHook,
```
mathiasmoeller, line 337: this is used before it was defined, right?

src/components/Swapping.tsx
```diff
@@ -403,6 +391,23 @@ const Swapping = ({ route, updateRoute, onSwapDone }: SwappingProps) => {
     resumeExecution()
   }
 
+  const switchChainHook = async (requiredChainId: number) => {
+    if (!web3.account || !web3.library) return
+    // check if execution stopped because of a required chain switch
+    const chainSwitchStep = steps.find((step) => step.action.fromChainId === requiredChainId)
```
mathiasmoeller, line 397: Do we have to switch the chain to claim funds for an nxtp transaction or does that work w/o changing?

src/components/Swapping.tsx
```diff
@@ -457,6 +462,13 @@ const Swapping = ({ route, updateRoute, onSwapDone }: SwappingProps) => {
       )
     }
 
+    const chainSwitchRequired = steps.some(
+      (step) => step.execution?.status === 'CHAIN_SWITCH_REQUIRED',
```
mathiasmoeller, line 466: Should we maybe switch to enums for the status types so we don't have the ugly strings lying around here? Would also be nicer for sdk users if they have proper autocompletion I would say

mathiasmoeller:
Closed in favor of #143

PR #153, opened 2021-12-16
feat: pass rpcs to LiFi SDK

mathiasmoeller:
- feat: pass rpcs and multicallAddresses to LiFi SDK
- refactor: extract avatar definitions
- refactor: replace types imports with sdk imports


src/components/Swap.tsx
```diff
@@ -323,6 +324,10 @@ const Swap = ({ transferChains }: SwapProps) => {
 
   useEffect(() => {
     const load = async () => {
+      LiFi.setConfig({
```
maxklenk, line 327: Please also set the api uri based on the environment the react app is deployed in.

maxklenk (changes requested):
Looks good, just add the api Uri please

maxklenk (approved):
Please release and add the sdk before merging

PR #165, opened 2022-01-03
feat(SwapForm): show slippage tooltip at "to" amount

mathiasmoeller:
Users are often confused why the final amount differs from the one shown initially. For this reason we should provide a clear message telling them that this is just an estimation.

image

image

### DIFF
--- a/src/components/SwapForm.tsx
+++ b/src/components/SwapForm.tsx
@@ -26,6 +26,7 @@ interface SwapFormProps {
   withdrawAmount: BigNumber
   setWithdrawAmount: Function
   estimatedWithdrawAmount: string
+  estimatedMinWithdrawAmount?: string
 
   transferChains: Array<Chain>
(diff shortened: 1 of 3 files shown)


src/components/SwapForm.tsx
```diff
@@ -294,6 +296,13 @@ const SwapForm = ({
               bordered={false}
               disabled
             />
+            {estimatedMinWithdrawAmount && (
```
Addminus, line 299: Be careful with short circuit operators. Maybe we should explicitly cast the expression like this `{!!condition && <Component />}`. We had issues with some of these code snippets displaying `0` if the condition was not met.
If `estimatedMinWithdrawAmount` can be casted to `string` or `number` this might happen in this case as well.

refer to: https://stackoverflow.com/a/53048160

Addminus (changes requested):

Addminus (approved):

PR #168, opened 2022-01-05
fix(SwapForm): render BigNumbers with value 0 as 0.0

mathiasmoeller:
amounts of type BigNumber that have a value of 0 should be rendered as 0.0 instead of adding decimal places until at least two non-zero digets are shown


maxklenk (changes requested):
The case that a string `'0'` is passed is not handled yet.

Also to avoid the instanceof checks, it would add a `floated.isZero()` check after parsing the string. This way we would first handle undefined, then string, and then BigNumbers.

maxklenk (approved):

PR #178, opened 2022-01-13
fix: handle behaviour when using site with unsupported chains

Addminus:


src/components/web3/connectors.ts
```diff
@@ -71,6 +71,19 @@ export const injected = new InjectedConnector({
...
 })
 
+// get our standard supported chain and try to append the possibly unknown chain the user is on
+export const getInjectedConnector = async () => {
+  const { ethereum } = window as any
+  const currentProvider = new providers.Web3Provider(ethereum)
+  const chainId = (await currentProvider.getNetwork()).chainId
+  // append the current chain to the supported chains.
+  // can push duplicate ids, when user is on supported chain but that does not seem to make any problems.
+  const chains = [...supportedChains.map((chain) => chain.id), chainId]
```
mathiasmoeller, line 81: This would make it unique
```suggestion
  const chains = Array.from(new Set([...supportedChains.map((chain) => chain.id), chainId]))
```

mathiasmoeller (approved):
Seems good to me 👍 
You could make the chainIds unique by using a `Set`. But if it's working already it's not a must I guess

PR #179, opened 2022-01-17
LF-253: iframe integration script

chybisov:
iframe integration script gets new functionality including drawer mode 💥 

image
image

### DIFF
--- a/public/widget.js
+++ b/public/widget.js
@@ -1,0 +1,225 @@
+(function () {
+  const host = 'https://li.finance' // 'https://li.finance';
+  let config = {
+    elementId: 'lifi-widget',
+    mode: 'drawer', // default | drawer,
+    buttonText: 'Li.Fi Swap',
+    options: {
+      fromChain: 'eth',
+      fromToken: null,
+      toChain: null,
+      toToken: null,
+      fromAmount: null
+    },
+  };
+  const drawerSettings = {
+    activeClass: 'is-active',
+    visibleClass: 'is-visible',
+    targetAttribute: 'data-drawer-target',
+    triggerAttribute: 'data-drawer-trigger',
+    closeAttribute: 'data-drawer-close',
+  }
+
+  const attributeToSelector = (str) => `[${str}]`;
+
+  const globalObject = window.lifi;
+  const queue = globalObject.q;
+  if (queue) {
+    for (let i = 0; i < queue.length; i++) {
+      if (queue[i][0].toLowerCase() === 'init') {
+        config = { ...config, ...queue[i][1] };
+        init();
+      }
+      else {
+        // handle other commands
+      }
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


public/widget.js
```diff
@@ -0,0 +1,222 @@
+(function () {
+  const host = 'https://li.finance'
```
mathiasmoeller, line 2: Do we want to make this configurable as well to allow e.g. local testing or staging?

public/widget.js
```diff
@@ -0,0 +1,222 @@
...
+      toToken: null,
+      fromAmount: null
+    },
+  };
+  const drawerSettings = {
+    activeClass: 'is-active',
+    visibleClass: 'is-visible',
+    targetAttribute: 'data-drawer-target',
+    triggerAttribute: 'data-drawer-trigger',
+    closeAttribute: 'data-drawer-close',
```
mathiasmoeller, line 20: The drawer gets injected directly, and is not inside an iframe right? Could general class names like `is-active` clash with user classes?

mathiasmoeller (commented):
Looks good!

I guess we want to have this in JavaScript so we don't have to compile the TS?

The possible routes below the swap field and the resulting horizontal scrolling isn't the prettiest thing I know 😅  What's the plan here? Do we wait for the new design before we touch anything here? Could it be a quick fix to show the possible routes below each other?

chybisov:
@mathiasmoeller Yeah, TS is not used there mainly because of how the current CRA config works. It can't compile files that are not in the /src folder. I'll figure this out for redesigned widget, not satisfied with that as well. This js file is also not minimized and obfuscated, which I plan to fix in a new one.

Regarding the quick fix for horizontal scroll, I agree it doesn't look lovely, we definitely can fix this, but I'm not sure we should spend time on this now. What do you think?

public/widget.js
chybisov: The same issue as with TS we don't have access to env variables here. If you have any suggestions, I would be happy to change this. 😄

public/widget.js
chybisov: Nice catch! 💪  I'll fix this.

public/widget.js
mathiasmoeller: Hmm yeah I get it. Would it make sense to pass it as a configuration parameter? I guess "users" should always use the production version so it doesn't make too much sense?
Let's just leave it for now I think.

mathiasmoeller (approved):
Yeah let's get a first version of the Widget released. We can still refine it in future iterations. Especially since the whole frontend part gets rebuild it probably doesn't make sense to spend too much time on that now 👍 
In the new setup we can also take into account proper compiling of the TS code and minification etc.

public/widget.js
chybisov: Yep, agree. Let's move forward.

PR #182, opened 2022-01-19
fix: replace import of types package with sdk

mathiasmoeller:
We implicitly still had a dependency to the types package which can cause versioning issues. 
Since the SDK exports all important types we should only use it.
To do this all duplicated types had to be removed.


maxklenk (approved):
Looks fine.

PR #183, opened 2022-01-19
feat: disconnect button

Addminus:


src/components/web3/DisconnectButton.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </>
+    )
+  } else {
+    return (
+      <>
+        <Popover placement="bottom" content={infoContent} trigger="click">
+          <Button style={style} shape="round" danger icon={<DisconnectOutlined />}>
+            Deactivate Wallet
+          </Button>
+        </Popover>
```
maxklenk, line 93: It would be good to show which wallet is connected. A nice addition would be to show an ens name instead of some of the wallet digest if available.

src/App.tsx
```diff
@@ -140,6 +141,9 @@ function App() {
                   {/* Links */}
                   <Col xs={0} sm={0} md={10} lg={10} xl={10} style={{ paddingRight: 10 }}>
                     <Row justify="end" gutter={15}>
+                      <Col>
+                        <WalletButtons></WalletButtons>
```
maxklenk, line 145: It would be nice to style the Buttons in the header similar to the Support Button. Perhaps even move the wallet button to the very right, so that we have the buttons next to each other.

maxklenk (changes requested):
Nice, I also had a look at the UI and have added some suggestions for it. Other then that it is working great.

If no wallet is connected we make some requests which fail. This was the case before already, but perhaps you have an idea why this happens.
`index.ts:56 POST https://speedy-nodes-nyc.moralis.io/5ed6053dc39eba789ff466c9/eth/mainnet 415`

src/App.tsx
Addminus: good idea. will do!

src/components/web3/ConnectButton.tsx
```diff
@@ -1,17 +1,55 @@
...
-function ConnectButton() {
+const addToActiveWallets = (address: string | null | undefined) => {
+  if (!address) return
+  const lowerCaseAddress = address.toLowerCase()
+  const activeWallets = readWallets()
+  activeWallets.push(lowerCaseAddress)
+  storeWallets(activeWallets)
+}
+
+const removeFromDeativatedWallets = (address: string | null | undefined) => {
```
mathiasmoeller, line 21:
```suggestion
const removeFromDeactivatedWallets = (address: string | null | undefined) => {
```

Webstorm has a really great spell checker... even validates grammar.... just sayin'....

src/services/localStorage.unit.spec.ts
```diff
@@ -1,10 +1,112 @@
...
 const SOME_NEWER_DATE = new Date('2021-04-10').getTime()
 const SOME_OLDER_DATE = new Date('2020-04-10').getTime()
 
+// Object.setPrototypeOf(localStorage.__proto__.setItem, jest.fn())
+// Object.setPrototypeOf(localStorage.__proto__.getItem, jest.fn())
+// Object.setPrototypeOf(localStorage.__proto__.removeItem, jest.fn())
+
+// Storage.prototype.setItem = jest.fn()
+// Storage.prototype.getItem = jest.fn()
+// Storage.prototype.removeItem = jest.fn()
```
mathiasmoeller, line 19: This can be removed I guess?

mathiasmoeller (approved):
Apart from the 2 small things it looks good 👍 
Let's not forget to tell the twitter peepz to create a post about how nobody does this correctly APART FROM US

PR #184, opened 2022-01-21
feat: check for execution status on page load

Addminus:
contains necessary logic to get new status on page load. 

Needs SDK PR: https://github.com/lifinance/sdk/pull/26

If SDK is not merged and released, no behaviour will change.

### DIFF
--- a/src/components/Swap.tsx
+++ b/src/components/Swap.tsx
@@ -28,7 +28,12 @@ import React, { useCallback, useEffect, useRef, useState } from 'react'
 import { v4 as uuid } from 'uuid'
 
 import { getRpcs } from '../config/connectors'
-import { deleteRoute, readActiveRoutes, readHistoricalRoutes } from '../services/localStorage'
+import {
+  deleteRoute,
+  readActiveRoutes,
+  readHistoricalRoutes,
+  storeRoute,
+} from '../services/localStorage'
 import { switchChain } from '../services/metamask'
 import { loadTokenListAsTokens } from '../services/tokenListService'
 import { deepClone, formatTokenAmount, formatTokenAmountOnly } from '../services/utils'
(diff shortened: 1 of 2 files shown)


src/components/Swap.tsx
```diff
@@ -322,6 +327,22 @@ const Swap = ({ transferChains }: SwapProps) => {
...
+    // get new execution status on page load
+    activeRoutes.map((route) => {
+      if (!web3 || !web3.library) return
+      const settings = {
+        updateCallback: (updatedRoute: RouteType) => {
+          storeRoute(updatedRoute)
+          setActiveRoutes(readActiveRoutes())
+          setHistoricalRoutes(readHistoricalRoutes())
+        },
+      }
```
mathiasmoeller, line 340: I guess the "old" update callback that we passed when we initially started the route doesn't exist anymore after a page load and that's why we need to pass a new one?

src/components/Swap.tsx
```diff
@@ -322,6 +327,22 @@ const Swap = ({ transferChains }: SwapProps) => {
...
+      LiFi.resumeRoute(web3.library.getSigner(), route, settings)
+      LiFi.moveExecutionToBackground(route)
+    })
+  }, [web3.library])
```
mathiasmoeller, line 344: This will be executed everytime the `web3.library` changes, do we know when this happens? Does this happen on chainswitch or smth like that? 
Is it a problem that it could happen that `resumeRoute` is triggered again and again?

src/components/Swap.tsx
Addminus: Yes exactly. The SDK is not saving anything anywhere and starts blank!
In general we should not access browser specific stuff (storage and other apis) in the sdk as it could be used on servers and cmd-lines as well

src/components/Swap.tsx
Addminus: Had this thought as well. `web3.library` changes everytime a new library gets loaded. So on page load on connect / disconnect and on change switch. And yes this means, that this sideEffect will be called multiple times. But this is not an issue because:

- Resume Route checks if the route has been halted (moved to bg or stopped). If it has not been halted, it will not execute.
  - Triggering an execution multiple times does no harm
- Triggering this method after a chain switch is a actually good thing:
   - If the step only has one route, nothing really happens. The route will be marked as DONE if done and moved to historical TXs (assuming all txs have been sent)
   - If the routes has multiple steps: the retrigger will alert the user that a chain switch is required, as the wrong signer for the tx is passed. This is a nice clue, that the execution is not execution in the background anymore and the reason for it.

mathiasmoeller (approved):
👍

PR #203, opened 2022-02-08
LF-266: wallet connect support

Addminus:


src/components/Swap.tsx
```diff
@@ -737,6 +744,17 @@ const Swap = () => {
...
+    if (!web3.account) {
+      return
+    }
+    if (isWalletConnectWallet(web3.account)) {
+      setShowWalletConnectChainSwitchModal({ show: true, chainId })
+      return
+    }
+    await switchChainMetaMask(chainId)
+  }
+
```
mathiasmoeller, line 757: For the widget we should probably extract this kind of logic into wallet specific handler files for better maintainability. But I guess for the "old frontend" it doesn't matter that much 🙈

mathiasmoeller (approved):

PR #215, opened 2022-02-14
fix: sync custom tokens with server

maxklenk:
When adding custom token via a url parameter, they should be loaded from the server.

### DIFF
--- a/src/components/Swap.tsx
+++ b/src/components/Swap.tsx
@@ -420,12 +420,48 @@ const Swap = ({ transferChains }: SwapProps) => {
         if (!newTokens[chain.key]) newTokens[chain.key] = []
         newTokens[chain.key].push(token)
       })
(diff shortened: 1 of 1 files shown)


src/components/Swap.tsx
```diff
@@ -420,13 +420,49 @@ const Swap = ({ transferChains }: SwapProps) => {
...
+        const index = tokens[chain.key].findIndex((token) => token.address === updatedToken.address)
+        if (index === -1) {
+          tokens[chain.key].push(updatedToken)
+        } else {
+          tokens[chain.key][index] = updatedToken
+        }
+        return tokens
+      })
+    })
+  }
```
mathiasmoeller, line 464: I guess this could be wrapped in a `useCallback` hook to get a cached version of it that gets only called if `token` really changed. But this is probably not really a great performance boost since this method doesn't get called very often.

mathiasmoeller (approved):

PR #225, opened 2022-02-23
feat: use sdk with updated bridging behaviour

mathiasmoeller:
- chore: update sdk version
- fix: fix dependencies to sdk dependencies


Addminus (approved):
Phew! lets hop this new SDK stuff works out!

PR #251, opened 2022-03-14
fix: show actually received amount and token

mathiasmoeller:
The actually received amount differs from the estimated amount. We should show the received amount if possible.
Also the received token differs in case the destination swap fails and the user receives the bridged token.

(Requires a [sdk release](https://github.com/lifinance/sdk/pull/45))


Addminus:
Please wait for #252
https://github.com/lifinance/lifi-web/pull/252

Addminus (approved):

PR #269, opened 2022-04-06
feat: show token validation badge

maxklenk:
Make use of Go+ to show token verification:
Screenshot 2022-04-06 at 2 33 10 PM

Risky token sample:
https://gopluslabs.io/token-security/56/0x373233a38ae21cf0c4f9de11570e7d5aa6824a1e

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -54,7 +54,7 @@ function App() {
 
   function embedView() {
     setMetatags({
-      title: 'Li.Finance - Swap (embed)',
+      title: 'LI.FI - Swap',
     })
     return (
       <div className="lifiEmbed">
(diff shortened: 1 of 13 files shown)


mathiasmoeller (approved):
Looks good except this one little eslint issue that github shows.

Do we want to add the to the tokenservice in the backend on the long run? All API users would benefit from having this information about tokens and caching etc would be easy.

maxklenk:
> Do we want to add the to the tokenservice in the backend on the long run? All API users would benefit from having this information about tokens and caching etc would be easy.

Yes, would be great to make that part of the API later, but first I want to evaluate which are the important information we want to include and the service only supports half of the chains we currently use.

Addminus (approved):

PR #271, opened 2022-04-08
Small fixes

maxklenk:


Addminus (commented):
Would it make sense to remove the about page?
We are not using it anymore, or do we have some future usecases for it?

Addminus (approved):

PR #275, opened 2022-04-15
feat: crosschain klima staking

Addminus:


src/components/SwapEtherspotKlimaZap.tsx
```diff
@@ -0,0 +1,1189 @@
...
+        // eslint-disable-next-line
+        console.warn('possibilities request did not contain required setup information')
+        return
+      }
+
+      // chains
+      // FIXME: limit available chains to etherspot supported ones
+      const chains = possibilities.chains.filter((chain) => {
+        return etherspotSupportedChains.includes(chain.id)
+      })
```
mathiasmoeller, line 442: Is the fixme still needed? Seems like it is already filtered

src/components/SwapEtherspotKlimaZap.tsx
```diff
@@ -0,0 +1,1189 @@
...
+    const token = findDefaultToken(crossChain.coin, crossChain.id)
+    const balance = getBalance(balances, crossChain.key, token.address)
+
+    const gasEstimate =
+      lastStep.estimate.gasCosts &&
+      lastStep.estimate.gasCosts.length &&
+      lastStep.estimate.gasCosts[0].amount
+    const requiredAmount = new BigNumber(gasEstimate || 0).shiftedBy(-18)
+    return !balance.isZero() && balance.gte(requiredAmount)
+  }
```
mathiasmoeller, line 687: We should have extracted that code long time ago 😅  It's in at least 3 files now

src/components/SwapEtherspotKlimaZap.tsx
```diff
@@ -0,0 +1,1189 @@
...
+    (chainKey: ChainKey, tokenId: string) => {
+      const token = tokens[chainKey].find((token) => token.address === tokenId)
+      if (!token) {
+        throw new Error('Token not found')
+      }
+      return token
+    },
+    [tokens],
+  )
+  // TODO: fix TODOs here
```
mathiasmoeller, line 699: 😂

src/components/SwapEtherspotKlimaZap.tsx
```diff
@@ -0,0 +1,1189 @@
...
+              disabled
+              style={{ color: 'rgba(0, 0, 0, 0.85)', fontWeight: '400' }}
+            />
+          </div>
+        </Col>
+      </Row>
+    )
+  }
+
+  const isTransferto = window.location.href.includes('transferto')
```
mathiasmoeller, line 948: I guess we don't need those checks anymore, maybe we should remove them as part of a separate story?

src/components/SwapEtherspotKlimaZap.tsx
```diff
@@ -0,0 +1,1189 @@
+import './SwapEtherspotKlimaZap.css'
```
mathiasmoeller, line 1: It's hard to find all the Etherspot related code in here. But we gonna test this a lot anyway I guess

src/components/SwapEtherspotKlimaZap.tsx
Addminus: Seperate story! This PR is already big enough!

src/components/SwapEtherspotKlimaZap.tsx
Addminus: I hope!

PR #286, opened 2022-04-26
feat(swap): warn user about high costs

Addminus:
FEEDBACK WANTED.
Currently I have played around with putting the warning in two spots. A popover when clicking the `swap` BTN and in the swap modal itself.  I liked the idea of not adding clicks to the process, but the entire message bloated the swap modal up.

Possibilities:
- Show only Popover on swap button
- Show warning in modal
- Show  full warnings in both places (kind of like in the pictures but with all info in modal)
- Show the warning popover when clicking the `Start cross chain swap` BTN in modal

image
image


mathiasmoeller:
Good job 👍 
I like the warning BEFORE you enter the Swap Modal (the popover) best. No need to even get there if the route doesn't make sense.
I also think the extra click needed ist not a big deal since this should only show up if the Route result is really bad (we might have to define what that means). For "good" routes this shouldn't show up so in the normal flow the user won't be bothered by it.

src/components/Swap.tsx
```diff
@@ -825,17 +826,58 @@ const Swap = () => {
...
-        shape="round"
-        type="primary"
-        icon={<SwapOutlined />}
-        size={'large'}
-        onClick={() => openModal()}>
-        Swap
-      </Button>
+    const totalExpenditure = fromAmountUSD.plus(gasCostUSD)
+    const amountReceivedPercentage = toAmountUSD.dividedBy(totalExpenditure)
+    const receivedAmountTooLow = amountReceivedPercentage.isLessThan(new BigNumber(0.75))
```
mathiasmoeller, line 837: We might want to be able to disable this thing here for our development I think 😅 I guess most of our test transfers might not be suitable here and we don't want to hide that message every time?

mathiasmoeller: Maybe something like a "don't show again" flag in local storage?

mathiasmoeller (commented):
Could we maybe extract that logic to keep `Swap.tsx` smaller? This way it would also be easier to add it to /ukraine and the new demo page.
And we could maybe reuse it in the widget?

PR #291, opened 2022-04-29
Lf-712: simple transaction as LIFI route alternative

Addminus:


src/components/SwapUkraine.tsx
```diff
@@ -692,6 +690,8 @@ const Swap = () => {
         try {
           currentRouteCallId = id
           const result = await LiFi.getRoutes(request)
+          // const result = await getRoute(request, web3.library?.getSigner())
```
mathiasmoeller, line 693: Is this important? Just to make sure it's not forgotten

mathiasmoeller (approved):
🤷‍♂️🙆‍♂️

PR #294, opened 2022-05-04
fix(Swapping): get rid of side effects

mathiasmoeller:
We were misusing unexpected side effects in the SDK's `executeRoute` method. We should use the reference to the initial `route` object but instead use the updated version provided by the `updateRoute` callback.

This PR is missing the same fixes for `SwappingCarbonOffset.tsx` and `SwappingEtherspotKlima.tsx` but I want to make sure the SDK fixes in https://github.com/lifinance/sdk/pull/62 actually work as expected.

### DIFF
--- a/src/components/Swapping.tsx
+++ b/src/components/Swapping.tsx
@@ -73,10 +73,9 @@ const Swapping = ({
   onSwapDone,
   fixedRecipient = false,
 }: SwappingProps) => {
-  const { steps } = route
-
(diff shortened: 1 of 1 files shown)


Addminus:
I have converted this to a draft PR because of the needed changes in `SwappingCarbonOffset.tsx` and `SwappingEtherspotKlima.tsx`. I'll go through the changes once the Etherspot showcases have been organized and cleaned up and apply them to the needed files

PR #312, opened 2022-05-19
refactor: improve swap success message handlers

Addminus:


src/hooks/useStepReturnInfo.ts
```diff
@@ -0,0 +1,43 @@
...
+  totalBalanceOfReceivedToken: TokenAmount | null
+  receivedTokenMatchesPlannedToken: boolean
+  toChain: Chain
+}
+
+export const useStepReturnInfo = (step: Step) => {
+  const web3 = useWeb3React<Web3Provider>()
+  const [stepReturnInfo, setStepReturnInfo] = useState<stepReturnInfo>()
+
+  useEffect(() => {
```
H3xept, line 21: AFAIK useEffect should always have a dependency array. Like this, the useEffect will re-execute after each re-render. Maybe we should have `useEffect(..., [step])` ?

src/hooks/useStepReturnInfo.ts
```diff
@@ -0,0 +1,43 @@
...
+    const getInfo = async () => {
```
H3xept, line 22: I think this should be declared outside of the `useEffect` by using a `useCallback` hook. This allows for function redefinition only when one of the fn's dependencies change

H3xept (changes requested):

src/hooks/useStepReturnInfo.ts
Addminus: whoops meant to do that

H3xept (approved):

PR #315, opened 2022-05-19
refactor: replace getpossibilities

Addminus:


src/components/Swap.tsx
```diff
@@ -248,50 +238,53 @@ const Swap = () => {
...
+      const resolvedSetupPromises = await Promise.all([chainPromise, tokenPromise, toolsPromise])
+      const chains = resolvedSetupPromises[0]
+      const tokens = resolvedSetupPromises[1].tokens
+      const tools = resolvedSetupPromises[2]
+
+      if (!chains || !tools.bridges || !tools.exchanges || !tokens) {
         // eslint-disable-next-line
         console.warn('possibilities request did not contain required setup information')
         return
       }
```
maxklenk, line 254: One reason to split the possibilities request into multiple ones was to make the responses smaller and faster. It would be great if we could handled them independently in the `Swap.tsx` implementation. This will especially speed up how quickly the chain selection can be displayed, even if the big chunk of tokens is still loading.

maxklenk (changes requested):

src/components/Swap.tsx
Addminus: Good point! Alle requests are now split up. I also introduced a new Provider which handles these setup requests and provides the data to all subpages. This further increases loading speed as the individual sites don't have to send of their individual requests for that data. 

This resulted in a bigger PR however...

mathiasmoeller (approved):
I like the the Context. This is how we should have done it from the beginning I guess!

PR #317, opened 2022-05-20
fix: migrate tool detail

Addminus:


src/services/localStorage.ts
```diff
@@ -248,6 +233,36 @@ const deleteWalletConnectInfo = () => {
...
+
+// TODO: Migrating and adding toolDetails. Can be removed after being deployed for a while (deployed on 20.05.2022)
+const migrateToolDetails = () => {
+  if (!isSupported()) return
+  const oldRoutes = readAllRoutes()
+  const updatedRoutes = oldRoutes.map((oldRoute) => {
+    oldRoute.steps.map((step) => {
+      step.toolDetails = { logoURI: '', name: step.tool, key: step.tool.split('-')[0] }
+      return step
+    })
```
maxklenk, line 258: I think we should add a filter to only add `toolDetails` if they don't exist already.

maxklenk (changes requested):

PR #339, opened 2022-06-08
feat: widget integration draft

chybisov:


maxklenk (changes requested):
Pretty cool. I like a lot how well the wallet management of the existing site can be used. Visual/Functional feedback:

Should be fixed:
- [ ] increase hight based on screen size (avoid scrolling)
- [x] "Choose a wallet" modal can't be closed when opened from widget
- [ ] Maybe "Swap % Bridge" as headlin in widget (make configurable?)
- [ ] Approval spinner stayed pending after replacing (speeding up the transaction)

Consider later:
- show historic routes
- two column view as on updated swap page
- dark mode for whole Swap V2 page

chybisov:
@maxklenk 

1. Let's see what we can do here.
2. Fixed.
3. This headline should be translatable, so not sure we need to make it configurable.
4. Could you please describe it in more detail?

chybisov:
@maxklenk issue with the replaced transaction should be fixed now. Are we ready to merge?

maxklenk:
I have added a feature flag, so we can show the new swap on staging, while sill hiding it on production. It only hides the menu item.

Can we please improve the mobile size of the widget (we will need different breakpoints):
- [ ] check background size (white area on the right)
- [ ] add some horizontal margin
- [ ] remove the space above and below the widget
- [ ] Try to place the support/social media links below the widget instead of fixed, the can be visible after scrolling only (we may also want to add a support link in the widget footer instead?)
Screenshot 2022-07-06 at 9 28 06 AM

- [ ] there is some problem with the menu on mobile
Screenshot 2022-07-06 at 9 22 27 AM

Safari:
- [ ] top border radius is hidden by overlapping element
- [ ] right scrollbar is hidden by overlapping element
- [ ] some content is to wide, adding a vertical scroll bar
Screenshot 2022-07-06 at 9 39 03 AM

Other things:
- [ ] Hide "/" in "You pay" box if no wallet is connected

Windows:
- not tested by we may need to check to prevent ugly scroll bars

Addminus:
Can we add a feature flag for develop and staging? that way it does not end up on production by accident

Addminus (changes requested):
I have two things: The notification overlay was removed. But its showing in this branch
image

The beta badge on mobile clips with the text
image

chybisov:
@maxklenk @Addminus, please take a look at all changes, and if you don't have any other suggestions, I'll merge it.

maxklenk (approved):
Awesome @chybisov, looks and works great. We should get it merged and tested by the team and community. Small UI improvements can be added later.

PR #340, opened 2022-06-14
fix: conditions for balance polling

Addminus:


src/components/SwapUkraine.tsx
```diff
@@ -257,14 +257,10 @@ const Swap = () => {
   }, [chainsTokensTools.bridges, chainsTokensTools.exchanges])
 
   useEffect(() => {
-    if (
-      chainsTokensTools.chainsLoaded &&
-      chainsTokensTools.tokensLoaded &&
-      chainsTokensTools.toolsLoaded
-    ) {
+    if (availableChains.length !== 0 && Object.keys(tokens).length !== 0) {
```
H3xept, line 260: This check is repeated a lot, maybe extract it to an helper fn?

H3xept (approved):

src/components/SwapUkraine.tsx
Addminus: can do!

PR #341, opened 2022-06-14
Production Release

Addminus:


src/components/Swap.tsx
```diff
@@ -278,7 +278,7 @@ const Swap = () => {
 
   //get tools
   useEffect(() => {
-    setAvailableExchanges(chainsTokensTools.bridges)
+    setAvailableExchanges(chainsTokensTools.exchanges)
```
H3xept, line 281: 🙈

src/components/SwapUkraine.tsx
```diff
@@ -283,33 +283,27 @@ const Swap = () => {
 
   // autoselect from chain based on wallet
   useEffect(() => {
-    LiFi.getChains().then((chains: Chain[]) => {
-      const walletChainIsSupported = chains.some((chain) => chain.id === web3.chainId)
+    if (!fromChainKey && startParamsDefined) {
```
H3xept, line 286: This useEffect block is used in multiple places, maybe abstract it into a hook?

src/components/Swap.tsx
Addminus: yep...

src/components/SwapUkraine.tsx
Addminus: multiple useEffect hooks in this repo are used *a lot*. I agree it would be nice but this repo has entered maintenance mode. The new transferto repository has been created and the work on the new site using the widget has started.

PR #357, opened 2022-06-30
fix: update Gnosis RPC

mathiasmoeller:


Addminus:
Seems like it solves the problem. Kind of bad from them to change the url like that

Addminus (approved):

PR #419, opened 2022-09-13
Lf 1524 extend flows for claim page

Addminus:


dennyscode:
Success State:
image

Error State:
image

Pending State:
image

Claiming State:
...

Not Qualified State:
image

PR #502, opened 2023-04-03
feat: implement new wallet menu

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-2703

[screen-capture-connected-menu.webm](https://user-images.githubusercontent.com/43956540/229505724-5c3b8ee4-74d8-41b0-afb2-7857da6d8073.webm)


packages/dapp/src/components/Navbar/Menu/ConnectedMenu/ConnectedMenu.tsx
```diff
@@ -1,55 +1,161 @@
...
+          },
+        }}
+      >
+        <SpotButton
+          name="Copy"
+          onClick={() => {
+            navigator?.clipboard?.writeText(account.address);
+            setCopiedToClipboard(true);
+            trackEvent({
+              category: 'menu',
```
Abhikumar98, line 109: For tracking fields,
`category`, `action` and `labels`; do we have events defined in a `type` or we are using strings?

if it's just typed as string then there is a possible case of typo in future which could affect the analytics

packages/dapp/src/components/Navbar/Menu/ConnectedMenu/ConnectedMenu.tsx
```diff
@@ -1,55 +1,161 @@
...
+        <SpotButton
+          name="Explore"
+          onClick={() => {
+            openInNewTab(
+              `${
+                getChainById(account.chainId).metamask.blockExplorerUrls[0]
+              }address/${account.address}`,
+            );
+            trackPageload({
+              source: 'connected-menu',
```
Abhikumar98, line 127: same as above, is this a string or typed field with defined strings?

```ts
type Sources = "connected-menu" | "some-other-place" | "another-place";
```

packages/dapp/src/components/Navbar/Menu/ConnectedMenu/ConnectedMenu.tsx
```diff
@@ -1,55 +1,161 @@
...
+              label: 'copyAddressToClipboard',
+              disableTrackingTool: [EventTrackingTools.arcx],
+            });
+          }}
+        >
+          <ContentCopyIcon />
+        </SpotButton>
+        <SpotButton
+          name="Explore"
+          onClick={() => {
```
Abhikumar98, line 120: I'd suggest define a function for `onClick`, and then use it here.

To increase readability and keep the render part of code clean :)

PS: My general rule is, if onClick goes more than 3-4 lines, then define a function for it. But lmk your thoughts on the same.

packages/dapp/src/components/Navbar/Navbar.style.tsx
```diff
@@ -305,7 +305,7 @@ export const NavbarPaper = styled(Paper, {
     '> .navbar-menu-list': {
       marginTop: 0,
       padding: !!isOpenSubMenu
-        ? openSubMenu === 'wallets'
+        ? openSubMenu === 'wallets' || openSubMenu === 'chains'
```
Abhikumar98, line 308: Similar comment as above,
Is `openSubMenu` typed or expecting regular string?

packages/dapp/src/providers/MenuProvider.tsx
```diff
@@ -95,6 +87,17 @@ const MenuProvider = ({ children }: MenuProviderProps) => {
...
   };
 
+  // Toggle Navbar Chains Menu
+  const onOpenNavbarChainsMenu = (open: boolean) => {
+    setMenu((oldSettings) => ({
+      ...oldSettings,
+      openNavbarSubMenu: open
+        ? SubMenuKeys.chains
+        : (SubMenuKeys.none as string),
+      openNavbarChainsMenu: open as boolean,
```
Abhikumar98, line 97: `open` already is being expected as boolean, this assertion isn't required :)

packages/shared/src/atoms/ChainSwitch/ChainSwitch.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const activeChain = useMemo(
+    () => chains.find((chainEl: Chain) => chainEl.id === account.chainId),
+    [chains, account.chainId],
+  );
+
+  const handleOpenChainsMenu = () => {
+    menu.onOpenNavbarChainsMenu(true);
+
+    trackEvent({
+      category: 'chain-menu',
```
Abhikumar98, line 34: https://github.com/lifinance/jumper.exchange/pull/502/files#r1157127574

packages/shared/src/atoms/ChainSwitch/ChainSwitch.style.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const ButtonChainSwitch = styled(IconButton)(({ theme }) => ({
+  color: theme.palette.text.primary,
+  width: '48px',
+  height: '48px',
+  marginLeft: theme.spacing(3),
+  backgroundColor:
+    theme.palette.mode === 'dark'
+      ? theme.palette.alphaLight300.main
+      : theme.palette.white.main,
```
Abhikumar98, line 13: I see this piece being repeated in couple of places. Can we make a function for this?

```ts
const getThemeBackgroundColor = (mode: "light" | "dark") => {
    return theme.palette.mode === 'dark'
      ? theme.palette.alphaLight300.main
      : theme.palette.white.main,
}
```

packages/shared/src/atoms/ChainSwitch/ChainSwitch.style.tsx
```diff
@@ -0,0 +1,32 @@
...
+  zIndex: 1400,
+  '&:hover': {
+    backgroundColor:
+      theme.palette.mode === 'dark'
```
Abhikumar98, line 17: https://github.com/lifinance/jumper.exchange/pull/502/files#r1157138142

packages/shared/src/atoms/ThemeSwitch/ThemeSwitch.style.tsx
```diff
@@ -13,6 +13,7 @@ export const ButtonThemeSwitch = styled(IconButton)(({ theme }) => ({
     theme.palette.mode === 'dark'
```
Abhikumar98, line 13: not in your changes, but could be another possible place to implement it

`getDefaultThemeBackgroundColor `
https://github.com/lifinance/jumper.exchange/pull/502/files#r1157138142

packages/dapp/src/components/Navbar/Menu/ConnectedMenu/ConnectedMenu.tsx
dennyscode: That´s a great point! I have not thought of that, but adding some kind of object with some pre-defined strings makes a lot of sense!

packages/dapp/src/components/Navbar/Menu/ConnectedMenu/ConnectedMenu.tsx
dennyscode: Yea, like this! Thanks, sir! :)
Since we do have those tracking-functions all over, this should be part of a new task. But I will work that out.

packages/dapp/src/components/Navbar/Navbar.style.tsx
dennyscode: We had it in place: subMenuKeys - jjust gotta use it! Done now :)

packages/dapp/src/providers/MenuProvider.tsx
dennyscode: Good point! Had this all over the place, now removed

packages/dapp/src/components/Navbar/Menu/ConnectedMenu/ConnectedMenu.tsx
dennyscode: I like that and I will go with your rule-of-thumb and will apply it to my coding.

dennyscode:
Added label "wait", because we also need to adjust the translations for the following wordings:

1.) **Copy**
2.) **Explore**
3.) **Copied to Clipboard** --> pops up as a React-Snackbar when "Copy" was clicked
4.) Not sure if we want to adjust the tooltip for "Network"? When hovering over that IconButton in the navigation. Currently it is returning the active network (e.g. "Optimism") but it could say "Change network".
5.) Same applies to the ThemeSwitch-Button, which just displays the other state like "Dark" (when it´s currently in light mode). Instead we could say "Switch theme to dark", or similar.

dennyscode:
Also, we should clear if the menu should always close: Currently it is closing when the user disconnects his wallet. If clicking on "Copy" or "Explore" it stays open. Does that work or should we adjust it?

packages/shared/src/atoms/ChainSwitch/ChainSwitch.style.tsx
Abhikumar98: Got it.
My major concern was regarding the usage of same code in multiple places.

Seemed like a good case to extract it out to a helper function to avoid duplication and maintain consistency.

packages/dapp/src/i18n/en.json
```diff
@@ -16,13 +16,17 @@
...
+      "copy": "Copy",
+      "explore": "Explore",
+      "copiedMsg": "Wallet address copied"
     },
     "language": {
       "key": "Language",
       "value": "English"
     },
     "themes": {
+      "changeTheme": "Change Theme",
```
dennyscode, line 29: @chybisov Could you give me a go on this if this works properly with Crowdin? Dont wanna break stuff but I hope I did it right to just add to the English translation (en.json). Thankkkkks! :)

packages/shared/src/atoms/ChainSwitch/ChainSwitch.style.tsx
chybisov: We have a lot of places where we use `theme.palette.mode === 'dark'` condition, and it's not always about `backgroundColor`. So we need to go with one approach everywhere, whether we create a helper function for all or leave it as it is. I'm not sure about the helper functions though, the current approach is more straight to the point. 🙂

packages/dapp/env.d.ts
```diff
@@ -1,6 +1,11 @@
 /// <reference types="vite/client" />
 
 interface ImportMetaEnv {
+  readonly BASE_URL: string;
+  readonly DEV: boolean;
+  readonly MODE: 'development' | 'testnet' | 'production' | 'staging';
+  readonly PROD: boolean;
+  readonly SSR: boolean;
```
dennyscode, line 8: This is not necessary, right? I am not sure about the type definitions on those env vars.
Does it even make sense to define the MODE variants here?

chybisov:
@Addminus, It was a bit premature move to merge this before the review 😄

@dennyscode please take a look at the video below. All menus open animation pop up from the top right corner instead of their anchor place. I remember this was an issue with the first menu version at some point as well, so it should be easy to fix 🙂

https://user-images.githubusercontent.com/18644653/232698827-d9a600d2-2703-48c7-af18-e5c343dbbe5c.mov

PR #506, opened 2023-04-05
feat: only show testnets on jumper-testing-url

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-2659

Thanks @abhishek-lifi


packages/dapp/src/components/Widget/Widget.tsx
```diff
@@ -3,5 +3,7 @@ import { useWidgetConfig } from '../../hooks/';
 
 export function Widget({ starterVariant }) {
   const widgetConfig = useWidgetConfig((starterVariant = { starterVariant }));
-  return <LiFiWidget config={widgetConfig} />;
+  return (
+    <LiFiWidget config={widgetConfig} integrator={'testnets-jumper.exchange'} />
```
chybisov, line 7: We should provide an integrator string based on the MODE variable to have the same code in every branch (testnet/dev/staging/main) without limitations.

packages/dapp/src/const/MenuConfig/ConnectedMenu/Connected-SubMenu-Chains.tsx
```diff
@@ -13,9 +13,18 @@ const ConnectedSubMenuChains = () => {
     [chains.length, account.chainId],
   );
 
+  let availableChains = [...chains];
```
chybisov, line 16: Probably unnecessary. Should be `let availableChains = chains` instead?

packages/dapp/src/const/MenuConfig/ConnectedMenu/Connected-SubMenu-Chains.tsx
```diff
@@ -6,16 +6,25 @@ import { useWallet } from '../../../providers/WalletProvider';
...
+  const { chains } = useChainInfos();
   const activeChain = useMemo(
     () => chains.find((chainEl: Chain) => chainEl.id === account.chainId),
-    [chains.length, account.chainId],
+    [chains, account.chainId],
   );
 
+  let availableChains = chains;
+
+  if ((import.meta as any).env.VITE_LIFI_SHOW_TESTNET) {
```
Addminus, line 18: You could simplify that check by checking the build `mode`. You added `vite --mode testnet` to the package.json scripts

packages/dapp/.env.testnet
```diff
@@ -2,5 +2,7 @@ VITE_GOOGLE_ANALYTICS_TRACKING_ID=G-T6ZDXLLXB5
 VITE_HOTJAR_ID=3307684
 VITE_HOTJAR_SNIPPET_VERSION=6
 VITE_LIFI_API_URL=https://staging.li.quest/v1/
+VITE_LIFI_SHOW_TESTNET=true
```
Addminus, line 5: checking for the build mode would remove the need for this env var

packages/dapp/env.d.ts
```diff
@@ -1,12 +1,18 @@
 /// <reference types="vite/client" />
 
 interface ImportMetaEnv {
+  BASE_URL: string;
```
Addminus, line 4: they should be read only, right?

Addminus (changes requested):

dennyscode:
Thanks Adrian! I did the fix!

packages/dapp/.env.testnet
```diff
@@ -4,3 +4,4 @@ VITE_HOTJAR_SNIPPET_VERSION=6
 VITE_LIFI_API_URL=https://staging.li.quest/v1/
 VITE_ARCX_API_KEY=bcc8c20c6f4fe15b16184a2cc8c2e1cb9125c7b574df7c0954984970a8d9de06
 VITE_CUSTOM_RPCS={ "59140": ["https://consensys-zkevm-goerli-prealpha.infura.io/v3/faf4bc4ea7344e5da5e56c55de087480"] }
+VITE_WIDGET_INTEGRATOR=testnet.jumper.exchange
```
dennyscode, line 7: Is there anything wrong here? It´s not highlighted compared to the other env files.

packages/dapp/env.d.ts
```diff
@@ -1,12 +1,19 @@
 /// <reference types="vite/client" />
 
 interface ImportMetaEnv {
+  readonly BASE_URL: string;
+  readonly DEV: boolean;
+  readonly MODE: 'development' | 'testnet' | 'production' | 'staging';
+  readonly PROD: boolean;
+  readonly SSR: boolean;
```
dennyscode, line 8: Is this unnecessary? Still running into typescript issues on those env´s! 

1.) I guess overwriting such as BASE_URL, DEV, PROD or SSR does not make much sense, right? It is taken from the Vite´s ImportMetaEnv.
2.) But does it make sense to define the MODE types? Or is it bad practise?

Here is the error I wish to get fixed:
image

dennyscode: ℹ️ Of course, same applies also the the develop-branch. This PR here is only being merged into the testnet-branch. So, any improvements need to be done to develop as well.

PR #507, opened 2023-04-11
feat: use consts for re-usable tracking strings

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-2731


packages/dapp/src/const/trackingKeys.ts
```diff
@@ -0,0 +1,29 @@
...
+  onRouteExecutionFailed = 'onRouteExecutionFailed',
+  onRouteExecutionUpdated = 'onRouteExecutionUpdated',
+  openConnectedMenu = 'open-connected-menu',
+  openSubmenu = 'open-submenu',
+  openSupportModal = 'open-support-modal',
+  switchChain = 'switch-chain',
+  switchLanguage = 'switch-language',
+  switchTab = 'switch-tab',
+  switchTheme = 'switch-theme-mode',
+}
```
chybisov, line 29: Let's stick to the [classic](https://www.typescriptlang.org/docs/handbook/enums.html) naming convention of enums and their properties here 🙂

chybisov (changes requested):

packages/dapp/src/const/trackingKeys.ts
dennyscode: @chybisov I hope that I hit the nail with those adjustments in my last commit? ⚒️

Addminus (approved):

packages/dapp/src/const/trackingKeys.ts
chybisov: @dennyscode sorry, but where did you find UPPER_CASE there? 😅

I meant this

```ts
enum Direction {
  Up,
  Down,
  Left,
  Right,
}
```


I should be more specific next time 😅

PR #514, opened 2023-04-14
refactor: use map func in chains-loop

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-2795

Small ticket here to fix a few lines: Replacing some weirdo loop + push to get an array by just using a simple map!


dennyscode:
Oh well that last commit "refactor rename trackingKeys" belongs to its own ticket and was reverted

Addminus (approved):

PR #519, opened 2023-04-18
fix: items should expand full width

dennyscode:
No Ticket, quick & dirty - fix this layout to align items vertically in a grid
Solution: Make them use 12 cols

image


dennyscode:
Yes, this was not clear and should have had the solved image instead! The image still shows the bug

PR #528, opened 2023-04-22
feat: integrate onRamper widget

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-2786

Current Version, still fixes to be done:

[onramper-integration-screen-capture.webm](https://user-images.githubusercontent.com/43956540/233752825-068d8f9a-de99-4fc4-8ed0-6cf7c531f78e.webm)

image
image


packages/dapp/src/components/OnRamper/OnRamper.tsx
```diff
@@ -0,0 +1,58 @@
...
+    return str.substring(1);
+  }
+  return str;
+}
+
+export const OnRamper = () => {
+  const theme = useTheme();
+
+  const onRamperConfig = {
+    defaultCrypto: 'ETH',
```
dennyscode, line 15: Something still doesn´t seem right with the defaultCrypto. Further down, on where this is used I run into the following issue - it returns AAVE (probably because it is not correctly parsed). Whenever I move this "defaultCrypto" as the first parameter to be passed in the query-string, it works like a charne - but then the themeMode does not apply correctly.
Seems, as those have to be first in line? Idk, I left it as it is, having AAVE as a default for now.

Addminus (approved):

PR #537, opened 2023-05-06
feat: add welcome component

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-2965

image

Todo:

- [x] Check fonts GT America, for license
- [x] GT America is loaded after page is initialized, leading to flickering
- [x] Add the Light-Theme version (!!)
- [x] scrolling is too much on starting screen with welcome stuff
- [x] check responsiveness
- [x] hide introduction stuff on click on any other element (currently just "Get started" disables all that)
- [x] [LF-3074](https://lifi.atlassian.net/browse/LF-3074): Use Server endpoints for chains and tools
- [x] [LF-3075](https://lifi.atlassian.net/browse/LF-3075): Disable interaction with website
- [x] [LF-3076](https://lifi.atlassian.net/browse/LF-3076): Wallet depended Welcomes screen display
- [x] [LF-3077](https://lifi.atlassian.net/browse/LF-3077): Adjust Welcome screen height

[LF-3074]: https://lifi.atlassian.net/browse/LF-3074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-3075]: https://lifi.atlassian.net/browse/LF-3075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-3076]: https://lifi.atlassian.net/browse/LF-3076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-3077]: https://lifi.atlassian.net/browse/LF-3077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dennyscode:
Current Version:

[screen-capture-introduction-component.webm](https://user-images.githubusercontent.com/43956540/236589478-f2032ee6-4f1b-4f28-878c-bbe267632960.webm)

dennyscode:
Current Version:

[welcome-screen-capture.webm](https://user-images.githubusercontent.com/43956540/236835662-4f35d0ca-284e-48b0-801b-ed3dcbbd5809.webm)

dennyscode:
Current Version with loading items highlighted:

[welcome-loading-screen-capture.webm](https://user-images.githubusercontent.com/43956540/236836204-0722edbd-d35c-4335-856f-86af1b4e4797.webm)

dennyscode:
@chybisov I used some tricks and workarounds to make this animation work:

1.) I have duplicated the widget in the "WelcomeWrapper.tsx". This way, I can add a opacity transition effect on the whole glow-effect which is kind of the highlight here. But after this transition, we still wanna have the Widget showing up which is why it´s duplicated.
2.) After the ClickEvent once the Animation finished, it will show up the original Widget as we used to.

I am not sure if there might be better ways to solve it and would like you to take a look on the performance. For this, I added the video above to represent re-renders. Thanks a lot for your efforts and support!

dennyscode:
[welcome-loading-screen-capture-lighttheme.webm](https://user-images.githubusercontent.com/43956540/236882129-9e40e3db-d5ff-4cb7-bd79-ac3ff5cc23ef.webm)

chybisov:
@dennyscode this is how I see it 🙂 

Some highlights:
1. Shadow color looks grayish instead of glowing and looks like a line.
2. Height needs to be corrected.
3. Please take a look at the error in the console.
4. Connect wallet button in the widget appears on top during the transition.
5. We should show this screen only once and save the state to localStorage. Also, clicking on the logo should show it again.

https://github.com/lifinance/jumper.exchange/assets/18644653/f17cddba-96f9-43a2-a9a9-4c21596f12c5

chybisov:
> @chybisov I used some tricks and workarounds to make this animation work:
> 
> 1.) I have duplicated the widget in the "WelcomeWrapper.tsx". This way, I can add a opacity transition effect on the whole glow-effect which is kind of the highlight here. But after this transition, we still wanna have the Widget showing up which is why it´s duplicated. 2.) After the ClickEvent once the Animation finished, it will show up the original Widget as we used to.
> 
> I am not sure if there might be better ways to solve it and would like you to take a look on the performance. For this, I added the video above to represent re-renders. Thanks a lot for your efforts and support!
> 
> 1.) Another thing is the Font here, which flickers. We have had this issue before as I remember and need to get this fixed once again. I guess adding it to the CSS before the page is actualy built was the fix, will be looking into it.

It would be better to avoid adding a new widget just for this case. I'm sure we can avoid this. Let's hop on a call tomorrow if you are still looking for a way to avoid it 🙂

packages/dapp/src/components/DualWidget/DualWidget.tsx
```diff
@@ -56,35 +65,90 @@ export function DualWidget() {
...
+    }, 300);
+  };
+
+  const isWalletConnected = useMemo(() => {
+    const activeWallet = JSON.parse(
+      localStorage.getItem(LOCAL_STORAGE_WALLETS_KEY),
+    );
+    return activeWallet?.length > 0 || false;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [account.isActive]);
```
dennyscode, line 81: @chybisov Actively disabling the linter on here - I want this to be re-loaded whenever there is a change in the account. Although linter tells me to remove it from the dependencies (because it is not used within the useMemo), I keep it. Is that a correct approach?

packages/dapp/src/components/DualWidget/DualWidget.tsx
```diff
@@ -56,35 +65,90 @@ export function DualWidget() {
...
-    <Grid
-      justifyContent="center"
-      alignItems="center"
-      container
-      sx={{
-        overflowX: 'hidden',
-      }}
+    <WelcomeWrapper
+      showWelcome={!isWalletConnected && showWelcome}
+      showFadeOut={!isWalletConnected && showFadeOut}
```
dennyscode, line 90: It is very repetetive and could be DRYer. By just combining both into a const which replaces them, would that be a better solution?

packages/dapp/src/components/DualWidget/DualWidget.style.tsx
```diff
@@ -1,19 +1,162 @@
...
+
+  '& > .widget-wrapper > div, & > .onramper-container': {
+    paddingTop: showWelcome && theme.spacing(6),
+  },
+
+  '& > div > div': {
+    top: showWelcome && theme.spacing(6),
+  },
+
+  '& > div > div, & > .onramper-container': {
```
dennyscode, line 111: @chybisov As my CSS is very unspecific, I thought it´d be cool if we could pass a className to the widget´s container. Do you think this adds value to our users / integrators as well? I´d sure love to be more specific on my selectors as this looks odd and not clean

packages/dapp/src/components/Widget/Widget.tsx
```diff
@@ -162,9 +163,11 @@ export function Widget({ starterVariant }) {
   ]);
 
   return (
-    <LiFiWidget
-      integrator={import.meta.env.VITE_WIDGET_INTEGRATOR as string}
-      config={widgetConfig}
-    />
+    <Box className="widget-wrapper">
+      <LiFiWidget
+        integrator={import.meta.env.VITE_WIDGET_INTEGRATOR as string}
```
dennyscode, line 168: @chybisov Thought about parsing a className here? Wdyt?

packages/dapp/src/components/DualWidget/DualWidget.style.tsx
```diff
@@ -1,19 +1,162 @@
...
+    position: 'absolute',
+    height: '150px',
+    pointerEvents: 'none',
+    left: 0,
+    right: 0,
+    bottom: '-4px',
+    transitionProperty: showWelcome && 'height, bottom',
+    transitionDuration: showWelcome && '.3s',
+    transitionTimingFunction: showWelcome && 'ease-in-out',
+  },
```
dennyscode, line 46: This chunk replaces the maskImage which was used before. It gives a gradient effect as a transition between the bottom welcome part and the top part of the widget container

chybisov: It doesn't work well for a dark theme. You can see on the screenshot there is a gradient on both sides, and it looks like a faded lightsaber in the middle of the screen. It should only be to the upside.

image

packages/dapp/src/components/Widget/Widget.tsx
chybisov: It needs to be apparent where this comes from, and why not use a styled component here?

packages/dapp/src/components/BackgroundGradient/BackgroundGradient.style.tsx
```diff
@@ -80,8 +80,8 @@ export const BackgroundGradients = styled('span')<any>(({ theme }) => ({
 }));
 
 export const BackgroundGradientBottomLeft = styled(BackgroundGradient, {
-  shouldForwardProp: (prop) => prop !== 'isDarkMode',
-})<BackgroundGradientProps>(({ theme, isDarkMode }) => ({
+  shouldForwardProp: (prop) => prop !== 'showWelcome',
```
chybisov, line 83: As I see, we don't use it. Should this be removed?

chybisov: Also, if we have `isDarkMode` props elsewhere, we should remove them in favor of `theme.palette.mode === 'dark'`.

packages/dapp/src/components/DualWidget/DualWidget.tsx
```diff
@@ -56,37 +66,146 @@ export function DualWidget() {
...
+                  : showWelcomeWrapper && showFadeOut
+                  ? NavbarHeight.SM
+                  : 0,
+              height:
+                showWelcomeWrapper && !showFadeOut
+                  ? '50%'
+                  : showWelcomeWrapper && showFadeOut
+                  ? `calc( 100% - ${NavbarHeight.SM} )`
+                  : '100%',
+            },
```
chybisov, line 145: Please let's move this mess to the styled component 😅

packages/dapp/src/components/DualWidget/DualWidget.tsx
```diff
@@ -56,37 +66,146 @@ export function DualWidget() {
...
+          }}
+        >
+          {import.meta.env.MODE === 'testnet' && (
+            <Grid item xs={12} mt={theme.spacing(6)}>
+              <TestnetAlert />
+            </Grid>
+          )}
+          <Fade in={true} easing={{ enter: '500ms' }}>
```
chybisov, line 167: Tried to test and still don't get why do we need this Fade here 🙂

packages/dapp/src/components/WelcomeWrapper/WelcomeWrapper.style.ts
```diff
@@ -0,0 +1,66 @@
...
+    height: 'auto',
+    top: 0,
+  },
+}));
+
+export const CustomColor = styled(Typography)(({ theme }) => ({
+  background:
+    theme.palette.mode === 'dark'
+      ? 'linear-gradient(270deg, #D35CFF 0%, #BEA0EB 94.17%)'
+      : 'linear-gradient(270deg, #31007A 0%, #8700B8 94.17%);',
```
chybisov, line 21: Why do we have such a value of `94.17`?

packages/dapp/src/components/WelcomeWrapper/WelcomeWrapper.style.ts
```diff
@@ -0,0 +1,66 @@
...
+    bottom: 0,
+    padding: theme.spacing(2, 2, 8),
+  },
+  '&:before': {
+    content: '" "',
+    height: '100%',
+    background:
+      theme.palette.mode === 'dark'
+        ? 'linear-gradient(180deg, #200a47 0%, #1A1033 21.15%)'
+        : 'linear-gradient(180deg, #e8dafb 0%, #e8dafb 21.15%);',
```
chybisov, line 58: Same here

packages/dapp/src/components/WelcomeWrapper/WelcomeWrapper.tsx
```diff
@@ -0,0 +1,117 @@
...
+              >
+                {
+                  <Trans
+                    i18nKey={`${i18Path}subtitle`}
+                    components={[
+                      // eslint-disable-next-line jsx-a11y/anchor-has-content
+                      <a
+                        className={'link-lifi'}
+                        href="https://li.fi"
+                        style={{ color: 'inherit' }}
```
chybisov, line 69: We use classes and inline styling here. Let's choose one approach.

packages/dapp/src/providers/DexsAndBridgesInfosProvider.tsx
```diff
@@ -0,0 +1,56 @@
...
+}) => {
+  const { data, isSuccess } = useQuery(
+    ['toolsStats'],
+    async () => {
+      const apiUrl = import.meta.env.VITE_LIFI_API_URL;
+      const result = await axios<ToolsResponse>({
+        method: 'GET',
+        url: `${apiUrl}/tools`,
+      });
+      return result;
```
chybisov, line 37: We should stop using axios and remove this dependency. We have migrated to a native fetch everywhere.

Also, we don't need a context as a cache layer here and should add a custom hook with react-query in it. react-query perfectly handles caching purpose.

packages/shared/src/fonts/gt-america.css
```diff
@@ -0,0 +1,63 @@
+@font-face {
+  font-family: GT-America;
+  font-style: bold;
+  font-weight: 800;
+  font-display: swap;
+  src: url('GT-America-Black.woff2') format('woff2'),
+    url('GT-America-Black.woff') format('woff');
+}
+
+
```
chybisov, line 10: Do we really want to blow out our bundle size with a new font just for a couple of titles on the welcome screen? I don't think so 😄

packages/dapp/src/components/DualWidget/DualWidget.style.tsx
dennyscode: Yes, that looked odd! Thank you for that idea, just fixed it. It is still just a color I picked and hard-coded here but to me it seems to adapt better to the rest of the background:

image

packages/dapp/src/components/Widget/Widget.tsx
dennyscode: It´s been a hazzle, the hover effect´s are triggered on parent components, resulting in changing styles on that given component. This is why I am passing a className which can be used on those parent styles.

packages/dapp/src/components/BackgroundGradient/BackgroundGradient.style.tsx
dennyscode: Great deal, good find and cleaned up!

packages/dapp/src/components/DualWidget/DualWidget.tsx
dennyscode: True dat!

packages/dapp/src/components/WelcomeWrapper/WelcomeWrapper.style.ts
dennyscode: Those values are copied from the Figma layout: It´s the gradient on that Welcome-Title.
image

packages/dapp/src/components/WelcomeWrapper/WelcomeWrapper.tsx
dennyscode: This also serves the purpose of checking what items are clicked when opening the Modals (Dex, Bridges,...).
It is being used in a "handleGetStarted" function, I am curious if you happen to find a cleaner way.

packages/shared/src/fonts/gt-america.css
dennyscode: Not sure if I am understanding correctly: You do not mean that we should rather not use that font here? But as I have added multiple font-files of the GT-America which are currently not used, that produces too much overhead, right?
So, we should still use GT-America, but just the files we have in use. Kinda thought of setting it up and having a whole bunch of GT-America ready for future usage - but it might not be the most efficient way in regards of performance.

So, I will clean this up!

packages/dapp/src/providers/DexsAndBridgesInfosProvider.tsx
dennyscode: For my notes: We also have axios to fetch the chains in a seperate hook. Gotta create another task to remove axios dependency on "ChainInfosProvider.tsx".

dennyscode: This is the ticket: https://lifi.atlassian.net/browse/LF-3299

packages/dapp/src/components/Widgets/Widgets.tsx
```diff
@@ -56,37 +63,70 @@ export function DualWidget() {
...
   }, [activeTab, onChangeTab, starterVariant, starterVariantUsed]);
 
-  useEffect(() => {
+  const handleGetStarted = async (event) => {
+    const classList = event?.target?.classList;
+    if (
+      classList?.contains?.('stats-card') ||
+      classList?.contains?.('link-lifi')
+    ) {
+      return 0;
```
chybisov, line 72: Why return 0 here?

packages/dapp/src/components/Widgets/Widgets.tsx
```diff
@@ -56,37 +63,70 @@ export function DualWidget() {
...
+    } else {
+      event.stopPropagation();
+      setShowWelcome(false);
+      await setTimeout(() => {}, 600);
```
chybisov, line 76: `await` has no effect here and I'm curious to know why do we even need this 😄

chybisov (changes requested):

packages/dapp/src/components/Widgets/Widgets.tsx
dennyscode: Well, thanks! A little memorial of those times when I had an extra "showFadeOut" - state. I am ready to forget about it and clean it up.

chybisov (approved):

PR #541, opened 2023-05-12
fix: parse on-ramper api key and more

dennyscode:
Re-Iterating on Ticket: https://lifi.atlassian.net/browse/LF-2786

_Leeroy | Onramper, [12 May 2023, 10:14:59]:
You're appending apiKey with ? change ? > &_

### DIFF
--- a/packages/dapp/src/components/OnRamper/OnRamper.tsx
+++ b/packages/dapp/src/components/OnRamper/OnRamper.tsx
@@ -35,22 +35,9 @@ export const OnRamper = () => {
     wgBorderRadius: 1.5,
   };
 
-  const searchParams = new URLSearchParams();
-  searchParams.append('themeName', onRamperConfig.themeName);
-  searchParams.append('defaultCrypto', onRamperConfig.defaultCrypto);
-  searchParams.append('containerColor', onRamperConfig.containerColor);
-  searchParams.append('primaryColor', onRamperConfig.primaryColor);
-  searchParams.append('secondaryColor', onRamperConfig.secondaryColor);
-  searchParams.append('cardColor', onRamperConfig.cardColor);
-  searchParams.append('primaryTextColor', onRamperConfig.primaryTextColor);
-  searchParams.append('secondaryTextColor', onRamperConfig.secondaryTextColor);
-  searchParams.append('borderRadius', onRamperConfig.borderRadius.toString());
-  searchParams.append(
-    'wgBorderRadius',
-    onRamperConfig.wgBorderRadius.toString(),
+  const searchParams = new URLSearchParams(
+    Object.entries(onRamperConfig).map(([key, value]) => [key, String(value)]),
   );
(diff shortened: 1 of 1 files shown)


chybisov (changes requested):

chybisov (approved):

packages/dapp/src/components/OnRamper/OnRamper.tsx
```diff
@@ -45,7 +45,7 @@ export const OnRamper = () => {
     onRamperConfig.secondaryTextColor
   }&borderRadius=${onRamperConfig.borderRadius}&wgBorderRadius=${
     onRamperConfig.wgBorderRadius
-  }?apiKey=${import.meta.env.VITE_ONRAMPER_API_KEY}`;
+  }&apiKey=${import.meta.env.VITE_ONRAMPER_API_KEY}`;
```
maxklenk, line 48: I would recommend using `URLSearchParams` instead. You can pass it a nice (even typed) object and let it handle the parsing into a query string.  https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#examples

chybisov: Good suggestion!

dennyscode: Thanks @maxklenk! Yes, that´s cleaner and I will fix that

packages/dapp/src/components/OnRamper/OnRamper.tsx
```diff
@@ -34,18 +34,25 @@ export const OnRamper = () => {
...
+  searchParams.append('secondaryColor', onRamperConfig.secondaryColor);
+  searchParams.append('cardColor', onRamperConfig.cardColor);
+  searchParams.append('primaryTextColor', onRamperConfig.primaryTextColor);
+  searchParams.append('secondaryTextColor', onRamperConfig.secondaryTextColor);
+  searchParams.append('borderRadius', onRamperConfig.borderRadius.toString());
+  searchParams.append(
+    'wgBorderRadius',
+    onRamperConfig.wgBorderRadius.toString(),
+  );
+  searchParams.append('apiKey', import.meta.env.VITE_ONRAMPER_API_KEY);
```
chybisov, line 52: It should be possible to simplify it to

`const searchParams = new URLSearchParams(onRamperConfig);`

Please check types, maybe conversion of some props to strings will be required 🙂

chybisov (changes requested):

packages/dapp/src/components/OnRamper/OnRamper.tsx
dennyscode: @chybisov: I have re-iterated on it. What do you think about the latest changes - basically looping through the existing onRamperConfig.

chybisov (approved):

Addminus (approved):

PR #565, opened 2023-06-19
feat: Safe Wallet multisig integration

Abhikumar98:


packages/dapp/src/components/MultiSigConfirmationModal/MultiSigConfirmationModal.tsx
```diff
@@ -0,0 +1,28 @@
...
+  onClose: () => void;
+}> = ({ open, onClose }) => {
+  return (
+    <Modal open={open} onClose={onClose}>
+      <MultiSigConfirmationModalContainer>
+        <MultiSigConfirmationModalContent>
+          <div>Request sent</div>
+
+          <Typography>
+            Please inform all the signers to confirm the transaction within the
```
Addminus, line 19: Can we make this multilang?

Addminus: i mean localized

Addminus (changes requested):

packages/dapp/public/manifest.json
```diff
@@ -1,6 +1,7 @@
 {
   "short_name": "Jumper Exchange",
   "name": "Jumper Exchange",
+  "description": "Multi-Chain Bridging & Swapping (powered by LI.FI) (testing)",
```
chybisov, line 4: Should we remove `(testing)` from here?

packages/dapp/src/components/DualWidget/WidgetEvents.tsx
```diff
@@ -112,9 +121,19 @@ export function WidgetEvents() {
     );
     widgetEvents.on(WidgetEvent.RouteExecutionFailed, onRouteExecutionFailed);
     widgetEvents.on(WidgetEvent.RouteHighValueLoss, onRouteHighValueLoss);
+    widgetEvents.on(WidgetEvent.SafeRouteInitiated, onSafeRouteInitiation);
```
chybisov, line 124: We can reuse RouteExecutionStarted event here, no need to introduce a new one

packages/dapp/src/components/DualWidget/WidgetEvents.tsx
```diff
@@ -112,9 +129,32 @@ export function WidgetEvents() {
...
+
+  return (
+    <>
+      <MultisigConnectedAlert
+        open={isMultisigConnectedAlertOpen}
+        onClose={() => setIsMultisigConnectedAlertOpen(false)}
+      />
+      <MultisigConfirmationModal
+        open={isMultiSigConfirmationModalOpen}
+        onClose={handleMultiSigConfirmationModalClose}
```
chybisov, line 156: We should stick to one way of `onClose` handling, preferably without inline functions.

packages/dapp/src/components/MultiSigConfirmationModal/MultiSigConfirmationModal.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const MultisigConfirmationModal: React.FC<{
+  open: boolean;
+  onClose: () => void;
+}> = ({ open, onClose }) => {
+  const i18Path = 'multisig.transactionInitiated';
+
+  const { t: translate } = useTranslation();
+
+  return (
+    <Modal open={open} onClose={null}>
```
chybisov, line 20: Shouldn't we pass `onClose` param here as well?

packages/dapp/src/components/Widget/Widget.tsx
```diff
@@ -19,6 +24,76 @@ export function Widget({ starterVariant }) {
...
 
+  const handleMultiSigTransactionDetails = async (txHash: string) => {
+    const safeProviderSDK = (account?.signer?.provider as any)?.provider?.sdk;
+
+    const safeTransactionDetails: GatewayTransactionDetails =
+      await safeProviderSDK.txs.getBySafeTxHash(txHash);
+
+    if (safeTransactionDetails.txStatus === TransactionStatus.SUCCESS) {
+      return {
+        status: 'SUCCESS',
```
chybisov, line 35: Maybe we should use 'DONE' here, as we always use DONE instead of SUCCESS in the SDK.

packages/dapp/src/components/Widget/Widget.tsx
```diff
@@ -19,6 +24,76 @@ export function Widget({ starterVariant }) {
...
+    };
+  };
+
+  const handleSendingBatchTransaction = async (
+    batchTransactions: BaseTransaction[],
+  ) => {
+    const safeProviderSDK = (account?.signer?.provider as any)?.provider?.sdk;
+
+    try {
+      console.log('COming here ', batchTransactions);
```
chybisov, line 71: I hope we will remove all console.logs before merging 🙂

packages/dapp/src/components/Widget/Widget.tsx
```diff
@@ -19,6 +24,76 @@ export function Widget({ starterVariant }) {
...
+      const { safeTxHash } = await safeProviderSDK.txs.send({
+        txs: batchTransactions,
+        params: {
+          safeTxGas: 500000,
```
chybisov, line 75: Are we sure we want this hardcoded?

packages/dapp/src/i18n/en.json
```diff
@@ -48,5 +48,17 @@
...
+  "multisig": {
+    "connected": {
+      "title": "Multisig wallet connected",
+      "description": "Please notify other wallet participants to be ready to sign.",
+      "cta": "Okay"
+    },
+    "transactionInitiated": {
+      "title": "Multiple signatures required",
+      "description": "Please notify other multisig wallet participants to sign before the transaction expires.",
+      "cta": "Okay"
```
chybisov, line 61: Let's add a top-level key `button` and put this `Okay` there to avoid duplicated translations.

chybisov (changes requested):

packages/dapp/src/components/MultiSigConfirmationModal/MultiSigConfirmationModal.tsx
Abhikumar98: I thought we want the user to acknowledge the message about it being a SAFE wallet, hence making the modal closable by clicking on the CTA rather than clicking outside and closing. wdyt?

packages/dapp/src/components/DualWidget/WidgetEvents.tsx
Abhikumar98: Got it. Sure, I'll fix it

packages/dapp/src/components/DualWidget/WidgetEvents.tsx
Abhikumar98: Got it, I'll pass an extra argument to indicate this is a SAFE Route

Abhikumar98: I'd prefer a separate event:

while emitting the events, we can pass only 2 args in `emitter.emit()` and it's always `EventType` followed by `route/data`.
So, we won't be able to pass a simple distinguisher whether the route is SAFE route or not.

and currently, on the receiving end of these events, there is no logic present it's directly function's being called and execution.

Hence, following the same structure I made another event rather than introducing logic in Jumper for it.

chybisov: On Jumper we already have access to Safe SDK and we can determine if we are executing a route inside Safe Wallet. Also, we can subscribe to the same event in different places so it's easier to utilize it.

packages/dapp/src/components/MultiSigConfirmationModal/MultiSigConfirmationModal.tsx
chybisov: Not sure, I don't like this technic 😄

packages/dapp/src/i18n/en.json
```diff
@@ -50,15 +50,16 @@
     "switchToMainnet": "Switch to Mainnet"
   },
   "multisig": {
+    "button": {
+      "cta": "Okay"
+    },
```
chybisov, line 60: I meant to add a button as a top-level key so everyone can use it in the future, not only multisig part. Also, let's name the key `okay`, `cta` is a too general term.

chybisov (changes requested):

packages/dapp/src/components/DualWidget/WidgetEvents.tsx
Abhikumar98: moving the logic from widget to jumper and resorting to use the `RouteUpdatedEvent`

packages/dapp/src/i18n/en.json
Abhikumar98: moving it up

packages/dapp/src/components/DualWidget/WidgetEvents.tsx
```diff
@@ -2,22 +2,18 @@ import { Route } from '@lifi/sdk';
...
+  WidgetEvent,
 } from '@lifi/widget';
 import { useEffect, useRef } from 'react';
 import { TrackingActions, TrackingCategories } from '../../const';
-import { useMenuStore } from '../../stores';
 
 export function WidgetEvents() {
   const lastTxHashRef = useRef<string>();
   const { trackEvent, trackTransaction, trackAttribute } = useUserTracking();
-  const [onOpenSupportModal] = useMenuStore((state) => [
```
Abhikumar98, line 18: Investigate why this is deleted

packages/dapp/src/components/MultisigConfirmationModal/MultisigConfirmationModal.tsx
```diff
@@ -0,0 +1,51 @@
+import { Modal, Typography } from '@mui/material';
+import React from 'react';
```
chybisov, line 2: Importing react is not necessary anymore, please remove it everywhere.

packages/dapp/src/components/Widget/Widget.tsx
```diff
@@ -157,12 +163,15 @@ export function Widget({ starterVariant }) {
...
       keyPrefix: `jumper-${starterVariant}`,
+      ...multisigWidget,
       sdkConfig: {
         apiUrl: import.meta.env.VITE_LIFI_API_URL,
         rpcs,
         defaultRouteOptions: {
           maxPriceImpact: 0.4,
+          allowSwitchChain: !isMultisigSigner, // avoid routes requiring chain switch for multisig wallets
         },
+        multisigConfig: { ...multisigSdkConfig },
```
chybisov, line 174: I guess destructuring is not needed here

packages/dapp/src/components/DualWidget/WidgetEvents.tsx
chybisov, line 1: Why do we have two WidgetEvents components? Maybe we should combine them together 🙂

packages/dapp/src/const/MenuConfig/useWalletSelectContent.tsx
```diff
@@ -34,43 +50,62 @@ export const useWalletSelectContent = () => {
...
   const walletMenuItems = useMemo<MenuListItem[]>(() => {
-    const installedWallets = supportedWallets.filter((wallet) =>
-      wallet.installed(),
+    const installedWallets = supportedWallets.filter(
+      async (wallet) => await wallet.installed(),
     );
 
     const notInstalledWallets = supportedWallets.filter(
       (wallet) => !wallet.installed() && wallet.name !== 'Default Wallet', // always remove Default Wallet from not installed Wallets
     );
```
chybisov, line 64: We should also adjust this part to utilize the async `installed` method and any other place if any.

packages/dapp/src/const/MenuConfig/useWalletSelectContent.tsx
```diff
@@ -34,43 +50,62 @@ export const useWalletSelectContent = () => {
...
 
+  useEffect(() => {
+    verifyMultisigEnvironment();
+  }, [account?.address]);
+
   const walletMenuItems = useMemo<MenuListItem[]>(() => {
-    const installedWallets = supportedWallets.filter((wallet) =>
-      wallet.installed(),
+    const installedWallets = supportedWallets.filter(
+      async (wallet) => await wallet.installed(),
```
chybisov, line 59: Async `filter` doesn't work like that.  Please review this line in the debugger and see how it works 🙂 Also, check `SelectWalletPage` implementation in the widget.

packages/dapp/src/hooks/useMultisig.ts
```diff
@@ -0,0 +1,205 @@
+import { WidgetConfig } from '@lifi/widget';
+import { useWallet } from '../providers/WalletProvider';
+import { BaseTransaction, ChainId, ChainKey, MultisigConfig } from '@lifi/sdk';
+import {
+  GatewayTransactionDetails,
+  TransactionStatus,
+} from '@safe-global/safe-apps-sdk';
+import SafeAppsSDK from '@safe-global/safe-apps-sdk/dist/src/sdk';
```
chybisov, line 8: Please import from `@safe-global/safe-apps-sdk`

packages/dapp/src/i18n/en.json
```diff
@@ -58,5 +58,22 @@
     "info": "You’re using Testnet",
     "testnet": "Any transactions or actions performed on the Testnet will not be reflected on the Mainnet.",
     "switchToMainnet": "Switch to Mainnet"
+  },
+  "button": {
+    "cta": "Okay"
```
chybisov, line 63: Should be `"okay": "Okay"`

chybisov (changes requested):

packages/dapp/package.json
```diff
@@ -7,11 +7,12 @@
...
     "@emotion/styled": "^11.11.0",
-    "@lifi/sdk": "^2.1.1",
-    "@lifi/wallet-management": "^2.1.3",
-    "@lifi/widget": "^2.1.4",
+    "@lifi/sdk": "^2.2.0-beta.0",
+    "@lifi/wallet-management": "^2.2.0-beta.2",
+    "@lifi/widget": "^2.2.0-beta.2",
     "@mui/icons-material": "^5.11.16",
     "@mui/material": "^5.13.6",
+    "@safe-global/safe-apps-sdk": "^7.11.0",
```
chybisov, line 15: Please bump it to the v8.

chybisov (changes requested):

packages/shared/src/types/settings.ts
```diff
@@ -31,4 +33,7 @@ export interface SettingsState extends SettingsProps {
 
   // Welcome Screen
   onWelcomeScreenEntered: (shown: boolean) => void;
+
+  // Destination Chain
+  onDestinationChainSelected: (chainId: number) => void;
```
Abhikumar98, line 38: @chybisov This was a single state use case, should I make a different store called `useMultisigStore` or since this is a minor state this would work?

packages/dapp/src/components/DualWidget/WidgetEvents.tsx
Abhikumar98: haha, this is what I was telling the other day.

packages/dapp/src/components/Widget/Widget.tsx
Abhikumar98: I'd like to create a copy of object wherever sending it. just to avoid any further unwanted updates. Although, it's more of a fail safe thing to do, if it doesn't make sense I can remove it.

packages/dapp/src/components/MultisigConfirmationModal/MultisigConfirmationModal.tsx
Abhikumar98: okay

packages/shared/src/types/settings.ts
Abhikumar98: got it. I'll move it

packages/dapp/src/const/MenuConfig/useWalletSelectContent.tsx
```diff
@@ -34,43 +50,68 @@ export const useWalletSelectContent = () => {
...
+        // separate into installed and not installed wallets
+        const installedWallets = supportedWallets.filter(
+          (_, index) => installed[index],
+        );
+        // always remove Default Wallet from not installed Wallets
+        const notInstalledWallets = supportedWallets.filter(
+          (wallet, index) =>
+            !installed[index] && wallet.name !== 'Default Wallet',
+        );
+        availableWallets = [...installedWallets, ...notInstalledWallets];
```
chybisov, line 71: Are you sure this works fine inside useMemo? We will exit the function before `availableWallets` can be set because Promise.all is async. useEffect maybe?

packages/dapp/src/stores/multisig/MultisigStore.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { StateCreator } from 'zustand';
+import { create } from 'zustand';
+import { persist } from 'zustand/middleware';
+
+// ----------------------------------------------------------------------
+
+/*--  Use Zustand  --*/
+
+export const useMultisigStore = create(
+  persist(
```
chybisov, line 11: Do we want it to be persistent?

packages/dapp/src/stores/multisig/MultisigStore.ts
```diff
@@ -0,0 +1,40 @@
+import { MultisigProps, MultisigState } from '@transferto/shared';
+import type { StateCreator } from 'zustand';
+import { create } from 'zustand';
+import { persist } from 'zustand/middleware';
+
+// ----------------------------------------------------------------------
+
+/*--  Use Zustand  --*/
+
```
chybisov, line 9: Useless comments 😅

packages/shared/src/types/multisig.ts
```diff
@@ -0,0 +1,9 @@
+import type { ChainId } from '@lifi/types';
+
+export interface MultisigProps {
+  destinationChain: ChainId;
+}
+export interface MultisigState extends MultisigProps {
+  // Destination Chain
+  onDestinationChainSelected: (chainId: number) => void;
```
chybisov, line 8: Maybe we should place these types closer to the store?

chybisov (changes requested):

packages/shared/src/types/multisig.ts
Abhikumar98: you are correct, but I'm following the existing structure followed by Menu Store and settings store

packages/dapp/src/stores/multisig/MultisigStore.ts
Abhikumar98: actually, it's not rquired. removing it

chybisov (approved):

Addminus (commented):
i have a question:

why are there so many merge commits a t the top?

Addminus (approved):

Abhikumar98:
> i have a question:
> 
> why are there so many merge commits a t the top?

I might've rebased with main bymistake sometime

PR #569, opened 2023-06-20
refactor: replace swap with exchange terminology

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-3465


chybisov (approved):

dennyscode:
@chybisov We still have "Swap" in place because of our translations. I am not sure what the best way will be to get this cleaned up as well. Maybe it does not need to? What do you think?

image

chybisov:
@dennyscode, we need to introduce a new key `exchange` and later on crowdin will remove the `swap` one.

chybisov (approved):

PR #583, opened 2023-06-28
bug: close welcomeScreen on wallet selection

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-3573

Fix: The Flag "welcomeScreenEntered" will now be set to true once a wallet is selected. As soon as the WalletConnect opens, the welcome screen disappears now.

[close-welcome-screen-on-wallet-selection-screen-capture.webm](https://github.com/lifinance/jumper.exchange/assets/43956540/137e5e48-8a9f-4204-85c2-e157d5f690b3)

### DIFF
--- a/packages/dapp/src/components/Navbar/Navbar.tsx
+++ b/packages/dapp/src/components/Navbar/Navbar.tsx
@@ -1,6 +1,5 @@
 import { useTheme } from '@mui/material/styles';
 import { BrandLogo } from '@transferto/shared/src/atoms/illustrations';
-import { shallow } from 'zustand/shallow';
 import { useWallet } from '../../providers/WalletProvider';
 import { useSettingsStore } from '../../stores';
 import { NavbarBrandContainer, NavbarContainer } from './Navbar.style';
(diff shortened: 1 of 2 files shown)


chybisov:
@dennyscode can we adjust the header z-index so it doesn't overlap WalletConnect?

dennyscode:
> @dennyscode can we adjust the header z-index so it doesn't overlap WalletConnect?

Great idea, let me see! I also had that thought

dennyscode:
@chybisov: Please check the short video for more infos
[close-welcome-screen-on-wallet-selection-screen-capture.webm](https://github.com/lifinance/jumper.exchange/assets/43956540/72d82572-691b-4100-a524-fded49546103)


1.) I just tried to fix the overlay of the navbar but I am curious on which approach you´d be taking: It could be solved with a.) zIndex or b.) padding-top. Which is the way to go and how to pass it to the wallet-connect iframe?

2.) I noticed that multiple wallet connect instances are piling up in the DOM. Could that lead to any issues and should be fixed?

chybisov:
@dennyscode yeah, I see... I'll handle both 🙂

chybisov (approved):

dennyscode:
Our two ideas seem to interfere and I am thinking on how to combine both:

1.) Should we tie this case only to the scenario of a user selecting "WalletConnect"? Currently, welcomeScreenEntered is set to true once any wallet is selected, meaning the WelcomeScreen will be hidden.

2.) In this ticket, we wanted to show the welcomePage, also if a user has already connected his wallet: https://lifi.atlassian.net/browse/LF-3572

By adding the edge-case only for WalletConnect, we could potentially achieve both.

PR #584, opened 2023-06-28
chore: update chains for gas feature

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-3497


packages/dapp/src/const/tabsMap.ts
```diff
@@ -0,0 +1,15 @@
...
+  [key: string]: {
+    index: number;
+    value: WidgetSubvariant | 'buy';
+  };
+}
+
+export const TabsMap: TabsMapType = {
+  Exchange: { index: 0, value: 'default' },
+  Gas: { index: 1, value: 'refuel' },
+  Refuel: { index: 1, value: 'refuel' },
```
chybisov, line 13: Why do we have gas and refuel? Maybe we should select one and proceed with it 🙂

chybisov (approved):

packages/dapp/src/const/tabsMap.ts
dennyscode: Makes sense! Good find, done

PR #597, opened 2023-07-07
Lf 3567 move active tabs logic out from jumper store

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-3567


dennyscode:
@chybisov I notice that the old junk is still in the localStorage. This might not harm and just sit, but is there a proper way to avoid piles of old junk or just cleaning up?

image

chybisov:
@dennyscode yes, you can increment the version of the store and provide a migration function in which you will clean that junk 🙂 Take a look at widget stores, there are some migrations.

Addminus (changes requested):
TODO: Implement @chybisov proposition in this PR

dennyscode:
Still WIP: Wanted to delete keys from store which is not part of defaultSettings defined for the settings. This way, if keys in localStorage do not match with the ones of defaultSetting, they should be removed. Unfortunately, not working just yet

Addminus (approved):

chybisov (approved):

PR #599, opened 2023-07-12
feat: add raleon (LF-3486)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-3486


Addminus (changes requested):
I would adjust the things we track with raleon:
- only the current tab
- chain switches
- wallet connects and disconnects
- transaction stuff

Addminus:
image

Addminus:
Raleon data is coming into the dashboard

chybisov (changes requested):
There is no strict naming convention for enums in TS, but there is one for C#, which inspired enums in TS. Please, let's use [PascalCasing](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces#:~:text=DO%20name%20classes%20and%20structs%20with%20nouns%20or%20noun%20phrases%2C%20using%20PascalCasing.) 🙂

Addminus:
we can do a general refactor of this pascal case thing. it does not affect the tracking functionality of raelon

chybisov (approved):

PR #601, opened 2023-07-13
style: adjustments on welcome screen (LF-3750)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-3750

image

- adjusted size of stats-cards for mobile and desktop
- removed the margin top of the title´s (dexs / bridges / chains) on statscard on mobile, left it on large screens
- distance between welcome-wrapper content text "find the best route" + 4x audited....", now has marginTop 16px across all screen sizes
- removed textDecorations (underlines) for the links
- have the modal stand out more on larger screens, adjusted marginTop calc


packages/dapp/src/components/Widgets/Widgets.style.tsx
```diff
@@ -84,6 +84,10 @@ export const WidgetContainer = styled(Box, {
     [`@media screen and (min-height: 700px)`]: {
       marginTop: showWelcome ? 'calc( 50vh - 680px / 2.75 - 80px)' : '0', // (mid viewheight - half-two/thirds widget height - header height )
     },
+
+    [`@media screen and (min-height: 900px)`]: {
```
chybisov, line 88: Should we use standard breakpoint numbers?
https://mui.com/material-ui/customization/breakpoints/

chybisov (changes requested):

packages/dapp/src/components/Widgets/Widgets.style.tsx
dennyscode: @chybisov I agree! Good, you´re asking - but how does that apply if looking at min-heights? In this approach, I am positioning the widget for the welcome-screen because of limited vertical space. I´d also prefer using as little of those "custom" breakpoints as possible.

chybisov: Oh, didn't notice that's about the height 😄

chybisov (approved):

chybisov (approved):

chybisov (approved):

Addminus (changes requested):
I have a blue border around the modals in this PR:
image

also have it on firefox

image

dennyscode:
Looked even funnier on my end, wondering how I missed that all before, nobody else complaiining! But it turns out the autoFocus or focus-visible pseudo-class interfered here! Now fixed ✅

image

Addminus (approved):

chybisov (approved):

PR #602, opened 2023-07-17
chore: set tsconfig strict to true (LF-3472)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-3472


dennyscode:
Current state, down from 89 errors:

image

dennyscode:
Last todo: Wait for the updated sdk package and upgrade to resolve the following issue:

image

dennyscode:
Noticed, it´s really slow after merging develop into current branch.
Check in useWalletSelectContent.tsx:

image

dennyscode:
@chybisov Also in here the following Migration:

image

packages/dapp/src/components/MultisigConfirmationModal/MultisigConfirmationModal.tsx
```diff
@@ -30,10 +30,10 @@ export const MultisigConfirmationModal: React.FC<{
             fontSize: '1.125rem',
           }}
         >
-          {translate(`${i18Path}.title`)}
+          {translate(`${I18_PATH}title`, 'translation')}
```
dennyscode, line 33: @Addminus  @chybisov Also, please take note on this usage of the translate-hook. Been using this now all over the place to fix the warnings on TemplateStringArrays. This seems to do the trick but I´d like the have you check since that´s quite some changes across all of our files.

packages/dapp/src/components/MultisigWalletHeaderAlert/MultisigWalletHeaderAlert.tsx
```diff
@@ -10,17 +10,18 @@ import {
...
   const { t: translate } = useTranslation();
+  const I18_PATH = 'multisig.alert.';
 
   return (
     <MultisigWalletHeaderAlertContainer>
       <MultisigWalletHeaderAlertTitle>
         <InfoRoundedIcon />
         <Typography fontWeight={700} marginLeft={1}>
-          {translate(`multisig.alert.title`)}
+          {translate(`${I18_PATH}multisig.alert.title`, 'translation')}
```
chybisov, line 20: Are we sure that the full path is `multisig.alert.multisig.alert.title`? 🙂

packages/dapp/src/components/MultisigConfirmationModal/MultisigConfirmationModal.tsx
chybisov: Also, I'm not sure why we started to use i18n like this

```ts
const { t: translate } = useTranslation();
```
but it would be cool to have it as default `t` everywhere. Fewer letters to type, and looks compact. 🙂

```ts
```

packages/dapp/src/components/Navbar/Menu/ChainsMenu/ChainsMenu.tsx
```diff
@@ -16,25 +15,19 @@ interface NavbarMenuProps {
...
-      (state) => [
-        state.openNavbarChainsMenu,
-        state.onOpenNavbarChainsMenu,
-        state.onOpenNavbarSubMenu,
-      ],
-      shallow,
-    );
+  const [openNavbarChainsMenu, onOpenNavbarChainsMenu] = useMenuStore(
+    (state) => [state.openNavbarChainsMenu, state.onOpenNavbarChainsMenu],
+  );
```
chybisov, line 25: Why did we remove `shallow` from here?

packages/dapp/src/components/Navbar/Menu/ChainsMenu/ChainsMenu.tsx
```diff
@@ -44,13 +37,15 @@ export const ChainsMenu = ({ handleClose, open }: NavbarMenuProps) => {
...
             key={`${el.label}-${index}`}
             label={el.label}
-            showButton={el.showButton}
+            showButton={el.showButton ? el.showButton : false}
             showMoreIcon={false}
-            suffixIcon={el.chainId === account.chainId && <CheckIcon />}
+            suffixIcon={
+              (el.chainId && el.chainId === account.chainId && <CheckIcon />) ||
+              undefined
+            }
```
chybisov, line 45: Let's make it more reactish way ->

```ts
el.chainId && el.chainId === account.chainId ? <CheckIcon /> : null
```

packages/dapp/src/components/Navbar/Menu/MainMenu/MainMenu.tsx
```diff
@@ -53,40 +44,39 @@ export const MainMenu = ({ handleClose }: MainMenuProps) => {
             label={el.label}
             prefixIcon={el.prefixIcon}
             triggerSubMenu={el.triggerSubMenu}
-            showButton={el.showButton}
+            showButton={el.showButton || false}
```
chybisov, line 47: Is there another solution? 🙂

packages/dapp/src/components/Navbar/Menu/WalletMenu/WalletMenu.tsx
```diff
@@ -83,25 +77,28 @@ export const WalletMenu = ({ handleClose }: NavbarMenuProps) => {
...
+          }address/${account.address}`
+        : '',
       pageload: true,
       disableTrackingTool: [EventTrackingTool.ARCx],
     });
   };
 
   const handleCopyButton = () => {
-    navigator?.clipboard?.writeText(account.address);
+    account.address && navigator?.clipboard?.writeText(account.address);
```
chybisov, line 101: How about

```ts
if (!account.address) {
  return
}
navigator?.clipboard?.writeText(account.address);
```
or
```ts
if (account.address) {
  navigator?.clipboard?.writeText(account.address);
}
```

packages/dapp/src/components/Navbar/MenuItemComponent.tsx
```diff
@@ -35,14 +33,17 @@ const MenuItemComponent = ({
...
 
   return open ? (
     <MenuItem
       disableRipple={showButton}
-      showButton={showButton}
+      showButton={showButton || false}
       autoFocus={autoFocus}
       onClick={() => {
-        !!triggerSubMenu && setOpenSubMenu(triggerSubMenu);
+        !!triggerSubMenu && onOpenNavbarSubMenu(triggerSubMenu);
```
chybisov, line 46: Can we use one condition for both cases?

packages/dapp/src/components/Widgets/Widgets.tsx
```diff
@@ -12,28 +17,31 @@ import { Widget } from '../Widget';
 import { WidgetEvents } from './WidgetEvents';
 import { WidgetContainer } from './Widgets.style';
 
+export type StarterVariantType = 'buy' | WidgetSubvariant;
```
chybisov, line 20: We have kind of related interface here `packages/dapp/src/components/Widget/Widget.tsx`. Can we combine them for reusability?

packages/dapp/src/components/Widgets/Widgets.tsx
```diff
@@ -81,38 +89,40 @@ export function Widgets() {
...
-      </WidgetContainer>
-      <WidgetContainer
-        onClick={handleGetStarted}
-        showWelcome={!welcomeScreenEntered}
-        isActive={_starterVariant === TabsMap.Refuel.value}
-      >
-        <Widget starterVariant={TabsMap.Refuel.value as WidgetSubvariant} />
-      </WidgetContainer>
-      {import.meta.env.VITE_ONRAMPER_ENABLED ? (
+      <>
```
chybisov, line 92: Do we need this here? 🙂

packages/dapp/src/hooks/useMultisig.ts
```diff
@@ -31,14 +36,14 @@ export const useMultisig = () => {
     return !!accountInfo?.safeAddress;
   };
 
-  const isSafeSigner = !!(account?.signer?.provider as any)?.provider?.safe
-    ?.safeAddress;
+  const isSafeSigner =
+    !!(account?.signer?.provider as any)?.provider?.safe?.safeAddress || false;
```
chybisov, line 40: Will it infer the type if we make it like this? Looks cleaner this way 🙂

`Boolean((account?.signer?.provider as any)?.provider?.safe?.safeAddress)`

packages/dapp/src/stores/menu/MenuStore.tsx
```diff
@@ -22,88 +23,91 @@ export const defaultMenu: DefaultMenuType = {
...
-    set(() => ({
-      [key]: value,
-    })),
-  setValues: (values) =>
-    set((state) => {
-      const updatedState: MenuProps = { ...state };
-      for (const key in values) {
-        if (Object.hasOwn(state, key)) {
-          updatedState[key] = values[key];
+export const useMenuStore = createWithEqualityFn<MenuState>(
```
chybisov, line 26: To have a previous behavior it is suggested to change it to `createWithEqualityFn(..., Object.is)`. If we make it `createWithEqualityFn(..., shallow)` then every usage of this hook will use `shallow` for comparison and I'm not sure we want it that way. Do we? 🤔

packages/dapp/src/types/resourceKey.ts
```diff
@@ -0,0 +1,64 @@
+import { ResourceKey as I18nResourceKey } from 'i18next';
+
+interface ExtendedResourceKey {
```
chybisov, line 3: Wow, I don't think we need these types 🙂 Let's discuss.

packages/shared/src/atoms/illustrations/BrandLogo.tsx
```diff
@@ -1,12 +1,17 @@
...
+
+const JumperLogo = ({ theme }: Props) => {
   const mainCol =
     theme.palette.mode === 'light'
-      ? theme.palette.accent1.main
-      : theme.palette.accent1Alt.main;
-  const subCol = theme.palette.accent2.main;
+      ? theme?.palette.accent1.main
+      : theme?.palette.accent1Alt.main;
+  const subCol = theme?.palette.accent2.main;
```
chybisov, line 14: If it is required in types, why do we need to check for undefined?

chybisov (changes requested):

packages/dapp/src/components/MultisigWalletHeaderAlert/MultisigWalletHeaderAlert.tsx
dennyscode: Oh, oh.. That must be wrong! Fixed now

packages/dapp/src/components/Navbar/Menu/ChainsMenu/ChainsMenu.tsx
dennyscode: Thanks! Applied your code but replaced the "null" with `undefined`

packages/dapp/src/components/Navbar/Menu/MainMenu/MainMenu.tsx
dennyscode: What do you think about replacing it with this:
`showButton={el.showButton ?? false}`

packages/dapp/src/components/Widgets/Widgets.tsx
dennyscode: Maybe not? Maybe, it´s the same as with those TemplateStringArrays which can be resolved by adjustments in the typescript config? For me, it resolved the following error for the moment:

image

packages/dapp/src/hooks/useMultisig.ts
dennyscode: Much better, thanks! Never used that approach, great you brought it up!

packages/dapp/src/stores/menu/MenuStore.tsx
dennyscode: Well, okay.. This was the reason I replaced those shallow´s on all those items. I definitely left it out for our Zustand store on "activeTabs" which really doesn´t need to be set to "shallow". But what´s with the others? Let´s discuss and see which works better

packages/dapp/src/types/resourceKey.ts
dennyscode: Yes, seeing it now it feels like this is just a pain in the a** to keep that manually up to date and should be removed.

packages/shared/src/atoms/illustrations/BrandLogo.tsx
dennyscode: To be fricking super safe :D 
No, just kidding and got rid of it! Nice find, I happened to use it all across as I started to believe it makes my life with typescipt easier.. but you are completely right: It´s already required and there´s no need for it right here!!

packages/dapp/src/components/Navbar/Menu/ChainsMenu/ChainsMenu.tsx
dennyscode: Because I added the "shallow" within the createWithEquality function which replaces it right here? Correct me, if I am wrong - we need to talk about all that "shallow"s anyways

packages/dapp/src/components/Navbar/Menu/MainMenu/MainMenu.tsx
dennyscode: It´s complaining otherwise if I just pass it like so:
`showButton={el.showButton}
`
image

dennyscode: Or shall I define the type to be `boolean | undefined`  to make it cleaner?

dennyscode:
- [ ] WelcomeWrapper.style.ts --> animation fadeOut
- [ ] Disconnect Wallet: Gotta press twice
- [ ] Template String Array - issue
- [ ] /Widgets.tsx --> <> ... </>

dennyscode:
Feels slow on switching tabs (exchange, gas, buy), gotta test

packages/dapp/src/components/Widgets/Widgets.tsx
dennyscode: It was resolved with `React.FC<PropsWithChildren<{}>>`

dennyscode:
@chybisov 

- touched the tsconfig now
- cleaned up optional chaining when not needed
- re-usable "starterVariant" type
- commented "disabled lines" linter
- ...

Addminus (approved):
Oi its a lot to review. 
I have no problems highlited in vs code. The mentioned package update seemed to work.
Mainly checked if these changes break something. 
the ticket mentionned the use of "any". We tackle this in another ticket as this one is so big?

packages/dapp/src/components/Navbar/Menu/WalletMenu/WalletMenu.tsx
```diff
@@ -113,11 +113,13 @@ export const WalletMenu = ({ handleClose }: NavbarMenuProps) => {
...
 
-  const handleMultisigEnvironmentCheck = async () => {
+  const handleMultisigEnvironmentCheck = useCallback(async () => {
     const response = await checkMultisigEnvironment();
 
     setIsMultisigEnvironment(response);
-  };
+    // Check MultisigEnvironment only on first render
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
dennyscode, line 122: @Abhikumar98 Could you please take a look and maybe test the multisig-wallet check? Thanks ahead

Abhikumar98: Hey @dennyscode,

This is working as expected and tbh, I don't think `useCallback` is required here. In both scenarios, with and without `useCallback` the code is working fine, although I'd suggest to revert back to using without it unless there is an issue being thrown.

chybisov (approved):

PR #603, opened 2023-07-17
chore: bump outdated dependencies to new major versions (LF-3650)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-3650


packages/dapp/package.json
```diff
@@ -20,13 +20,13 @@
     "@transferto/shared": "*",
     "@types/react": "^18.2.15",
     "@widgetbot/react-embed": "^1.7.0",
-    "ethers": "^5.7.2",
+    "ethers": "^6.6.2",
```
chybisov, line 23: Should have mentioned this in the task, but we don't need bump ethers. 🙂

chybisov (changes requested):

packages/shared/src/style/reset.style.ts
```diff
@@ -1,7 +1,4 @@
-import { normalize } from 'polished';
-
 export const resetStyle = `
-${normalize()};
```
dennyscode, line 4: @chybisov I just removed this. Do you see any issues here?

packages/shared/package.json
```diff
@@ -14,40 +14,30 @@
...
+    "@lifi/wallet-management": "^2.2.4",
+    "@mui/icons-material": "^5.14.3",
+    "@mui/material": "^5.14.3",
+    "i18next": "^22.5.1",
     "js-cookie": "^3.0.5",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-i18next": "^12.3.1",
     "typescript": "^5.1.6",
     "use-media": "^1.4.0",
```
chybisov, line 40: I guess we can remove `js-cookie` and `use-media`

packages/shared/src/style/reset.style.ts
chybisov: Since nothing from resetStyle is used I'd remove the whole style folder

packages/dapp/package.json
```diff
@@ -7,44 +7,44 @@
...
     "react-dom": "^18.2.0",
     "react-ga4": "^2.1.0",
-    "react-hotjar": "^5.5.0",
+    "react-hotjar": "^6.1.0",
     "react-i18next": "^12.3.1",
     "react-query": "^3.39.3",
-    "react-router-dom": "^6.14.1",
+    "react-router-dom": "^6.14.2",
     "typescript": "^5.1.6",
     "use-media": "^1.4.0",
```
chybisov, line 34: `use-media` can be deleted from here as well

chybisov (changes requested):

packages/shared/package.json
dennyscode: Good find, removed!

packages/dapp/package.json
dennyscode: Nice, use-media is removed as well now

packages/shared/src/style/reset.style.ts
dennyscode: Right, much cleaner now

packages/dapp/package.json
dennyscode: Alright, left it as 5.7.2

chybisov (changes requested):
Looks good 🙂

Can we please also move not used devDependencies from shared/dapp package.json to the root one? e.g. lerna, prettier etc. We will have 1-2 probably at most there after this move 🙂

packages/shared/package.json
```diff
@@ -17,14 +17,6 @@
...
     "start": "tsc --build --clean && tsc --build && tsc --build tsconfig.json"
   },
-  "devDependencies": {
-    "@babel/core": "^7.22.10",
-    "@types/node": "^20.4.8",
-    "@types/react": "^18.2.19",
-    "babel-loader": "^9.1.3",
-    "eslint-config-prettier": "^8.10.0",
-    "prettier": "^3.0.1"
-  },
```
dennyscode: @chybisov Just removed that complete devDependenies-stack from the shared package.json. It works just fine locally, do you see issues with that? (As you pointed out, we might be left with just a couple ones but here we got rid of it all)

chybisov: @dennyscode, that's fine. Can you please try to remove some from dapp folder?

package.json
```diff
@@ -4,7 +4,11 @@
   "version": "1.0.0",
   "repository": "https://github.com/lifinance/jumper.exchange",
   "devDependencies": {
+    "@babel/core": "^7.22.10",
```
chybisov, line 7: I guess we don't need babel anymore since we use Vite 🙂

packages/shared/package.json
dennyscode: I got to remove the "emotion" packages from dapp and shared folder, too. Not sure, if there´s more to be removed but that´s where we are now.

packages/dapp/package.json
```diff
@@ -5,46 +5,43 @@
...
     "@types/react-dom": "^18.2.7",
-    "@vitejs/plugin-react": "^4.0.3",
+    "@vitejs/plugin-react": "^4.0.4",
     "husky": "^8.0.3",
-    "lerna": "^7.1.3",
+    "lerna": "^7.1.4",
     "rollup-plugin-polyfill-node": "^0.12.0",
-    "vite": "^4.4.4",
+    "vite": "^4.4.9",
     "web-vitals": "^3.4.0"
```
chybisov, line 45: We should also move lerna, husky, types to the root-level package.json

chybisov (approved):

Abhikumar98 (approved):
LGTM

PR #617, opened 2023-08-11
style: fix blurry snackbar (LF-4459)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-4459

Fixed blurry Snackbar, because it was placed too close to Navbar.
Solution: Move it down 2px

image


Abhikumar98 (approved):

chybisov (approved):
TIL we have a snackbar 👀😄

PR #619, opened 2023-08-11
chore: implemented dynamic loading of translations

Abhikumar98:


package.json
```diff
@@ -33,5 +33,9 @@
   "workspaces": [
     "packages/*"
   ],
-  "packageManager": "yarn@3.6.1"
+  "packageManager": "yarn@3.6.1",
+  "dependencies": {
```
Abhikumar98, line 37: this will be moved to respective package

Abhikumar98:
> 

Sure, adding it

Abhikumar98:
> Please check this out. When we load the page there is no fallback and we see plain keys. Can we make english as fallback language that is always there?
> 
>  Screen.Recording.2023-08-14.at.11.43.37.AM.mov

Hey @chybisov,
`en` is already set as the fallback language here, but since we are importing all the languages dynamically it takes sometime to load the fallback as well, here the fallback is working as when there is no language selected.

Although in our case we want to have a "pre-loaded" language, which will be already imported and loaded prior to lazy loading of languages and this isn't working as expected:

I've tried:
1. Importing the en.json file but to no avail, as it still picks the file loaded from the backend resource
2. Supplying just the fallback language to resource, expecting it to later "lazily" add more languages to resources but this isn't working as expected as well.

But this latency in languages being loaded is expected imo, but lemme know if there's a possible solution that I've missed out.

Thanks !

Abhikumar98:
> > Please check this out. When we load the page there is no fallback and we see plain keys. Can we make english as fallback language that is always there?
> > Screen.Recording.2023-08-14.at.11.43.37.AM.mov
> 
> Hey @chybisov, `en` is already set as the fallback language here, but since we are importing all the languages dynamically it takes sometime to load the fallback as well, here the fallback is working as when there is no language selected.
> 
> Although in our case we want to have a "pre-loaded" language, which will be already imported and loaded prior to lazy loading of languages and this isn't working as expected:
> 
> I've tried:
> 
> 1. Importing the en.json file but to no avail, as it still picks the file loaded from the backend resource
> 2. Supplying just the fallback language to resource, expecting it to later "lazily" add more languages to resources but this isn't working as expected as well.
> 
> But this latency in languages being loaded is expected imo, but lemme know if there's a possible solution that I've missed out.
> 
> Thanks !

Update:

Tried a config of partially loading the resources from backend and using static ones, seems to be working for now.

packages/dapp/src/providers/I18nProvider.tsx
```diff
@@ -5,10 +5,19 @@ import React, { PropsWithChildren, useMemo } from 'react';
...
 export const I18NProvider: React.FC<PropsWithChildren<{}>> = ({ children }) => {
   const languageMode = useSettingsStore((state) => state.languageMode);
 
+  const resources = {
+    en: {
+      translation: {
+        ...en,
+      },
+    },
+  };
```
chybisov: I think this works fine as well

```ts
const resources = {
    en: {
      translation: en,
    },
  };
```

Or we can even make it inline in the init function below

packages/dapp/src/const/MenuConfig/useLanguagesContent.tsx
```diff
@@ -23,12 +24,14 @@ export const useLanguagesContent = () => {
     });
   };
 
-  const languages = Object.keys(i18n.store.data)
+  console.log({ data: i18n.store.data });
```
chybisov, line 27: 🙂

chybisov (changes requested):

chybisov:
@Abhikumar98, the last update is really nice and works well. Please see a couple of comments above 🙂

After we merge this I'll update crowdin configuration to a new file structure and we can proceed with updating widget with the same config 🙂

Abhikumar98:
hey @chybisov,
All the changes mentioned above are done, let's have a last review and merge if it's good to go :)

chybisov (approved):

packages/dapp/package.json
```diff
@@ -20,7 +20,9 @@
     "@widgetbot/react-embed": "^1.8.0",
     "ethers": "^5.7.2",
     "i18next": "^23.4.4",
-    "i18next-browser-languagedetector": "^7.1.0",
+    "i18next-browser-languagedetector": "^7.0.2",
```
dennyscode, line 23: Was it necessary to downgrade this package? So, should we keep this in mind when bumping our packages in order to not break anything?

dennyscode (approved):

packages/dapp/package.json
chybisov: @dennyscode, nice catch. No, this shouldn't be downgraded 🙂

chybisov (changes requested):
Please upgrade i18next packages.

chybisov (approved):

PR #620, opened 2023-08-15
fix: allow multiple urls for tab opening (LF-4488)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-4488

Problem: Due to refactoring the url `jumper.exchange/gas` did not open the associated "Gas"-Widget. It only loaded on `jumper.exchange/refuel` because the property "WidgetSubVariant" and the url-names don´t match. E.g. "Swap" refers to widgetVariant "default" and "Gas" is referring to "refuel".

Solution: With this change, we can not only define specific url but allow multiple url´s to open certain Tabs.

### DIFF
--- a/packages/dapp/src/components/Widgets/Widgets.tsx
+++ b/packages/dapp/src/components/Widgets/Widgets.tsx
@@ -41,7 +41,7 @@ export function Widgets() {
         return TabsMap.Exchange.variant;
       }
     } else {
-      // default: Exchange-Tab
+      // default and fallback: Exchange-Tab
       return TabsMap.Exchange.variant;
     }
   }, []);
--- a/packages/dapp/src/components/Widgets/Widgets.tsx
+++ b/packages/dapp/src/components/Widgets/Widgets.tsx
@@ -26,43 +26,44 @@ export function Widgets() {
   const [starterVariantUsed, setStarterVariantUsed] = useState(false);
   const [_starterVariant, setStarterVariant] = useState<
     WidgetSubvariant | 'buy'
-  >(TabsMap.Exchange.value);
+  >(TabsMap.Exchange.variant);
 
   const starterVariant: StarterVariantType = useMemo(() => {
     let url = window.location.pathname.slice(1);
     if (Object.values(LinkMap).includes(url as LinkMap)) {
       if (!!TabsMap.Buy.destination.filter((el) => el === url).length) {
-        return TabsMap.Buy.value;
+        return TabsMap.Buy.variant;
       } else if (
         !!TabsMap.Refuel.destination.filter((el) => el === url).length
       ) {
-        return TabsMap.Refuel.value;
+        return TabsMap.Refuel.variant;
       } else {
-        return TabsMap.Exchange.value;
+        return TabsMap.Exchange.variant;
       }
     } else {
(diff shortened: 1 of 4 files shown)


packages/dapp/src/components/Widgets/Widgets.tsx
```diff
@@ -26,41 +26,44 @@ export function Widgets() {
...
-      return TabsMap.Exchange.value;
+      // default and fallback: Exchange-Tab
+      return TabsMap.Exchange.variant;
     }
   }, []);
 
   const getActiveWidget = useCallback(() => {
     if (!starterVariantUsed) {
-      starterVariant === TabsMap.Exchange.value
+      starterVariant === TabsMap.Exchange.variant
```
Abhikumar98, line 51: can we convert this to a switch case logic instead of chained ternary?


```
switch(starterVariant) {
    case TabsMap.Exchange.variant:
        setActiveTab(TabsMap.Exchange.index)
    .
    .
    .
}
```

packages/dapp/src/components/Widgets/Widgets.tsx
```diff
@@ -26,41 +26,44 @@ export function Widgets() {
...
-        : starterVariant === TabsMap.Buy.value
+        : starterVariant === TabsMap.Buy.variant
         ? setActiveTab(TabsMap.Buy.index)
         : setActiveTab(TabsMap.Exchange.index);
       setStarterVariant(starterVariant);
       setStarterVariantUsed(true);
     } else {
       if (activeTab === TabsMap.Exchange.index) {
-        setStarterVariant(TabsMap.Exchange.value);
+        setStarterVariant(TabsMap.Exchange.variant);
```
Abhikumar98, line 62: or this also works^

Abhikumar98 (approved):

packages/dapp/src/components/Widgets/Widgets.tsx
```diff
@@ -26,41 +26,58 @@ export function Widgets() {
...
+          break;
+        case TabsMap.Refuel.index:
+          setStarterVariant(TabsMap.Refuel.variant);
+          break;
+        case TabsMap.Buy.index:
+          setStarterVariant(TabsMap.Buy.variant);
+          break;
+        default:
+          setStarterVariant(TabsMap.Exchange.variant);
+          break;
```
Abhikumar98, line 80: break isn't really needed here

packages/dapp/src/components/Widgets/Widgets.tsx
```diff
@@ -26,41 +26,58 @@ export function Widgets() {
...
+        case TabsMap.Refuel.variant:
+          setActiveTab(TabsMap.Refuel.index);
+          break;
+        case TabsMap.Buy.variant:
+          setActiveTab(TabsMap.Buy.index);
+          break;
+        default:
+          setActiveTab(TabsMap.Exchange.index);
+          break;
```
Abhikumar98, line 63: break isn't really needed here

dennyscode: Makes sense! Thanks @Abhikumar98

0xsebn (approved):

PR #630, opened 2023-08-26
fix: making pwa for jumper.exchange

Abhikumar98:


packages/dapp/public/favicon-144x144.png
chybisov, line 1: Can we find a better icon in terms of resolution?

chybisov (dismissed):

packages/dapp/public/favicon-144x144.png
Abhikumar98: Hey,
Yes definitely we can pick better assets. I was just experimenting with this hence used any placeholder assets.

I'll be using the correct ones in sometime

Abhikumar98:
Closing this because it has didn't show the expected behaviour on iOS

packages/dapp/public/favicon-144x144.png
chybisov: @Abhikumar98, it still looks very blurry with low resolution. Let's use SVG.

chybisov (changes requested):
Do we still need other icons if we have SVG? I assume SVG can cover all resolutions

Abhikumar98:
> Do we still need other icons if we have SVG? I assume SVG can cover all resolutions
CleanShot 2023-10-04 at 19 43 39

We can use it for all sizes, but majorly went for this to get rid of the warning.
If this isn't causing a significant difference in package size, I thought of get rid of these warnings.

chybisov (approved):

chybisov:
@Abhikumar98 okay, let's use at least not blurry png please 🙂

chybisov (approved):

Addminus (approved):

PR #631, opened 2023-08-29
feat: improve ga tracking (LF-3179) 

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-3179

- improved usage of ENV variables of ga-id - now differentiates between staging and production
- added user-properties
- cleaned up TrackingKeys by adding: TrackingUserProperties and TrackingEventParameters to differentiate between Custom Dimensions/Metrics and User-Properties
- added tracking for welcome-screen, feature-cards and multiple more such as user_properties resulting from certain widget-events (like: hadSuccessfulTxx, hadFailure, hadAcceptedHighValueLoss -> that will be passed on as user-attributes)


chybisov:
@dennyscode 😆
image

dennyscode:
Unfortunately, using different naming-conventions which is not adding to the consistence of tracked data:

- preferring kebab-cases for most keys
- in past, been using camelCase for the transaction events the way it is set up in our api: Throughout the past months we had been tracking this way and at this point, I am leaving it as it is to not screw up our data
- for some reason, user_properties cannot be set as custom dimensions with kebab-casing, that´s why I got with snake-casing

-> that´s how we end up with 3 different naming conventions. Guess that moving it all to snake-cases would work, though.

image

packages/dapp/src/components/FeatureCards/FeatureCard.tsx
```diff
@@ -72,6 +80,21 @@ export const FeatureCard = ({ data, isSuccess, assets }: FeatureCardProps) => {
...
+                  [TrackingEventParameters.FeatureCardTitle]:
+                    data?.fields?.title,
+                  [TrackingEventParameters.FeatureCardId]:
+                    data?.fields?.displayConditions?.id,
+                },
+                disableTrackingTool: [
+                  EventTrackingTool.ARCx,
+                  EventTrackingTool.Raleon,
+                ],
+              });
```
chybisov, line 97: Can we please extract these inline functions to some handlers at the top of the component? It becomes hard to maintain and generally considered a bad practice 🙂

chybisov (changes requested):

packages/dapp/src/providers/I18nProvider.tsx
```diff
@@ -42,6 +45,12 @@ export const I18NProvider: React.FC<PropsWithChildren<{}>> = ({ children }) => {
     });
 
     if (!languageMode) {
+      trackAttribute({
+        data: {
+          [TrackingUserProperties.Language]: i18n.language,
+          [TrackingUserProperties.DefaultLanguage]: i18n.language,
+        },
+      });
```
chybisov, line 53: What do we want to track here - language change or user language on page load? This should be tested carefully.

chybisov (changes requested):

packages/dapp/src/providers/I18nProvider.tsx
dennyscode: My idea was: In order to track languages, it´d be interesting to know:
- default language on pageload
- active language after it´s been changed

So, if there is no language set in localStorage, it should be hitting that function and save both values as user_properties.
The default language shall not change ever, but active language will be modified once another language has been chosen by the user (in the menu).

Yes, this needs to be tested.

dennyscode: I am also updating the Notion documentation to keep everything structured and list events / user_properties / custom dimensions

packages/dapp/src/components/FeatureCards/FeatureCard.tsx
dennyscode: Good point! Thank you, I found a few of those and refactored

dennyscode (commented):
@Addminus LGTM

packages/dapp/src/const/trackingKeys.ts
```diff
@@ -1,33 +1,108 @@
+/*
+  Name of an action that a user makes on the site. 
+  GA: in GA this is the name of the even that is fired
+*/
+export enum TrackingActions {
```
chybisov, line 5: By convention, we should use a singular type name for an enumeration unless its values are bit fields. 🙂

TrackingActions -> TrackingAction
TrackingCategories -> TrackingCategory
TrackingEventParameters -> TrackingEventParameter

packages/dapp/src/types/userTracking.types.ts
```diff
@@ -39,11 +39,16 @@ export interface trackPageloadProps {
...
 
 export interface TrackConnectWalletProps {
-  account: WalletAccount;
+  account?: WalletAccount;
   data?: { [key: string]: string | number | boolean };
   disableTrackingTool?: EventTrackingTool[];
   disconnect?: boolean;
 }
+
+export interface TrackDisconnectConnectWalletProps {
```
chybisov, line 48: Should it be `TrackDisconnectWalletProps`? 👀

packages/dapp/src/components/Widgets/WidgetEvents.tsx
```diff
@@ -127,10 +164,13 @@ export function WidgetEvents() {
       trackEvent({
         action: TrackingActions.OnRouteHighValueLoss,
         category: TrackingCategories.WidgetEvent,
-        label: 'click-highValueLossAccepted',
+        label: 'click_high_value_loss_accepted',
```
chybisov, line 167: Do we want to make these strings also typed somewhere?

chybisov (changes requested):

packages/dapp/src/types/userTracking.types.ts
dennyscode: Right! Thanks

dennyscode: I renamed them in last commit

packages/dapp/src/components/Widgets/WidgetEvents.tsx
dennyscode: Hmm, not sure if that is necessary. But it´s a good question anyways. Wdyt @Addminus

packages/dapp/src/const/trackingKeys.ts
dennyscode: Good idea! It´s all renamed in latest commits

packages/dapp/src/components/Navbar/Menu/WalletMenu/WalletMenu.tsx
```diff
@@ -94,15 +99,21 @@ export const WalletMenu = ({ handleClose }: NavbarMenuProps) => {
       pageload: true,
       disableTrackingTool: [EventTrackingTool.ARCx, EventTrackingTool.Raleon],
     });
+    account.chainId &&
+      openInNewTab(
+        `${
+          getChainById(account.chainId).metamask.blockExplorerUrls[0]
+        }address/${account.address}`,
```
Abhikumar98, line 106: Let's extract this to a new variable and then use the variable here instead of forming a string in the function argument

packages/dapp/src/components/Navbar/SubMenuComponent.tsx
```diff
@@ -51,6 +58,28 @@ const SubMenuComponent = ({
...
+        category: TrackingCategory.SubMenu,
+        action: TrackingAction.OpenMenu,
+        label: `open_submenu_${el.triggerSubMenu.toLowerCase()}`,
+        data: {
+          [TrackingEventParameter.Menu]: el.triggerSubMenu,
+          [TrackingEventParameter.PrevMenu]: prevMenu,
+        },
+        disableTrackingTool: [EventTrackingTool.Raleon, EventTrackingTool.ARCx],
+      });
+    } else {
```
Abhikumar98, line 74: minor suggestion: `else` isn't required here 😄

packages/dapp/src/components/StatsCard/StatsCard.tsx
```diff
@@ -117,6 +121,16 @@ export const StatsCards = ({
       open: openChainsPopper,
       setOpen: setOpenChainsPopper,
       handleOnClick: () => {
+        trackEvent({
+          category: TrackingCategory.WelcomeScreen,
+          action: TrackingAction.OpenStatsModal,
+          label: 'chains_stats',
+          data: { [TrackingEventParameter.StatsModal]: 'chains_stats' },
```
Abhikumar98, line 128: "`chains_stats`" is it a defined type or raw string?

packages/dapp/src/components/Widgets/WidgetEvents.tsx
```diff
@@ -72,14 +87,35 @@ export function WidgetEvents() {
...
+              [TrackingEventParameter.TxLink]: update.process.txLink || '',
+              [TrackingEventParameter.Type]: update.process.type,
+              [TrackingEventParameter.GasCostUSD]: update.route.gasCostUSD,
+              [TrackingEventParameter.ErrorCode]:
+                update.process.error?.code || '',
+              [TrackingEventParameter.ErrorMessage]:
+                update.process.error?.message || '',
+              [TrackingEventParameter.InsuranceFeeAmountUSD]:
+                update.route.insurance.feeAmountUsd,
+              [TrackingEventParameter.InsuranceState]:
```
Abhikumar98, line 117: Data is quite huge, let's use it via variable?

packages/dapp/src/components/Widgets/WidgetEvents.tsx
Abhikumar98: +1

packages/dapp/src/components/Navbar/Menu/WalletMenu/WalletMenu.tsx
dennyscode: Thanks Abhishek! Can you check if that would be suitable the way I changed it now?
https://github.com/lifinance/jumper.exchange/pull/631/commits/4a712c6d859b149e5c3ec1f6f2eeebbc9156aae3

packages/dapp/src/components/Navbar/SubMenuComponent.tsx
dennyscode: Actually, it is needed the way I have set it up. The idea is that you can pass a "triggerSubMenu" to such a MenuItem. This may be any "menuKey" as we called it, such as Wallet, Chains, Themes.. That way, another sub-menu can be opened. 

Other times when a MenuItem serves to trigger some function, else is actually needed in order to make that menu work. E.g. Opening a tab with some url

packages/dapp/src/components/StatsCard/StatsCard.tsx
dennyscode: We have typed the keys but not the values. The idea was that these may be used as Custom Dimensions / Metrics in Google-Analytics.

So, that way "chains_stats" is a raw string.

packages/dapp/src/components/Widgets/WidgetEvents.tsx
dennyscode: That does not feel right to just type all possible labels here, but I´d do it if you think it´s adding value.
I´d still love to hear @Addminus opinion as he beared with me on this whole topic.

packages/dapp/src/components/Widgets/WidgetEvents.tsx
dennyscode: @Abhikumar98 

Thanks for sharing that idea! Generally, I am with you on re-structuring bigger chunks into small pieces for the sake of clean and readable code. 
In this case, imo it´s all in that spot where it belongs, easy to read and follow up. I do not see the need of seperation.

@Addminus Wdyt?

PR #632, opened 2023-09-01
fix: removed redundant boolean type casts

Abhikumar98:
Removed redundant `!!` boolean casts from variables which are already `boolean` in type


dennyscode (approved):
Works as before but cleaner! Thanks!

chybisov (approved):

PR #633, opened 2023-09-07
chore: bump packages

dennyscode:
image


dennyscode:
**Not been able to test the widget, due to missing routes** but it´s unrelated to those changes

PR #634, opened 2023-09-07
fix: el.onClick in submenu-component (LF-3752)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-4752


packages/dapp/src/components/Navbar/SubMenuComponent.tsx
```diff
@@ -75,8 +75,9 @@ const SubMenuComponent = ({
             <MenuLinkItem
               autoFocus={index > 0 ? true : false}
               onClick={() => {
-                !!el.triggerSubMenu && onOpenNavbarSubMenu(el.triggerSubMenu);
-                el.onClick();
+                !!el.triggerSubMenu
```
chybisov, line 78: I guess we removed `!!` everywhere, so maybe we should stop adding new ones 😄

chybisov (changes requested):

packages/dapp/src/components/Navbar/SubMenuComponent.tsx
chybisov: When using `... ? ... : ...`, the first part is treated as a boolean anyway, it's not the case for `... && ...` though 🙂 Just a matter of preference, don't like `!!` everywhere without actual purpose 😄

dennyscode: This makes a whole lot of sense now. I was using it wrong because I thought it needs to be fed with a boolean!
I have removed the double-exclamation marks now. Thank you a lot for spotting that!

chybisov (approved):

PR #635, opened 2023-09-07
style: modal-header bg img on moz as fallback (LF-4760)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-4760

[screen-capture-statsmodal-header-bg.webm](https://github.com/lifinance/jumper.exchange/assets/43956540/393f56a9-9049-4eeb-8984-674cdcbf35cd)


chybisov:
@dennyscode Looks good 🙂 Curious: Is it some old Firefox issue? https://caniuse.com/css-backdrop-filter says backdropFilter can be used since v103.

dennyscode:
@chybisov Not sure, all the sudden I am checking my browsers and it´s all looking odd and nasty :D 
How is it looking for you on Firefox and dark-mode? https://jumper.exchange

chybisov:
@dennyscode okay, it doesn't work for me as well 😄 Maybe we are just using it wrong, according to the specs it should work 😄

dennyscode:
@chybisov 

Oh well! That backdrop-filter has given me a headache for a while on firefox and I never figured out why. I found this post on stackoverflow which is pretty much describing the issue: 

_glmcv wrote:_
> Following setup is causing the backdrop-filter not to work in Firefox (tested in version 114 on macOS):
> 
>     The element with the backdrop-filter: blur(); has position: sticky; set
>     A parent element has a border-radius and overflow property set

https://stackoverflow.com/questions/74277643/firefox-backdrop-filter-blur-not-working-with-sticky-positioning

So, _deleting sticky does give us a blurred header_ but screws up scrolling.

Concluding, this PR is stilll no fix but rather acts as a fallback on mozilla. 
Good you brought it up, now I understand where that was coming from.

chybisov (approved):
Thank you for the good dive in! I guess we can merge this as a fallback 🙂

PR #639, opened 2023-09-08
fix: duplicated language keys

chybisov:
While updating Crowdin to the new language scheme, I noticed we left duplicated language keys in our translations and this PR fixes this.

@Abhikumar98 you might be interested.


Abhikumar98:
Ahh understood, this would have been a bit tricky, thanks for fixing this @chybisov !!

PR #646, opened 2023-09-12
chore: refactor the frontends chain list usage

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-4516


packages/dapp/src/providers/ChainsProvider.tsx
```diff
@@ -0,0 +1,8 @@
+import { useChains } from '../hooks/useChains';
+
+export const ChainsProvider = () => {
+  // populating chains store
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const chains = useChains();
+  return <></>;
+};
```
chybisov, line 8: This is not the correct way of using Providers. Do we have some special needs that we can't just use react-query and its cache for the useChains hook?

chybisov (changes requested):

packages/dapp/src/providers/ChainsProvider.tsx
dennyscode: Thanks Eugene, I was already preparing a list of questions in regards of all that. But this is a great shortcut thanks: Your solution sounds like cleaner way than adding another Zustand store! I will go for that

dennyscode: yaa.. that felt not right using my way of "Provider" :D That was on my list of questions

PR #651, opened 2023-09-13
chore: use storage-persister for useQuery caching

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-4516


chybisov:
Sorry, but why do we need this? 😅

PR #654, opened 2023-09-19
chore: add structuredClone polyfill (LF-3599)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-3599


packages/shared/src/utils/structuredClone.ts
```diff
@@ -0,0 +1,17 @@
...
+    options?: { lossy?: boolean | undefined } | undefined,
+  ): StructuredSerializeOptions | undefined =>
+    structuredClone(value, options)) as {
+    <T = any>(value: T, options?: StructuredSerializeOptions | undefined): T;
+    <T>(
+      value: T,
+      transfer?: { transfer: readonly TransferListItem[] } | undefined,
+    ): T;
+  };
+}
```
chybisov, line 17: Is basic configuration from [@ungap/structured-clone](@ungap/structured-clone) insufficient?

chybisov: Maybe `as any` would be enough here 😅

dennyscode: Sounds legit! Done

packages/shared/src/utils/structuredClone.ts
```diff
@@ -0,0 +1,10 @@
+// Attach the polyfill as a Global function
+import structuredClone from '@ungap/structured-clone';
+
+if (!('structuredClone' in globalThis)) {
+  globalThis.structuredClone = ((
+    value: StructuredSerializeOptions | undefined,
+    options?: { lossy?: boolean | undefined } | undefined,
+  ): StructuredSerializeOptions | undefined =>
+    structuredClone(value, options)) as any;
```
chybisov, line 9: does this work?

`globalThis.structuredClone = structuredClone as any`

dennyscode: - [ ]  I have tested it locally and it seems to do its job just fine when it comes to cloning a (nested) object as seperate instance
- [ ] I have played with `if (!('structuredClone' in globalThis)) {`
...replacing 'structuredClone' with non-existent window objects. This adds this function to it
- [ ] in regards of structuredClone on my browser, there was no change and it still uses the native structuredClone function. So it seems to do what a fallback should do and only uses this polyfill when needed
- [ ] @ilijabojanovic  What I have not tested is the actualy behavior on browsers which do not support it: Which is the reason for this ticket. We need  to check that on old browsers --> https://caniuse.com/?search=structuredClone

image

chybisov: I mean, we should probably make it as one-liner without complicated types because it's a polyfill anyway 🙂

`globalThis.structuredClone = structuredClone as any`

dennyscode: Makes sense, I simplified it

chybisov: @dennyscode, why not simplify it even further? 😅

chybisov (approved):

PR #655, opened 2023-09-19
feat: add fontaine

chybisov:
Ticket: https://lifi.atlassian.net/browse/LF-3583

@dennyscode I tried to debug the fontaine plugin and found that our fonts in CSS have version parameters appended to the extension, which prevents the plugin from correctly processing them. So, after removing them and some other minor adjustments, it works. 🙂
image


dennyscode:
[screen-capture-fontaine-issue-rerendering.webm](https://github.com/lifinance/jumper.exchange/assets/43956540/ee885306-01b7-46e6-836b-0fabe1b185b1)

@chybisov Can we take another few minutes on this tomorrow? On my machine, it still looks a little funny. Is there a way, we can try to use another fallback language instead? Also, it seems to be loading 3 times, especially noticeable on localhost but also seems to happen on the deployed version, just quicker and hard to notice.

dennyscode (approved):

PR #658, opened 2023-09-21
refactor: cleanup import paths (LF-4923)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-4923


packages/dapp/src/App.tsx
```diff
@@ -1,9 +1,7 @@
+import { Menus, Navbar, Widgets } from 'components';
```
dennyscode, line 1: This works just fine in dev mode, yarn build and all - but when committing it throws me an error on all imports I have modified:

`packages/dapp/src/App.tsx(1,40): error TS2307: Cannot find module 'components' or its corresponding type declarations.`

dennyscode: This commit seems to have fixed it!

https://github.com/lifinance/jumper.exchange/pull/658/commits/140021105350458d6262b93ccdc964e938616737

packages/shared/tsconfig.json
```diff
@@ -1,13 +1,27 @@
...
+      "@fonts/*": ["./src/fonts/*"],
+      "@fonts": ["./src/fonts"],
+      "@organisms/*": ["./src/organisms/*"],
+      "@organisms": ["./src/organisms"],
+      "@stories/*": ["./src/stories/*"],
+      "@stories": ["./src/stories"],
+      "@types/*": ["./src/types/*"],
+      "@types": ["./src/types"],
+      "@utils/*": ["./src/utils/*"],
+      "@utils": ["./src/utils"]
```
chybisov, line 23: Why do we need these prefixes?

chybisov (commented):
Can we please remove @transferto completely, move everything to dapp and also move everything from dapp to root src? We've never made it work and it's just sitting there with all the obsolete stuff we no longer need 😂

dennyscode:
This PR will be closed in favor of: https://github.com/lifinance/jumper.exchange/pull/659

PR #673, opened 2023-10-04
chore: remove german translations (LF-5150)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-5150

[screen-capture-languages.webm](https://github.com/lifinance/jumper.exchange/assets/43956540/b0c87044-5b32-473d-98fb-8fff476c7fe1)


dennyscode:
Widget still has german language! Wait

packages/dapp/src/components/Widget/Widget.tsx
```diff
@@ -96,7 +96,7 @@ export function Widget({ starterVariant }: WidgetProps) {
           : '0px 2px 4px rgba(0, 0, 0, 0.08), 0px 8px 16px rgba(0, 0, 0, 0.16)',
       },
       languages: {
-        default: i18n.language as LanguageKey,
+        default: i18n.resolvedLanguage as LanguageKey,
```
dennyscode, line 99: `resolvedLanguage` did the job! 

Example: The LanguageDetector picks up 'de-DE' on a German browser setting. 
LanguageDetector sets the `language` attribute of that i18n instance object accordingly, but that´s causing a problem..

Since the Widget translation handles German, 'de', it uses it! Although we don´t support the German version on Jumper.

By using `resolvedLanguage` in the widget-config, it will only pick up the languages we support on Jumper - those passed in as resources from our translation.json´s. 

There could be other workarounds to solve it. but I think above is best:
- whitelisting languages in an array on supportedLanguages in i18n init options
- filtering found language and resetting after i18n initializtion in .then

chybisov: Sounds good 👌

chybisov (approved):

PR #674, opened 2023-10-06
fix: blockchain explorer hook

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-5175

### DIFF
--- a/packages/dapp/src/hooks/useBlockchainExplorerURL.ts
+++ b/packages/dapp/src/hooks/useBlockchainExplorerURL.ts
@@ -5,13 +5,9 @@ export const useBlockchainExplorerURL = () => {
   const { account } = useWallet();
   const { getChainById } = useChains();
 
-  if (account.isActive && account.chainId) {
-    try {
-      return `${
-        getChainById(account.chainId).metamask.blockExplorerUrls[0]
-      }address/${account.address}`;
-    } catch {
-      throw Error('No blockchain-explorer found');
-    }
+  if (account?.address && account?.chainId) {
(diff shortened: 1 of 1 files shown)


packages/dapp/src/hooks/useBlockchainExplorerURL.ts
```diff
@@ -5,13 +5,9 @@ export const useBlockchainExplorerURL = () => {
   const { account } = useWallet();
   const { getChainById } = useChains();
 
-  if (account.isActive && account.chainId) {
-    try {
-      return `${
-        getChainById(account.chainId).metamask.blockExplorerUrls[0]
```
Abhikumar98, line 11: Do we know for which chainId it was throwing the error?

also, can we have a 
```ts
console.error(`No blockchain explorer found for ${account.chainId}`)
```

chybisov: If we don't have block explorer, please check [data-types](https://github.com/lifinance/data-types) package and add the missing one.

chybisov (approved):

packages/dapp/src/hooks/useBlockchainExplorerURL.ts
dennyscode: @chybisov We do have a blockchain explorer available for each chain we provide - I have tested all of them. 
Apparently, this hook just fired before any chains were fetched - resulting in an error. But this seems to be a rare edge case but now, it should not throw any errors anymore because that if-clause is catching it.

Abhikumar98 (approved):

PR #678, opened 2023-10-10
fix: hide browser wallets for mobile

Abhikumar98:
https://lifi.atlassian.net/browse/LF-3298

Hide browser wallets when for mobile devices

### DIFF
--- a/packages/dapp/src/const/MenuConfig/useWalletSelectContent.tsx
+++ b/packages/dapp/src/const/MenuConfig/useWalletSelectContent.tsx
@@ -1,5 +1,5 @@
 import { supportedWallets, Wallet } from '@lifi/wallet-management';
-import { Avatar } from '@mui/material';
+import { Avatar, useMediaQuery } from '@mui/material';
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useMultisig } from '../../hooks/useMultisig';
 import { useWallet } from '../../providers/WalletProvider';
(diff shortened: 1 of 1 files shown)


dennyscode (changes requested):
1.) What do you think of adding **isDesktopView** to the dependency array? I mean, usually it should be set on first load and detect if mobile or not.. But just wondering.

```
  useEffect(() => {
    initializeWalletSelect();
    // fix: remove 'initializeWalletSelect' from dep´s to fix infinite loop / freeze
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [account.address]);
```

2.) I guess there is multiple ways of detecting a mobile browser vs desktop. Looking at userAgent, it seems as this is not so reliable and can be modified. But this could also be another way of detecting a mobile gadget:

`const isMobile = () => /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent);`

Do you see advantages or disadvantages in that vs. detecting screen size, as it´s done in your approach?

Abhikumar98:
> 1.) What do you think of adding **isDesktopView** to the dependency array? I mean, usually it should be set on first load and detect if mobile or not.. But just wondering.
> 
> 
> 
> ```
> 
>   useEffect(() => {
> 
>     initializeWalletSelect();
> 
>     // fix: remove 'initializeWalletSelect' from dep´s to fix infinite loop / freeze
> 
>     // eslint-disable-next-line react-hooks/exhaustive-deps
> 
>   }, [account.address]);
> 
> ```
> 
> 
> 
> 2.) I guess there is multiple ways of detecting a mobile browser vs desktop. Looking at userAgent, it seems as this is not so reliable and can be modified. But this could also be another way of detecting a mobile gadget:
> 
> 
> 
> `const isMobile = () => /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent);`
> 
> 
> 
> Do you see advantages or disadvantages in that vs. detecting screen size, as it´s done in your approach?

1. At first I wanted to include it in the dependency array, but it didn't really make sense because the user wouldn't be able to change the screen size in mobile or tablet devices. Hence didn't include.

2. Screen size is quite reliable imo, and yes we can use useragents, but screen size check itself covers almost all cases unless there's a super edge case

The advantage to this is it's simple.

dennyscode (approved):

chybisov (approved):

PR #684, opened 2023-10-11
fix: query handling on blockchain-explorer hook

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-5207

### DIFF
--- a/packages/dapp/src/hooks/useBlockchainExplorerURL.ts
+++ b/packages/dapp/src/hooks/useBlockchainExplorerURL.ts
@@ -3,14 +3,12 @@ import { useWallet } from '../providers';
 
 export const useBlockchainExplorerURL = () => {
   const { account } = useWallet();
-  const { getChainById } = useChains();
+  const { isSuccess, getChainById } = useChains();
 
-  if (account?.address && account?.chainId) {
+  if (account.isActive && account.chainId) {
     const chain = getChainById(account.chainId);
-    if (chain?.metamask) {
+    if (isSuccess) {
       return `${chain.metamask.blockExplorerUrls[0]}address/${account.address}`;
-    } else {
-      console.error(`No blockchain explorer found for ${account.chainId}`);
     }
   }
 };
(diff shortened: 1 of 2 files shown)


dennyscode (commented):
Let´s see if that fixes it from now, on!

packages/dapp/src/hooks/useBlockchainExplorerURL.ts
```diff
@@ -3,14 +3,12 @@ import { useWallet } from '../providers';
...
+  const { isSuccess, getChainById } = useChains();
 
-  if (account?.address && account?.chainId) {
+  if (account.isActive && account.chainId) {
     const chain = getChainById(account.chainId);
-    if (chain?.metamask) {
+    if (isSuccess) {
       return `${chain.metamask.blockExplorerUrls[0]}address/${account.address}`;
-    } else {
-      console.error(`No blockchain explorer found for ${account.chainId}`);
```
Abhikumar98, line 13: let's leave this error log here.
It's tends to be informative imo if something goes wrong.

The flow of execution was being interrupted due to `throw Error(...)` https://github.com/lifinance/jumper.exchange/blob/2c3148e060b233b01ff889cac4cfce75ebec9137/packages/dapp/src/hooks/useChains.ts#L30

dennyscode: @Abhikumar98 Sounds legit, I added it back

packages/dapp/src/hooks/useBlockchainExplorerURL.ts
```diff
@@ -3,11 +3,11 @@ import { useWallet } from '../providers';
 
 export const useBlockchainExplorerURL = () => {
   const { account } = useWallet();
-  const { getChainById } = useChains();
+  const { isSuccess, getChainById } = useChains();
```
Addminus, line 6: can we make this bit more readable by renaming isSuccess in this file?
`const {isSuccess: chainsLoaded, getChainById} = useChains();`

Addminus (changes requested):

Addminus (approved):

chybisov (approved):

PR #685, opened 2023-10-11
feat: integrate cookie 3

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-5147

image

- [x] removed Raleon setup
- [x] added Cookie3 to EventTrackingTools keylist
- [x] added provider + hook
- [x] added Cookie3 trackEvent to our setup

Notes:
- first pageView must be sent manually
- staging-mode (boolean) is set in config, depending if (env.mode !== production)
- attributes (user properties) are not supported
- more stuff --> https://docs.cookie3.co/developers/Apis


packages/dapp/src/providers/WalletProvider.tsx
```diff
@@ -101,14 +105,12 @@ export const WalletProvider: React.FC<PropsWithChildren<{}>> = ({
       await liFiWalletManagement.connect(wallet);
       trackEvent({
         action: TrackingAction.ConnectWallet,
+        category: TrackingCategory.Wallet,
+        label: 'connect-wallet',
```
dennyscode, line 109: @Addminus Upfront, I could bet you´re not a big fan of it, as we recently removed the labels and category for all of those within WalletProvider. Cookie3-events rather use a label and like to have a category as required props.
Since we have a very clean setup in place where only those few are lacking a label and a category, I found it reasonable to add it back here. 

To be fair: It´s not stated in their documentation but going from their Typescript definitions and also from what is being sent (analyzing the network tab), that "name" (a.k.a our label) seems to be meant for something.
image

Wdyt?

image

packages/dapp/src/hooks/userTracking/useUserTracking.ts
```diff
@@ -114,14 +109,17 @@ export function useUserTracking() {
...
       if (
-        !disableTrackingTool?.includes(EventTrackingTool.Raleon) &&
-        account.isActive &&
-        account.address
+        !disableTrackingTool?.includes(EventTrackingTool.Cookie3) &&
+        cookie3?.trackEvent({
+          category,
+          action,
+          name: label,
+          //value: todo add to trackingEventProps
```
dennyscode, line 117: @Addminus Also, to keep track of USD a user has spent or received, we might consider using that "value" prop.
Currently, we don´t have it in place within our trackEvents, but we could possibly add it as an optional parameter to the ones, where we actually need them: e.g. Widget-Events (started / completed / failed / updated).

For now, it works without - but it could add value to the data for later analyzation.
Wdyt?

image

packages/dapp/src/const/cookie3.ts
```diff
@@ -0,0 +1,6 @@
+import { UserOptions } from '@cookie3/analytics';
+
+export const cookie3Config: UserOptions = {
+  siteId: 405,
+  staging: import.meta.env.MODE !== 'production',
```
dennyscode, line 5: Using it in localhost, it´s changing the url accordingly:

image

packages/dapp/src/hooks/userTracking/useUserTracking.ts
```diff
@@ -114,14 +109,17 @@ export function useUserTracking() {
...
+        })
       ) {
-        window.raleon.registerEvent(action, account.address, category);
       }
     },
-    [account.address, account.isActive, arcx],
+    [arcx, cookie3],
```
Addminus, line 122: why is this not dependant on the account data anymore?

packages/dapp/src/providers/WalletProvider.tsx
Addminus: we can add the label as param and only use it in the cookie 3 event function. as with name: as far as i understand its the name of the action the user did. use the Tracking action keys for it

Addminus (changes requested):

packages/dapp/src/hooks/userTracking/useUserTracking.ts
dennyscode: We had used the `account.address` and `account.isActive` only for Raleon and don´t need it for cookie3:

**This was before:**
image

packages/dapp/src/providers/WalletProvider.tsx
dennyscode: To be on the same page:
1.) I added the label as a required field to all events. This way, we can also make use of that number it in the cookie3 event. 

2.) You were saying (?): "name" attribute should use our action-value and "action" should be used for what we got as a label? I might misinterpret here, but from that I would rather stick with what we´ve got. Both are pretty descriptive and it should be very readable anyways.

packages/dapp/src/hooks/userTracking/useUserTracking.ts
dennyscode: I have now also added an optional value-property to the trackEvent hook.

packages/dapp/src/components/FeatureCards/FeatureCard.tsx
```diff
@@ -52,7 +52,7 @@ export const FeatureCard = ({ data, isSuccess, assets }: FeatureCardProps) => {
             data.fields.displayConditions.id,
           url: data.fields.url,
         },
-        disableTrackingTool: [EventTrackingTool.ARCx, EventTrackingTool.Raleon],
+        disableTrackingTool: [EventTrackingTool.ARCx],
```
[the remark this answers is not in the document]
dennyscode: @Addminus Is there a reason why we do not want to track events with Cookie3 at all? 

Cookie3 offers that functionality to send events and we have it all in place. I´d be curious on how Cookie3 handles them and maybe it´s good for comparison with GA?

packages/dapp/src/hooks/userTracking/useUserTracking.ts
```diff
@@ -113,15 +109,16 @@ export function useUserTracking() {
...
-        account.isActive &&
-        account.address
-      ) {
-        window.raleon.registerEvent(action, account.address, category);
+      if (!disableTrackingTool?.includes(EventTrackingTool.Cookie3)) {
+        cookie3?.trackEvent({
+          category,
+          action,
+          name: label,
+          value: value ?? undefined,
```
Addminus, line 117: not necessary

Addminus (changes requested):

packages/dapp/src/hooks/userTracking/useUserTracking.ts
dennyscode: Thats right! I got rid of it

Addminus (approved):

chybisov (approved):

PR #686, opened 2023-10-12
fix: trigger trackEvent only on closed wallet-select-menu

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-5219


packages/shared/src/organisms/WalletManagementButtons/WalletManagementButtons.tsx
```diff
@@ -66,7 +66,7 @@ export const WalletManagementButtons: React.FC<
   const handleWalletSelectClick = (
     event: React.MouseEvent<HTMLButtonElement>,
   ) => {
-    openNavbarWalletSelectMenu &&
+    !openNavbarWalletSelectMenu &&
```
dennyscode, line 69: It was triggering on closing events, but should go the other way around

Addminus (approved):

chybisov (approved):

PR #687, opened 2023-10-12
fix: check installed wallets on wallet selection

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-5218

[wallet-select-logic-screen-capture.webm](https://github.com/lifinance/jumper.exchange/assets/43956540/454cb133-ced4-4846-9742-c9a737f4d359)

- now saving the installed wallets in zustand
- when selecting a wallet, looking up if that one is installed
- otherwise console.error e.g:  _Wallet 'Taho' is not installed_

- [x] improve ux when wallet is selected which is not installed (go with Snackbar?)


dennyscode:
@Addminus 

1.) Now, a wallet won´t connect if it has not been installed. We could potentially add a Snackbar, as a proper warning. In this first iteration, I wanna check if that approach works. 
2.) I am afraid that this works properly across all supported wallets. It´s rough to test.. Wdyt?

Addminus:
can we add the same snackbar as when we copy the wallet address but in red with "<wallet name> not installed"
image

image

dennyscode:
@Addminus Nice! What do you think of the current version?

[wallet-not-installed-snackbar-screen-capture.webm](https://github.com/lifinance/jumper.exchange/assets/43956540/84381805-f286-479f-b64e-5a7f6f9c255a)

Addminus (approved):
looks nice

packages/dapp/src/i18n/en/translation.json
```diff
@@ -32,7 +32,8 @@
       "switchChain": "Switch Chain",
       "copy": "Copy",
       "explore": "Explore",
-      "copiedMsg": "Wallet address copied"
+      "copiedMsg": "Wallet address copied",
```
chybisov, line 35: Please don't shorten keys -> copiedMessage 🙂

packages/dapp/src/components/Navbar/Menu/Menus.tsx
```diff
@@ -1,25 +1,57 @@
...
+    event?: React.SyntheticEvent | Event,
+    reason?: string,
+  ) => {
+    if (reason === 'clickaway') {
+      return;
+    }
+    onOpenSnackbar(false);
+  };
+
+  console.log(openSnackbar);
```
chybisov, line 31: Please no logs 🙂

chybisov (changes requested):

packages/dapp/src/components/Navbar/Menu/Menus.tsx
```diff
@@ -1,25 +1,57 @@
...
       <WalletMenu handleClose={handleClose} />
       <WalletSelectMenu handleClose={handleClose} />
       <ChainsMenu handleClose={handleClose} />
       <SupportModal />
+      <Snackbar
+        open={openSnackbar.open}
+        autoHideDuration={2000}
+        onClose={handleCloseSnackbar}
+        anchorOrigin={{ vertical: 'top', horizontal: 'right' }}
+        sx={{ top: '80px !important' }}
```
chybisov, line 45: Why do we need `!important` here?

packages/dapp/src/components/Navbar/Menu/Menus.tsx
dennyscode: This is cleaned up. Good find!

packages/dapp/src/components/Navbar/Menu/Menus.tsx
dennyscode: We don´t it just had been like that ever since I started to use MUI Snackbar. I have now adjusted it in the theme itself with styleOverrides of the Snackbar component.

packages/dapp/src/i18n/en/translation.json
dennyscode: I will keep that in mind from now on when creating new translation-keys. Also here, it had been like this ever since I created that task of "Copying a wallet address". So, we should probably stick with it now to avoid crowdin conflicts, right?

PR #709, opened 2023-10-25
chore: bump arc x version to 2 0

0xsebn:
Jira: https://lifi.atlassian.net/jira/software/c/projects/LF/boards/15?selectedIssue=LF-5275

Attributes removed since sdk already tracks it on its own


dennyscode (changes requested):
A couple more ideas I picked up from the docs:

1.) Not fully sure what is being picked up by default and what not. But in the docs it stated, that disconnection will only be picked up by ArcX on Metamask wallet. So, I am tended to:

Adding the "disconnect" function provided by ArcX. We currently do not include it in our "trackDisconnectWallet". Since it relies on the account, I believe we have to make another adjustment.

image
image

2.) Same with the pageload. Although, it´s doing things straight outta the box, not sure if that applies with the pageload as well. But maybe we can check later on and actually log into the dashboard and test.

Right now, we´re just sending an "event" 
**BUT** actually, all of these "trackPageload" have arc-x disabled in the "disabledTrackingTool" array. So, not really important as it´s not used anyways.

My thought at some point was to send events for internal pages (like on jumper.exchange/something) and use that arcx´s pageload function for external links.

image

Let´s have a chat later

dennyscode (approved):

chybisov (approved):

Addminus (changes requested):
Actually taking into consideration the upcoming jumper blog, that page tracking would be very interesting! I think we should add it now for internal pages (like denny suggested). implementing a track pageview function for that on useUserTracking would be suitable. Also i noticed the docs had a switch chain function. a new function for tracking that would be nice as well?

dennyscode:
@Addminus 

1.) Yes, for that blog it´d make sense to have it in place. Currently, we are just not using it at all - but taking that into consideration it could make sense to have it prepared.
2.) On the switch-chain: I am not sure what is being automatically picked up by the SDK-Provider/Wrapper that comes from arcx and what not. Some functions in the past, we had integrated, were already triggered without further code on our end. @0xsebn Since you were chatting with Kerman last, could you check on that part?

packages/dapp/src/components/WelcomeWrapper/WelcomeWrapper.tsx
dennyscode, line 1: Those changes applied here, can now go into WelcomeScreen.tsx. Just that minor thing of removing arcx from that disabledTrackingTools array, I would bet.

dennyscode (changes requested):
Just that little thing. One more change that just changed recently due to the update of the welcome screen. Thanks for bearing with it! :)

packages/dapp/src/components/WelcomeWrapper/WelcomeWrapper.tsx
0xsebn: got it, thanks denny :D

dennyscode (approved):

chybisov (changes requested):
Not sure why husky didn't manage this, but could you please fix formatting issue like missing spaces 🙂

chybisov (approved):

Addminus (approved):

PR #711, opened 2023-10-25
style: update feature cards

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-5427

[feature-cards-update-screen-capture.webm](https://github.com/lifinance/jumper.exchange/assets/43956540/c42100d5-f5ff-472f-a365-07aed4606299)

image
image

- [ ] Still gotta update documentation on Notion, a few things have changed
- [ ] Activate all cards by Publishing them on Contentful, currently all in draft mode (!)

### DIFF
--- a/packages/dapp/src/components/FeatureCards/FeatureCard.style.tsx
+++ b/packages/dapp/src/components/FeatureCards/FeatureCard.style.tsx
@@ -8,23 +8,20 @@ import {
 import { styled } from '@mui/material/styles';
 
 export interface CardProps extends Omit<MuiCardProps, 'component'> {
-  gradient?: string;
+  backgroundImageUrl?: string;
 }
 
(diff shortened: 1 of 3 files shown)


0xsebn:
looking good, not sure why the check is failing though

dennyscode:
I found one bug that was fixed in my last commit:

Since we can add a "custom" label which can also be used to set the "mode" of the card to light or dark, it is now able to pick them up. This is needed to set the appropriate typography color in order to have a proper contrast.
Before that fix, it only worked on custom-mode "dark". Now, it also works with such a "light" mode

Addminus (commented):
How are the old cards handled? For me they still exist on contentful, especially since the gradientcolor is not used aymore in the code.

Addminus (approved):

PR #734, opened 2023-11-30
Fix typos

boulderuphillroller:
Smol typos


chybisov:
Thank you! Actually, that section was a bit outdated. Updated now 🙂

PR #743, opened 2023-12-22
fix: wrap tool-card content

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-6357

image


chybisov (approved):

src/components/WelcomeScreen/ToolCard/ToolCard.style.ts
```diff
@@ -3,7 +3,6 @@ import { styled } from '@mui/material';
 import { alpha } from '@mui/material/styles';
 
 export const SCard = styled('div')(({ theme }) => ({
```
chybisov, line 5: Why is it called SCard? Maybe we should try to find a more self-descriptive name 😄

dennyscode: Not sure if TCard does any better, but at least it´s closer to the idea of using "ToolCard". I have added that change. Open for a more descriptive name if you can think of one, though

chybisov: It can be `ToolCardContainer`. 🙂

chybisov (approved):

Addminus (approved):

PR #752, opened 2024-01-10
feat: enable translations for strapi queries

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-6468


src/components/FeatureCards/FeatureCards.tsx
```diff
@@ -5,20 +5,53 @@ import { FeatureCard } from 'src/components';
...
+      (foundItem) =>
+        foundItem?.attributes?.DisplayConditions.id ===
+        el.attributes?.DisplayConditions.id,
+    );
+    if (translatedItem) {
+      return deepMerge(el, translatedItem);
+    }
+    return el;
+  });
+
```
chybisov, line 51: Should we put these in one `featureCardsFetched` useMemo?

dennyscode: Thanks Eugene! Finally returned to that one you raised: I refactored it and put it into a single useMemo which seems better: But generally, I´d like to discuss this a little more with you

chybisov: Sure! Just let me know when it's best for you and my DM is always open 😄

src/components/FeatureCards/FeatureCards.tsx
```diff
@@ -5,36 +5,67 @@ import { FeatureCard } from 'src/components';
...
   useEffect(() => {
-    if (Array.isArray(featureCardsFetched)) {
-      !!featureCardsFetched.length &&
-        setFeatureCards(featureCardsFetched?.slice(0, 4));
+    if (Array.isArray(cleanedTranslation)) {
+      !!cleanedTranslation.length &&
+        setFeatureCards(cleanedTranslation?.slice(0, 4));
     }
-  }, [featureCardsFetched]);
+  }, [cleanedTranslation]);
```
chybisov, line 68: I think we can move this logic to useMemo as well. There is no need for additional effect and re-render just to slice an array 🙂

chybisov (changes requested):

PR #767, opened 2024-02-01
feat: improve feature cards

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-6813

few topics to be covered:
- entire feature card clickable ✅
- differentiate between cta clicked + bg clicked Tracking ✅
- set default bg color according to theme mode (as a backup for blank images)  ✅
- apply following logic: show drafts only on develop, show ONLY publsihed ones on staging + production ✅
- use "ID" from field, not from displayConditions ✅
- pick bg img according to "mode" - if set --> right now, a img needs to be provided for light + dark, even though a card "mode" might be set to dark -> Now, only the dark image will be used in those cases ✅


dennyscode:
Backing up Bg Colors according to "mode" setup:

image

dennyscode:
Found "12-14-production" as an ID -> This should be more descriptive!

dennyscode:
@Addminus @chybisov @tcheee 

Let´s discuss:

1.) Now, the whole feature-card is clickable -> I have added another parameter to differentiate if bg or cta clicked. Is it too much? Do we need it?
2.) Bg colors: Could be improved -> Check my comment above under the attached image of the current state
3.) BG Image -> I have added a logic, to only apply the correct bg image when a "mode" has been set. If "mode" === dark, it would ONLY use the dark image. Before, you needed to attach the same "dark" image to light + dark always, even if the card was marked as "dark". Does it improve, or is it too much? (I can see an issue, if we ever come up with two versions of a specific "dark" or "light" mode -> e.g. have two different "dark" bg´s depending on theme)
4.) Important Note: We moved the ID from displayConditions to a new field, to make queries filterable by ID (which is not possible for JSONs)

src/components/FeatureCards/FeatureCard.tsx
```diff
@@ -76,56 +80,98 @@ export const FeatureCard = ({ data, isSuccess }: FeatureCardProps) => {
...
+        url.origin,
+      );
+    }
+  }, [
+    data.attributes.BackgroundImageDark.data?.attributes.url,
+    data.attributes.BackgroundImageLight.data?.attributes.url,
+    data.attributes.DisplayConditions.mode,
+    theme.palette.mode,
+    url.origin,
+  ]);
```
dennyscode, line 122: @chybisov Wdyt? 😄

dennyscode:
Updated fallback bg:
image

src/components/FeatureCards/FeatureCard.tsx
chybisov: Maybe this 😄

```ts
const mode = data.attributes.DisplayConditions.mode || theme.palette.mode;
const imageUrl = mode === 'dark'
  ? data.attributes.BackgroundImageDark.data?.attributes.url
  : data.attributes.BackgroundImageLight.data?.attributes.url
return new URL(
  imageUrl,
  url.origin,
);
```

src/components/FeatureCards/FeatureCard.tsx
```diff
@@ -33,8 +34,12 @@ export const FeatureCard = ({ data, isSuccess }: FeatureCardProps) => {
   useEffect(() => {
     data?.attributes.DisplayConditions &&
       data?.attributes.DisplayConditions.showOnce &&
-      onDisableFeatureCard(data?.attributes.DisplayConditions?.id);
-  }, [data?.attributes.DisplayConditions, onDisableFeatureCard]);
+      onDisableFeatureCard(data?.attributes.uid);
+  }, [
```
chybisov, line 38: Please let's KISS it 😄

```ts
if (data?.attributes.DisplayConditions?.showOnce) {
     onDisableFeatureCard(data?.attributes.uid);
}
```

src/components/FeatureCards/FeatureCard.style.tsx
```diff
@@ -5,18 +5,22 @@ import { styled } from '@mui/material/styles';
 
 export interface CardProps extends Omit<MuiCardProps, 'component'> {
   backgroundImageUrl?: string;
+  dark?: boolean;
 }
 
 export const FCard = styled(MuiCard, {
```
chybisov, line 11: I remember we renamed this to something more self-explanatory, didn't we? 👀

chybisov:
> @Addminus @chybisov @tcheee
> 
> Let´s discuss:
> 
> 1.) Now, the whole feature-card is clickable -> I have added another parameter to differentiate Tracking according to if bg or cta clicked. Is it too much? Do we need it? 2.) Bg colors: Could be improved -> Check my comment above under the attached image of the current state 3.) BG Image -> I have added a logic, to only apply the correct bg image when a "mode" has been set. If "mode" === dark, it would ONLY use the dark image. Before, you needed to attach the same "dark" image to light + dark always, even if the card was marked as "dark". Does it improve, or is it too much? (I can see an issue, if we ever come up with two versions of a specific "dark" or "light" mode -> e.g. have two different "dark" bg images which should change when the theme switches between light + dark theme) 4.) Important Note: We moved the ID from displayConditions to a new field, to make queries filterable by ID (which is not possible for JSONs)

I'd suggest having dark and light images, and if we don't have one of them, then use the other one. For your issue with two dark images, I think we can name one of them as light even though it is dark. 😄

src/hooks/useFeatureCards.ts
```diff
@@ -18,6 +18,7 @@ export const useFeatureCards = (): UseFeatureCardsProps => {
   apiUrl.searchParams.set('populate[0]', 'BackgroundImageLight');
   apiUrl.searchParams.set('populate[1]', 'BackgroundImageDark');
   import.meta.env.MODE !== 'production' &&
+    import.meta.env.MODE !== 'staging' &&
```
tcheee, line 21: Super small change but don't we have a 'development' MODE?

src/components/FeatureCards/FeatureCard.tsx
dennyscode: Nice one, thanks

src/hooks/useFeatureCards.ts
dennyscode: YES! That simplifies the condition a lot :D Thanks!

dennyscode:
> > @Addminus @chybisov @tcheee
> > Let´s discuss:
> > 1.) Now, the whole feature-card is clickable -> I have added another parameter to differentiate Tracking according to if bg or cta clicked. Is it too much? Do we need it? 2.) Bg colors: Could be improved -> Check my comment above under the attached image of the current state 3.) BG Image -> I have added a logic, to only apply the correct bg image when a "mode" has been set. If "mode" === dark, it would ONLY use the dark image. Before, you needed to attach the same "dark" image to light + dark always, even if the card was marked as "dark". Does it improve, or is it too much? (I can see an issue, if we ever come up with two versions of a specific "dark" or "light" mode -> e.g. have two different "dark" bg images which should change when the theme switches between light + dark theme) 4.) Important Note: We moved the ID from displayConditions to a new field, to make queries filterable by ID (which is not possible for JSONs)
> 
> I'd suggest having dark and light images, and if we don't have one of them, then use the other one. For your issue with two dark images, I think we can name one of them as light even though it is dark. 😄

We´ve agreed to keep in the logic

src/components/FeatureCards/FeatureCard.tsx
dennyscode: This is genious!

src/components/FeatureCards/FeatureCard.style.tsx
dennyscode: Renamed it to: "isDarkCard"

chybisov (approved):

tcheee (approved):

PR #772, opened 2024-02-10
fix: update wallet address btn padding

ymittal:
image

On prod, I noticed uneven padding which kinda irked me and I am not sure if this was intentional. So putting up a quick fix to make the left and right padding the same!

**After**
image


chybisov:
@ymittal which wallet do you use? Looks like you don't have a wallet icon there.

chybisov:
@ymittal please describe detailed steps to reproduce. I tried to install Rabby and Rainbow alongside my MetaMask, but I see all icons when I connect.

@dennyscode we need to add a fallback if we don't have an icon for a wallet or it is unavailable for some reason.

chybisov:
We will make more improvements to the wallet menu in the next few days, stay tuned and thank you for your attention to details 🙏🙂

PR #777, opened 2024-02-12
feat: personalised feature cards for users

tcheee:
Testing requirements:
- create a new user inside Strapi with your wallet address in Jumper User collection
- associate a specific card to your user in the Feature Card collection
- connect your wallet and you should view the new cards


src/components/FeatureCards/FeatureCards.tsx
```diff
@@ -25,22 +31,38 @@ export const FeatureCards = () => {
...
+        setFeatureCards(featureCardsFetched?.slice(0, 2));
     }
   }, [featureCardsFetched]);
+
+  useEffect(() => {
+    if (Array.isArray(personalizedCardsFetched)) {
+      setPersonalizedFeatureCards(personalizedCardsFetched?.slice(0, 1));
+    }
+  }, [personalizedCardsFetched]);
+
```
chybisov, line 50: There is no need to use `useEffect` here and we can just use data received from hooks.

cc @dennyscode

src/hooks/useFetchUser.ts
```diff
@@ -0,0 +1,63 @@
...
+import type { FeatureCardData } from 'src/types';
+
+export interface UseFetchUserProps {
+  featureCards: FeatureCardData[] | undefined;
+  isSuccess: boolean;
+  isConnected: boolean;
+}
+
+const STRAPI_CONTENT_TYPE = 'jumper-users';
+export const useFetchUser = (): UseFetchUserProps => {
```
chybisov, line 12: What do you think about changing the name to `usePersonalizedFeatureCards` or if we have any other information about the user we can rename it to `useUser`.

src/hooks/useFetchUser.ts
```diff
@@ -0,0 +1,63 @@
...
+      const result = await response.json();
+      return result.data;
+    },
+    enabled: !!account.address,
+    refetchInterval: 1000 * 60 * 60,
+  });
+  const featureCards = data?.[0]?.attributes?.feature_cards.data;
+
+  return {
+    featureCards: featureCards ?? undefined,
```
chybisov, line 59: It will be undefined anyway so `?? undefined` seems unnecessary

src/hooks/useFetchUser.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export interface UseFetchUserProps {
+  featureCards: FeatureCardData[] | undefined;
+  isSuccess: boolean;
+  isConnected: boolean;
+}
+
+const STRAPI_CONTENT_TYPE = 'jumper-users';
+export const useFetchUser = (): UseFetchUserProps => {
+  const { account } = useWallet();
```
chybisov, line 13: We merged widget v3 PR so I guess you should use `useAccount` here 🙂

src/hooks/useFetchUser.ts
```diff
@@ -0,0 +1,63 @@
...
+      'filters[EvmWalletAddress][$eqi]',
+      account?.address,
+    );
+  }
+  import.meta.env.MODE !== 'production' &&
+    apiUrl.searchParams.set('publicationState', 'preview');
+  const apiAccesToken =
+    import.meta.env.VITE_STRAPI_DEVELOP === 'true'
+      ? import.meta.env.VITE_LOCAL_STRAPI_API_TOKEN
+      : import.meta.env.VITE_STRAPI_API_TOKEN;
```
chybisov, line 40: We have configs for all environments so maybe there should be only one `VITE_STRAPI_API_TOKEN` variable with different values for different envs?

chybisov (changes requested):

src/hooks/useFetchUser.ts
dennyscode: I disagree on that one. I believe useUser is not very descriptive: We have an actual collection type for "Users". Whenever we will have to fetch users - I´d rather call that "useUsers". Here, it was just about fetching "personalized" feature cards by just adding one more filter.

chybisov: Okay, the first suggestion still applies 😄

src/components/FeatureCards/FeatureCards.tsx
```diff
@@ -1,47 +1,67 @@
...
+      return personalizedCards
+        ?.filter(
+          (el, index) =>
+            el.attributes.DisplayConditions &&
+            !disabledFeatureCards.includes(el.attributes.DisplayConditions?.id),
+        )
+        .slice(0, 1);
     }
-  }, [featureCardsFetched]);
+    //trigger featureCardsFetched-filtering only once
```
chybisov, line 45: Why do we want to trigger this only once?

dennyscode: Thomas has applied it from the existing useMemo. Up there, at some point I had added the "eslint-disable-next-line" in order to only run this snippet once and not repeatedly trigger it. To be honest, now I have tried it without and it seems to work as well. Wdyt?

Anyways, it seems as we don´t need it on the _slicedPersonalizedFeatureCards_ because it does not affect the deps. We can get rid of this (line 39+40)
```
    //trigger featureCardsFetched-filtering only once
    // eslint-disable-next-line react-hooks/exhaustive-deps
```

chybisov: Yes, I think we should avoid disabling the hooks rules unless it is really needed and even in that way it shows that we are doing something wrong 😄

chybisov (approved):

dennyscode (approved):

PR #779, opened 2024-02-13
feat: show last wallet first in wallet management

Addminus:
Ticket: https://lifi.atlassian.net/browse/LF-6934

In order to provider a better UX and to give the user feedback on the success of his wallet connection, the wallet digest button should always show the last connected wallet.


chybisov:
I see why we are making this change and don't think we should make it this way. Let's discuss this on Thursday, I have an idea of how we can improve here.

chybisov (approved):

PR #823, opened 2024-03-06
fix: center chain badge on wallet avatar

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-7251

**Info: I could not get this centered 100% correctly!**

Testing:
- Background-Color (as seen in the image underneath) is not happening on code! But rather just for testing
- to test properly, I applied the following styling temporarily to the Span-Contaner: `span: {background-color: rgba(0,0,0,0.5}; border-radius: 44px}`
- attached image is based on a 200% zoom level + browser zoom
- so going into the nitty gritty details!

image

### DIFF
--- a/src/components/Menus/WalletMenu/WalletMenu.style.ts
+++ b/src/components/Menus/WalletMenu/WalletMenu.style.ts
@@ -88,7 +88,7 @@ export const WalletCardBadge = styled(Badge)(({ theme }) => ({
   },
   '> .MuiAvatar-root': {
     overflow: 'hidden',
-    '--g': '#0000 100%,#000',
+    '--g': '#0000 98%,#000',
     '--s': '100% 100% no-repeat',
     '--mask':
       'radial-gradient(circle 23px at calc(100% - 8px) calc(100% - 8px),var(--g)) 100% 100%/var(--s)',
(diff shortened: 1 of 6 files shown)


dennyscode:
While investigating on where this was coming from I came to the conclusion that some changes in this PR resulted in this issue:
https://github.com/lifinance/jumper.exchange/pull/801

src/components/Menus/WalletMenu/WalletMenu.style.ts
```diff
@@ -78,9 +78,17 @@ export const WalletCardButtonContainer = styled(Container)(({ theme }) => ({
...
+    height: 44,
+    padding: 0,
+    transform: 'unset',
+    bottom: '-11.5px',
+    right: '-13.5px',
+  },
   '> .MuiAvatar-root': {
     overflow: 'hidden',
-    '--g': '#0000 98%,#000',
+    '--g': '#0000 100%,#000',
```
chybisov, line 91: I think it looks better with a bit of blur when not zoomed in. Without blur, there is no antialiasing effect 
image

chybisov: Also, I think we should have different masks for different image sizes. If you take a look at the small icon, there is a bigger cutout
image

chybisov (changes requested):

tcheee (approved):

src/components/Menus/WalletMenu/WalletMenu.style.ts
dennyscode: Good points! Here is some updates:

WalletCard
image

WalletButton
image

Both:
image

dennyscode: Updated again:
image

dennyscode:
image
image

PR #842, opened 2024-03-18
feat: change integrator string for gas tab

tcheee:


src/components/Widgets/Widget.tsx
```diff
@@ -37,6 +37,8 @@ export function Widget({ starterVariant }: WidgetProps) {
   const wagmiConfig = useConfig();
   const { isMultisigSigner, getMultisigWidgetConfig } = useMultisig();
   const { multisigWidget, multisigSdkConfig } = getMultisigWidgetConfig();
+  const isGasVariant =
+    window?.location?.href?.includes(TabsMap.Refuel.variant) ?? false;
```
dennyscode, line 41: Another approach could be: "activeTab" from store. Could give that a try but not sure which one works better.

.env.development
```diff
@@ -3,6 +3,7 @@ VITE_HOTJAR_ID=3307684
 VITE_HOTJAR_SNIPPET_VERSION=6
 VITE_LIFI_API_URL=https://develop.li.quest/v1
 VITE_WIDGET_INTEGRATOR=dev.jumper.exchange
+VITE_WIDGET_INTEGRATOR_REFUEL=dev.jumper.exchange.gas
```
dennyscode, line 6: Also update the env declaration file accordingly.

dennyscode (changes requested):

.env.development
tcheee: Done thanks

src/components/Widgets/Widget.tsx
tcheee: Way cleaner, thank you!

dennyscode (approved):

Addminus (approved):

PR #844, opened 2024-03-20
feat: fix bug on multiwallet connection + loading state

tcheee:


src/hooks/useLoyaltyPass.tsx
```diff
@@ -96,28 +96,28 @@ export const useLoyaltyPass = (): UseLoyaltyPassProps => {
...
-      address: account?.address,
+      isLoading: isLoading,
       points: undefined,
       tier: undefined,
       pdas: [],
     };
   }
 
-  return { ...data, isSuccess: isSuccess };
+  return { ...data, isSuccess: isSuccess, isLoading: isLoading };
```
dennyscode, line 122: Let´s try that: 
`return { ...data, isSuccess, isLoading };`

dennyscode (changes requested):

dennyscode (approved):

src/components/ProfilePage/LevelBox/LevelBox.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export const LevelBox = ({ level, loading }: LevelBoxProps) => {
+  return (
+    <TierBadgeBox>
+      {loading ? (
+        <Skeleton variant="text" sx={{ fontSize: '42px', width: '132px' }} />
+      ) : (
+        <ProfilePageTypography fontSize={'18px'} lineHeight={'24px'}>
+          {`LEVEL ${level}`}
```
Addminus, line 17: Do we want to localize that string?

tcheee: No, it is the LEVEL inside your loyalty, the marketing team will speak about "LEVEL" inside the "Loyalty Pass". As for as I know we don't localize "loyalty pass" so should not "level" as well

Addminus (approved):

PR #853, opened 2024-03-24
refactor: rename store action handlers

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-6891


src/config/config.ts
```diff
@@ -41,8 +39,6 @@ export const defaultSettings: DefaultSettingsType = {
     'auto',
   languageMode: setLanguage() as LanguageKey,
   clientWallets: [],
-  activeWalletName:
-    (localStorage.getItem(localStorageKey.activeWalletName) as string) || '',
```
dennyscode, line 45: We did not take advantage of that value and never used it! So, I got rid of it

src/stores/settings/SettingsStore.tsx
```diff
@@ -35,28 +34,15 @@ export const useSettingsStore = createWithEqualityFn(
...
-          activeWalletName,
-        });
-      },
-
-      onWalletDisconnect: () => {
-        set({
-          activeWalletName: 'none',
-        });
-      },
-
```
dennyscode, line 50: Same here, although we did set the wallet name once the wallet connection started + cleared the name once the wallet disconnects - it´s never been used.

So, those to handlers got removed as well.

DNR500 (approved):

0xsebn (approved):
Good change, will have to update this on this on the scan as well.

tcheee (approved):
Ok, thank you for handling this!

PR #855, opened 2024-03-25
fix: menus on tablet sizes

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-7571

[fix-menu-screen-capture.webm](https://github.com/lifinance/jumper.exchange/assets/43956540/c8a00a7c-7dc2-4cc1-b639-16e3ac622bc2)

### DIFF
--- a/src/components/Menu/Menu.style.ts
+++ b/src/components/Menu/Menu.style.ts
@@ -112,15 +112,15 @@ export const MenuPaper = styled(Paper, {
   },
 
   [theme.breakpoints.up('sm' as Breakpoint)]: {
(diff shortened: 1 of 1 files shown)


DNR500 (approved):

0xsebn (changes requested):
Is it possible to use a border radius value from theme? Not sure if we have a specific borderRadius value on theme.shape ? Besides that looks good to me.

tcheee (approved):
Good for me.

0xsebn (approved):

PR #856, opened 2024-03-25
chore: capitalize wallet name in snackbar

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-6972


src/utils/capitalizeFirstLetter.ts
```diff
@@ -0,0 +1,3 @@
+export const capitalizeFirstLetter = (text: string | undefined) => {
+  return text !== undefined && text.charAt(0).toUpperCase() + text.slice(1);
```
DNR500, line 2: This looks fine as you want to do the capitalisation in the hook. Just be aware you can do this with CSS easily too and its often good to think about when its best to use either approach

```
&:first-letter{
  text-transform: capitalize
}
```

DNR500 (approved):

src/utils/capitalizeFirstLetter.ts
0xsebn: Thats a good point. I agree that this approach is a bit better just for terms of simplicity.

0xsebn (approved):
Looks like build is failing but besides that good to go. Either approach (the one mentioned by Nathan) or the one implemented looks good to me.

src/utils/capitalizeFirstLetter.ts
dennyscode: Great point. I followed you advice!

PR #859, opened 2024-03-27
feat: hide safe wallet unless you are in the safe environment

tcheee:


src/hooks/useCombinedWallets.ts
```diff
@@ -60,7 +60,8 @@ export const useCombinedWallets = () => {
       (connector) =>
         isWalletInstalled(connector.id) &&
         // We should not show already connected connectors
-        account.connector?.id !== connector.id,
+        account.connector?.id !== connector.id &&
+        connector.id !== 'safe',
```
Addminus, line 64: is there a way to grap the actual safe connector id from the wallet management package instead of hardcoding the string here?

tcheee: I am not sure, I have checked and we are looking for the id/name whenever we are trying to identify multisig in general (like the example below)

image

Addminus (approved):

PR #880, opened 2024-04-09
feat: deploy nextJS version

kalote:
Deploy NextJS version of jumper.exchange on testnet


kalote:
testnet has been sunset

PR #897, opened 2024-04-14
chore: next-js fixes

dennyscode:
- introduce client-only wrapper for widget
- style fallback widget skeleton
- restructuring -> remove dynamic imports
- extract tracking related providers and components to its own tracking provider file
- persist theme mode via cookie to remove flickering
- preserver welcome screen state on page load via cookie to remove flickering
--> cookie usage caused DYNAMIC_SERVER_USAGE error -> fixed with force-dynamic
- introduce (custom) useWelcomeScreen-hook to re-use across files
- logos flickered on page load, fixed by adding media queries and adding classNames


src/app/[lng]/(exchange)/exchange/page.tsx
```diff
@@ -1,9 +1,25 @@
...
 
 const Page = () => {
-  return <App starterVariant="default" />;
+  const activeTheme = cookies().get('theme')?.value as
+    | ThemeModesSupported
+    | undefined;
+  const welcomeScreenClosed = cookies().get('welcomeScreenClosed')?.value as
+    | 'false'
+    | 'true'
+    | undefined;
```
Comment (tcheee) [inline src/app/[lng]/(exchange)/exchange/page.tsx:14]: We might want to design a dedicated hook for this in the future.

src/app/[lng]/buy/page.tsx
```diff
@@ -1,9 +1,25 @@
...
 
 const Page = () => {
-  return <App starterVariant="buy" />;
+  const activeTheme = cookies().get('theme')?.value as
+    | ThemeModesSupported
+    | undefined;
+  const welcomeScreenClosed = cookies().get('welcomeScreenClosed')?.value as
+    | 'false'
+    | 'true'
+    | undefined;
```
Comment (chybisov) [inline src/app/[lng]/buy/page.tsx:14]: It would be nice to create a reusable function/hook out of this

src/app/[lng]/(refuel)/refuel/page.tsx
```diff
@@ -1,9 +1,25 @@
-import dynamic from 'next/dynamic';
+import { cookies } from 'next/headers';
+import App from 'src/app/ui/app/App';
+import type { ThemeModesSupported } from 'src/types/settings';
 
-const App = dynamic(() => import('../../../ui/app/App'), { ssr: false });
+export const dynamic = 'force-dynamic';
```
Comment (chybisov) [inline src/app/[lng]/(refuel)/refuel/page.tsx:5]: What is this for? 👀

src/app/[lng]/page.tsx
```diff
@@ -1,19 +1,29 @@
...
+import { cookies } from 'next/headers';
+// import type { Viewport } from 'next';
+import type { ThemeModesSupported } from 'src/types/settings';
+import App from '../ui/app/App';
 
-import type { Viewport } from 'next';
+// export const viewport: Viewport = {
+//   width: 'device-width',
+//   initialScale: 1,
+// };
```
Comment (chybisov) [inline src/app/[lng]/page.tsx:9]: It seems unnecessary and subject to removal 🙂

src/app/page.tsx
```diff
@@ -1,20 +1,3 @@
...
-const App = dynamic(() => import('./ui/app/App'), { ssr: false });
-
-export default async function Page({ params: { lng } }: PageProps) {
-  if (locales.indexOf(lng) < 0) {
-    lng = fallbackLng;
-  }
-
-  return <App starterVariant="default" />;
+export default async function Page() {
+  return <></>;
```
chybisov, line 2: Why do we need this empty page?

src/components/Widgets/ClientOnly.tsx
chybisov, line 1: This shouldn't be inside the Widgets folder as it can be useful in other parts of the app as well

src/components/Widgets/WidgetView.style.ts
```diff
@@ -0,0 +1,107 @@
...
+  boxShadow: '0px 8px 32px rgba(0, 0, 0, 0.08)',
+  borderRadius: theme.shape.borderRadius,
+
+  marginTop: !welcomeScreenClosed ? '24px' : theme.spacing(3.5),
+  cursor: !welcomeScreenClosed ? 'pointer' : 'auto',
+  [`@media screen and (min-height: 700px)`]: {
+    marginTop: !welcomeScreenClosed
+      ? 'calc( 50vh - 680px / 2.75 - 40px)'
+      : theme.spacing(3.5), // (mid viewheight - half-two/thirds widget height - navbar height )
+  },
```
chybisov, line 25: I guess we can avoid these magic numbers by using WidgetSkeleton the same way we use the main Widget component so everything related to its positioning is kind of inherited 🤔

src/components/Widgets/useHydrated.ts
chybisov, line 1: Same as ClientOnly

chybisov (approved):

chybisov (changes requested):
Miss-clicked previous review 😄

src/app/[lng]/(refuel)/refuel/page.tsx
Comment (dennyscode) [reply src/app/[lng]/(refuel)/refuel/page.tsx:5]: @chybisov Thanks for asking! I ran into a weird issue after making use of cookies within those "page.tsx" files.
It ran nice with next dev, but after doing the builld, I was not able to run the page without "DYNAMIC_SERVER_USAGE" error.

src/app/page.tsx
dennyscode: I need to have a "entry" page. But this will not be used as we make use of locales using `next-i18n-router` in middleware.

src/components/Widgets/ClientOnly.tsx
dennyscode: Cool, I will move this to it´s own component.

src/components/Widgets/useHydrated.ts
dennyscode: Do you mean, this should go into the "hooks" folder? Or should it be, whereever the ClientOnly component is sitting?

src/app/[lng]/(refuel)/refuel/page.tsx
Comment (dennyscode) [reply src/app/[lng]/(refuel)/refuel/page.tsx:5]: I have removed it in most places by adding those pages as an async function. One spot though, still uses it

src/app/[lng]/buy/page.tsx
Comment (dennyscode) [reply src/app/[lng]/buy/page.tsx:14]: Good idea! Done!

src/app/[lng]/(exchange)/exchange/page.tsx
Comment (dennyscode) [reply src/app/[lng]/(exchange)/exchange/page.tsx:14]: Nice idea! Done

src/app/page.tsx
dennyscode: Tried removing it again, but it would leave me with an error when accessing "/". For some reason, this file is not being picked up though and takes the file from the [lng] folder. But apparently, I can not delete this file?
I left it in

src/components/Widgets/useHydrated.ts
dennyscode: Update: I moved useHydrated to the hooks folder and ClientOnly to the providers folder.

src/components/Widgets/WidgetView.style.ts
dennyscode: Re-used the WidgetContainer component and only have overwritten styles which cannot be re-used

src/components/Widgets/useHydrated.ts
chybisov: ClientOnly is a simple component, not a provider. I think it should be in the `components` folder.

src/app/page.tsx
chybisov: Let's return null from here if possible, at least, but it would be nice to investigate the issue further and see what can be put inside this page to avoid having an empty page.

kalote (approved):

tcheee (approved):
Approved only to be able to test it manually as feature branch are down for now

kalote (approved):

kalote (approved):

PR #906, opened 2024-04-17
fix: jumper glitch on first connection welcome screen

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-7905


tcheee (approved):
tested it manually. Worked.

PR #907, opened 2024-04-17
feat: add feature card exclusions

oktapodia:
Story link: https://lifi.atlassian.net/browse/LF-7349


src/components/FeatureCards/FeatureCards.tsx
```diff
@@ -47,7 +47,21 @@ export const FeatureCards = () => {
...
+      // Remove excluded feature cards
+      const excludedFeatureCards = cards?.filter((el, index) => {
+        if ((el.attributes.featureCardsExclusions?.data || []).length === 0) {
+          return true;
+        }
+
+        const exclusions =
+          el.attributes.featureCardsExclusions?.data.map(
+            (item) => item.attributes.uid,
+          ) ?? [];
```
tcheee, line 59: why this "?? []" check if you already look for the `el.attributes.featureCardsExclusions?.data` to be not null in the if condition above?

src/hooks/useStrapi.ts
```diff
@@ -140,8 +140,12 @@ export const useStrapi = <T>({
   // feature cards -->
   if (contentType === STRAPI_FEATURE_CARDS) {
     // populate images on feature card query
-    apiUrl.searchParams.set('populate[0]', 'BackgroundImageLight');
-    apiUrl.searchParams.set('populate[1]', 'BackgroundImageDark');
+    apiUrl.searchParams.set('populate[BackgroundImageLight]', '*');
+    apiUrl.searchParams.set('populate[BackgroundImageDark]', '*');
```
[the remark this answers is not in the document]
oktapodia: I would like to filter the fields returned by the featureCardsExclusions to only have the uid and reduce the payload size, strapi does not seem to support indexes number and string at the same selector level.

The syntax above is doing the same as before (just written differently) to get 1 level of the nested object.

I can keep it as it was before but I'll have to remove the fields selector from the `featureCardsExclusions` field

dennyscode (approved):
This works like a charm!

My testing approach:
a.) Setup
1.) Created some testing cards (#1, #2)
2.) Within card #2 I added the card #1 to the exclusions
b.) Testing
1.) Enter welcome screen to see cards
2.) Close card #1 to have it added to the disabledFeatureCards array in localstorage
3.) Reload the page
4.) No card #2 shown, woohoo!! :)

[feature-card-exclusion-screen-capture.webm](https://github.com/lifinance/jumper.exchange/assets/43956540/aa215cea-cea0-451a-86f8-b402adce5cc5)

src/components/FeatureCards/FeatureCards.tsx
```diff
@@ -49,7 +49,23 @@ export const FeatureCards = () => {
 
   const slicedFeatureCards = useMemo(() => {
     if (Array.isArray(cards) && !!cards.length) {
-      return cards
+      // Remove excluded feature cards
+      const excludedFeatureCards = cards?.filter((el, index) => {
```
tcheee, line 53: @oktapodia please make the same logic for personalized feature cards as well.

tcheee (changes requested):

PR #910, opened 2024-04-17
feat: integrate widget into blog

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-7296
image
image

Pass in props via strapi article:
- fromChain
- fromToken
- toChain
- toToken
- fromAmount

test here: 
/learn/denny-how-to-bridge-tokens-from-ethereum-to-zksync/?fromAmount=1


src/components/Blog/BlogWidget.tsx
```diff
@@ -0,0 +1,46 @@
...
+}: BlogWidgetProps) => {
+  const allowChainsArray = (allowChains || '')
+    .split(',')
+    .map((el) => parseInt(el))
+    .filter((num) => !isNaN(num));
+
+  return (
+    <>
+      <Box
+        sx={{ display: 'flex', justifyContent: 'flex-end', marginBottom: 3 }}
```
oktapodia, line 30: Better to create a styled components into the related .style.ts file to follow the app structure

src/components/Blog/CustomRichBlocks.style.ts
```diff
@@ -0,0 +1,72 @@
...
+    0.75,
+  ),
+  margin: theme.spacing(2, 0),
+  fontSize: '18px',
+  lineHeight: '32px',
+  fontWeight: 400,
+}));
+
+// Headings
+export const BlogH1 = styled(Typography)(({ theme }) => ({
```
oktapodia, line 20: Seems a lot of style override, can't we apply that at the theme level? there's also a lot of duplication between each H* variants

src/components/Blog/CustomRichBlocks.tsx
```diff
@@ -22,19 +29,45 @@ interface ImageData {
   image: MediaAttributes;
 }
 
+interface WidgetRouteSettings {
```
oktapodia, line 32: Duplication of https://github.com/lifinance/jumper.exchange/pull/910/files#diff-c1e6733c6c5412839aca9d7ae8f74d4e0739f922a1376a092498da06b3addc8cR6 woud be good to extract it

src/components/Blog/CustomRichBlocks.tsx
```diff
@@ -22,19 +29,45 @@ interface ImageData {
...
+        case 3:
+          return <BlogH3 variant="h3">{children}</BlogH3>;
+        case 4:
+          return <BlogH4 variant="h4">{children}</BlogH4>;
+        case 5:
+          return <BlogH5 variant="h5">{children}</BlogH5>;
+        case 6:
+          return <BlogH6 variant="h6">{children}</BlogH6>;
+        default:
+          return <BlogH1 variant="h1">{children}</BlogH1>;
```
oktapodia, line 68:
```suggestion
        case 1:
        default:
          return <BlogH1 variant="h1">{children}</BlogH1>;
```

Would be better as it is doing the same thing, you can also remove the case 1 in that case

src/components/Blog/CustomRichBlocks.tsx
```diff
@@ -55,8 +88,43 @@ export const CustomRichBlocks = ({
         } catch (error) {
           return;
         }
-      } else if (children[0].props.text.includes('<WIDGET>')) {
-        return <Widget starterVariant="default" />;
+      } else if (children[0].props.text.includes('<WIDGET')) {
```
oktapodia, line 91: Would be good to put the related if `props.text` part into the big if text, logic of the different type of components could be splitted to different files.

e.g.
``` 
import RichBlockCTA from '@/components/RchBlocks/RchBlockCTA'
import RichBlockWidget from '@/components/RchBlocks/RchBlockWidget'
//...
```

src/components/Widgets/Widget.tsx
```diff
@@ -37,10 +37,25 @@ const refuelAllowChains: ChainId[] = [
 
 interface WidgetProps {
   starterVariant: StarterVariantType;
+  fromChain?: number;
```
oktapodia, line 40: The type is very similar to the one above https://github.com/lifinance/jumper.exchange/pull/910/files#diff-c1e6733c6c5412839aca9d7ae8f74d4e0739f922a1376a092498da06b3addc8cR6 would be good to extract it

src/components/Widgets/Widget.tsx
```diff
@@ -49,18 +64,22 @@ export function Widget({ starterVariant }: WidgetProps) {
   const { multisigWidget, multisigSdkConfig } = getMultisigWidgetConfig();
   const { activeTab } = useActiveTabStore();
   const isGasVariant = activeTab === TabsMap.Refuel.index;
-
+  console.log('WIDGET', {
```
oktapodia, line 67: Not sure it is expected to keep the console.log

src/components/Widgets/Widget.tsx
```diff
@@ -81,9 +100,16 @@ export function Widget({ starterVariant }: WidgetProps) {
           setWalletSelectMenuState(true);
         },
       },
+      fromChain: fromChain,
+      fromToken: fromToken,
+      toChain: toChain,
+      toToken: toToken,
+      fromAmount: fromAmount,
```
oktapodia, line 107:
```suggestion
      fromChain,
      fromToken,
      toChain,
      toToken,
      fromAmount,
```

would do the same

oktapodia (changes requested):
Couple of comments

src/components/Blog/BlogWidget.tsx
dennyscode: Much better, thanks!

src/components/Blog/CustomRichBlocks.tsx
dennyscode: Good idea! It´s refactored now

src/components/Blog/CustomRichBlocks.tsx
dennyscode: Good point, that does the job just as well with less code

src/components/Widgets/Widget.tsx
dennyscode: That´s right! I tried to simplify this one as well by re-using the existing interface

src/components/Widgets/Widget.tsx
dennyscode: So true! Now, since we are re-using that interface - it´s deprecated and cleaner anyways. But good point!

src/components/Blog/CustomRichBlocks.style.ts
dennyscode: Good point! Moved it to the theme-level

src/components/Blog/CustomRichBlocks.tsx
dennyscode: Let´s sync up on this one. I am afraid I don´t get it.

src/components/Widgets/Widget.tsx
dennyscode: Thanks, it was still wip. So, I would have removed it - but well spotted! I have removed it by now.

src/components/Blog/CustomRichBlocks.tsx
tcheee: +1

src/components/Widgets/Widget.tsx
```diff
@@ -200,13 +197,7 @@ export function Widget({ starterVariant, themeVariant }: WidgetProps) {
...
-        fallback={
-          <WidgetSkeleton
-            welcomeScreenClosed={welcomeScreenClosed}
-            // title={starterVariant === 'default' ? 'Exchange' : 'Gas'}
-            // buttonTitle={starterVariant === 'default' ? 'Connect' : 'Get Gas'}
-          />
-        }
+        fallback={<WidgetSkeleton welcomeScreenClosed={welcomeScreenClosed} />}
       >
         <LiFiWidget integrator={config.integrator} config={config} />
```
tcheee, line 202: why not directly integrate the LiFiWidget inside the blog rather than reusing this component?

src/theme/theme.ts
```diff
@@ -399,316 +399,170 @@ const themeCustomized: Theme = createTheme({
       'sans-serif',
     ].join(','),
     lifiHeaderDisplay: {
-      fontFamily: [
```
tcheee, line 402: Is it to avoid loading too many fonts in the app?

dennyscode: No, it´s more: That fontFamily is already being set as default and here I repeatedly defined it again for no reason.

tcheee (approved):

src/components/Widgets/Widget.tsx
```diff
@@ -7,52 +8,34 @@ import { useActiveTabStore } from '@/stores/activeTab/ActiveTabStore';
...
-interface WidgetProps {
-  starterVariant: StarterVariantType;
-  themeVariant?: ThemeVariantType;
-}
 
-export function Widget({ starterVariant, themeVariant }: WidgetProps) {
-  const [loaded, setLoaded] = useState(false);
+export function Widget({
+  starterVariant,
+  fromChain,
```
oktapodia, line 30: Might block the current widget implementation if we do not have default value for those

src/components/Widgets/WidgetSkeleton.style.ts
```diff
@@ -6,6 +6,7 @@ export const WidgetSkeletonContainer = styled(WidgetContainer)(({ theme }) => ({
   display: 'flex',
   flexDirection: 'column',
   alignItems: 'center',
+  maxWidth: 416,
```
oktapodia, line 9: Would love to avoid those hardcoded values in the future and work with grid/flex instead

oktapodia (approved):
few comments but not blocking, LGTM

src/components/Widgets/WidgetSkeleton.style.ts
dennyscode: I agree, but I cannot think of a proper way to avoid it right now. I´d love to hear your inputs, though to improve it.

Here, I built a Skeleton to represent something similar as the widget, which has a default width of 416px. I double-checked on the playground https://playground.li.fi/

src/components/Widgets/Widget.tsx
dennyscode: Valid point, but it does not seem to break anything.

PR #911, opened 2024-04-17
feat: design profile mobile version page

oktapodia:


src/components/illustrations/JumperLogo.tsx
```diff
@@ -66,7 +66,7 @@ export const JumperLogo = () => {
         <style type="text/css">
           {`
             .jumper-logo {
-              width: 248px;
+              width: 175px;
```
dennyscode, line 69: Good call! We may even decrease it further to something like 155-160ish? Whatever, good one!

src/components/Tabs/Tabs.style.ts
```diff
@@ -7,14 +7,10 @@ import { Tab as MuiTab, Tabs, styled } from '@mui/material';
 export const TabsContainer = styled(Tabs, {
   shouldForwardProp: (prop) => prop !== 'styles',
 })<TabsProps>(({ theme }) => ({
-  position: 'absolute',
-  left: '50%',
-  transform: 'translateX(-50%)',
```
dennyscode, line 12: Awesome!!!!! Why have I not thought of that one before

src/components/ProfilePage/QuestCard/QuestCard.tsx
```diff
@@ -169,7 +169,7 @@ export const QuestCard = ({
                   lineHeight={'18px'}
                   fontWeight={600}
                   sx={{
-                    padding: '8px',
+                    padding: 8,
```
dennyscode, line 172: Just to make sure and please double-check: My understanding is - using numbers in sx-props will apply the actual spacing. So, if I am not mistaken, this should turn into 64px. Other than that, it is great style to use numbers instead of px and we should do whenever we can (also the code may speak different still to this point)

src/components/ProfilePage/LevelBox/TierBox.tsx
```diff
@@ -27,8 +27,12 @@ export const TierBox = ({ points, loading }: TierBoxProps) => {
 
   return (
     <TierMainBox>
-      <Box sx={{ padding: '18px' }}>
-        <TierInfoBox>
+      <Box sx={{ padding: { xs: 1, sm: 2 } }}>
+        <TierInfoBox
+          sx={{
+            display: { xs: 'none', sm: 'flex' },
```
dennyscode, line 33: Woa! I did not even know we use do it like that!!! Great deal

src/components/ProfilePage/LevelBox/TierBox.style.ts
```diff
@@ -29,7 +29,7 @@ export const TierBadgeBox = styled(Box)(({ theme }) => ({
 }));
 
 export const TierInfoBox = styled(Box)(() => ({
-  marginBottom: '16px',
+  marginBottom: 16,
```
dennyscode, line 32: ❤️

src/components/ProfilePage/LevelBox/PointsBox.tsx
```diff
@@ -12,7 +12,7 @@ export const PointsBox = ({ points }: PointsBoxProps) => {
   return (
     <CenteredBox>
       <ProfilePageTypography
-        fontSize={'80px'}
+        fontSize={80}
```
dennyscode, line 15: ❤️

src/components/ProfilePage/ProfilePage.tsx
dennyscode, line 1: Great job!!!! 💪 
I am still calling for some more spacing but we´ll sync up with Vilen on that.

image

src/components/ProfilePage/QuestCard/QuestCard.tsx
dennyscode: Here we go: Take-away - no numbers in sx props... unfortunately

image

dennyscode (changes requested):
Only thing is that number in sx prop and some more spacing. Other than that, real nice!

dennyscode:
Btw: Can we also fix that please? This bugs me every time!

image
image

It´s well done for the completed missions:
image

dennyscode (changes requested):
image

dennyscode (changes requested):
image

dennyscode (changes requested):
image

dennyscode (changes requested):
different spacings between titles and top border of component

image

image

tcheee (approved):

tcheee (approved):
LGTM but we need to make sure to fix the tablet version.

src/components/illustrations/JumperLogo.tsx
oktapodia: Seems ok to me, Have to get feedback from the designer if we need to change it I guess

src/app/layout.tsx
```diff
@@ -39,7 +39,7 @@ export default async function RootLayout({
         </Script>
       </head>
 
-      <body>
+      <body suppressHydrationWarning>
```
tcheee, line 42: What does it do Seb?

oktapodia: That's to avoid extension blocking the hydratation because they are adding extra stuff to the body element (like grammarly, etc..)

PR #944, opened 2024-04-30
fix: tabs always centered

oktapodia:


dennyscode (changes requested):
In my opinion, we should use this PR to also fix the mobile screens for screensizes < 416px

tcheee (approved):
LGTM. Tested with connected and disconnected wallet.

dennyscode (approved):
@oktapodia Good job removing that absolute positioning. Works like a charm!

PR #957, opened 2024-05-08
fix: updating install husky command over yarn

callMeTheQA:
Since `yarn husky install` is not working, "install" is deprecated the correct command is `yarn add husky` according to https://classic.yarnpkg.com/en/package/husky


tcheee (approved):

oktapodia (approved):
LGTM

PR #968, opened 2024-05-13
refactor: update to use lifi widget skeleton

DNR500:
Jira: [LF-7886](https://lifi.atlassian.net/browse/LF-7886)

We have developed an adaptive skeleton for the Lifi Widget and this PR should put it to use in the Jumper project.

To render correctly on the server-side the skeleton needs access to the theme cookie - currently this is being passed as 
 the activeTheme prop from the serverside component level. However I have notice a CookieProvider component that isn't currently being used - so I want to ask if I should be trying to use that CookieProvider to pass the value? Is there an intended pattern for that provider? [src/providers/CookieProvider.tsx](https://github.com/lifinance/jumper.exchange/blob/e085d9096cbe57c5cafcef7fe46f12b871456022/src/providers/CookieProvider.tsx)

Should be able to see the skeleton when refreshing on the following pages
http://localhost:3001/
http://localhost:3001/refuel
http://localhost:3001/buy

For the Widget on the Blog I have used this page..
http://localhost:3001/learn/denny-how-to-bridge-tokens-from-ethereum-to-zksync/

[LF-7886]: https://lifi.atlassian.net/browse/LF-7886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DNR500:
We had a chat about the passing the cookie as a prop. Should be fine for now. Should be good to consider it for merging

oktapodia (approved):
LGTM

PR #977, opened 2024-05-16
chore: create health-check api call

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-8029


src/app/[lng]/meta/%5Fhealth/route.ts
```diff
@@ -0,0 +1,7 @@
+// API call: /meta/_health
+
```
Comment (tcheee) [inline src/app/[lng]/meta/%5Fhealth/route.ts:2]: A few comment: 
- please no "any"
- what about putting this inside an "api" folder ?
- we should specify "force-dynamic" for this one, no?

Comment (dennyscode) [reply src/app/[lng]/meta/%5Fhealth/route.ts:2]: Good points, thank you for the review!

1.) Good call, I did not use "req" at all and got rid of it.
2.) Please correct me: If I put this into another folder called "api", I will have to use middleware in order to match the required api-url "/meta/_health". Without the "api" folder, it seemed to be the easiest approach.
3.) I had to double-check and ask cody: 

> Do I need force-dynamic to not make this api route cachable?
> 
> No, you don't need to use force-dynamic for this API route. By default, Next.js API routes are not cached, and they are always server-rendered on every request.
> The force-dynamic option is used for static pages (e.g., pages/index.tsx) to prevent them from being statically pre-rendered at build time. It forces the page to be dynamically rendered on every request, similar to an API route.
> Since this file (@src/app/[lng]/meta/_health/route.ts) is an API route, it will always be dynamically rendered on every request, and there's no need to use force-dynamic.

tcheee (approved):

PR #978, opened 2024-05-16
chore: optimize font usage

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-8034

Still making use of inter-var to keep the same font styles as we used to have.

Now:
image

Before: 
image


dennyscode:
I´d prefer to not use our inter-var file to reduce the loaded fonts even more. That way, we´d only use the Inter font provided by Google fonts. That state can be found in the following commit "chore: improve font usage" which we could revert: https://github.com/lifinance/jumper.exchange/pull/978/commits/d1e5079989f1b3b17410e66f36766483fbfef832

image

It comes with the problem that it won´t match the font we used to have:
[font-optimization-screen-capture.webm](https://github.com/lifinance/jumper.exchange/assets/43956540/e26fe5dd-ea64-42a3-b563-aee81274db13)

oktapodia:
It seems that the default font-size is different + letter-spacing css rule, we should be able to match it

tcheee (approved):
lgtm

chybisov (approved):

PR #980, opened 2024-05-17
feat: enable filtering on dates for general feature cards

tcheee:


src/hooks/useStrapi.ts
```diff
@@ -147,6 +147,15 @@ export const useStrapi = <T>({
       'uid',
     );
     apiUrl.searchParams.set('filters[PersonalizedFeatureCard][$nei]', 'true');
+    //filter url
+    const currentDate = new Date(Date.now()).toISOString().split('T')[0];
```
oktapodia, line 151: I'd love to use a date formatter instead of this litte hack, but not blocking

src/hooks/useStrapi.ts
```diff
@@ -147,6 +147,15 @@ export const useStrapi = <T>({
...
+    apiUrl.searchParams.set(
```
oktapodia, line 152: We need to swap to the strapi api soon-ish to make it cleaner and resilient imo

oktapodia (approved):
LGTM

dennyscode (changes requested):
Once I set startCampaign (older than date of today) on strapi that Feature-Card is not shown.

dennyscode (approved):

PR #1002, opened 2024-05-28
fix: full width on menu item links

dennyscode:
quick fixing clickable area:
image

### DIFF
--- a/src/components/Menu/Menu.style.ts
+++ b/src/components/Menu/Menu.style.ts
@@ -145,6 +145,7 @@ export const MenuItemLink = styled(Link, {
   display: 'flex',
   justifyContent: 'space-between',
   height: 48,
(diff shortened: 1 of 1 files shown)


src/components/Menu/Menu.style.ts
```diff
@@ -145,6 +145,7 @@ export const MenuItemLink = styled(Link, {
   display: 'flex',
```
oktapodia, line 145: I strongly recommend to use the Classic [MenuItem](https://mui.com/material-ui/api/menu-item/) instead of a custom one, will avoid many issues in the future starting with this one.

You can set it up as `Link` type via the `component` property

tcheee (approved):

oktapodia (approved):
LGTM

PR #1006, opened 2024-05-29
feat: make the html title variable by page

tcheee:


oktapodia (approved):
LGTM

src/app/[lng]/learn/[slug]/page.tsx
```diff
@@ -1,8 +1,50 @@
...
+            height: 450,
+            alt: 'banner image',
+          },
+        ],
+        type: 'article',
+      },
+    };
+  } catch (err) {
+    return {
+      title: `Jumper Learn | ${sliceStrToXChar(params.slug.replaceAll('-', ' '), 45)}`,
```
Comment (dennyscode) [inline src/app/[lng]/learn/[slug]/page.tsx:43]: Usually, the slug and title seem to be similar for the most part. The advantage of using the title seems to be that it also comes with uppercase letters on some strings which may look little more polished. Also, mostly we don´t use those dashes in titles, so there might not be a need for the replaceAll. 
Wdyt?

src/app/[lng]/learn/page.tsx
```diff
@@ -1,6 +1,14 @@
 import { getArticles } from '@/app/lib/getArticles';
 import { getFeaturedArticle } from '@/app/lib/getFeaturedArticle';
 import LearnPage from '@/app/ui/learn/LearnPage';
+import { Metadata } from 'next';
+
+export async function generateMetadata(): Promise<Metadata> {
+  return {
+    title: 'Jumper Learn',
+    description: 'Jumper Learn is the blog of Jumper Exchange.',
```
Comment (dennyscode) [inline src/app/[lng]/learn/page.tsx:9]: May be a first shot, but I am not convinced this is a good description. Probably this is not important, but seeing this does not get me any excited.

src/app/[lng]/profile/page.tsx
```diff
@@ -1,6 +1,13 @@
 import ProfilePage from '../../ui/profile/ProfilePage';
+import { Metadata } from 'next';
+
+export async function generateMetadata(): Promise<Metadata> {
+  return {
+    title: 'Jumper Profile',
+    description: 'Jumper Profile is the profile page of Jumper Exchange.',
```
Comment (dennyscode) [inline src/app/[lng]/profile/page.tsx:7]: Same here, maybe we can be a little more creative?

src/utils/splitStringToXChar.ts
```diff
@@ -0,0 +1,3 @@
+export const sliceStrToXChar = (s: string, len: number): string => {
+  return s.length < len ? s : s.slice(0, len - 3) + '...';
+};
```
dennyscode, line 3: export const sliceStrToXChar = (s: string, len: number): string =>
  s.length <= len ? s : `${s.slice(0, len - 3)}...`;

-> template string
-> also allow title if it has 45 characters

PR #1012, opened 2024-05-31
feat: add more public rpc endpoints and randomize order

tcheee:


oktapodia (approved):
LGTM

src/utils/shuffleArray.ts
```diff
@@ -0,0 +1,9 @@
+export const shuffleArray = (array: any[]): any[] => {
```
oktapodia, line 1: just use `string[]` as type please

oktapodia, line 1:
```suggestion
export function shuffleArray<T>(array: T): T {
```

Use inferring instead of any please

dennyscode (approved):

PR #1021, opened 2024-06-07
feat: add session id

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-8666


oktapodia (changes requested):
Seems to be overall a lot of duplication. can you do this useEffect/existence check into the hook itself to avoid those duplications?

tcheee (approved):
lgtm

tcheee:
> Seems to be overall a lot of duplication. can you do this useEffect/existence check into the hook itself to avoid those duplications?

It is mainly for testing purpose for QA team and will be remove when we are ready to move to staging

oktapodia (approved):
LGTM

PR #1033, opened 2024-06-11
feat: multi platform build

kalote:
As we will change our infrastructure instance type (from x86 to arm), we need to build our images using multi-platform buildx


tcheee (approved):

oktapodia (approved):
LGTM

PR #1051, opened 2024-06-18
fix: reference error session-storage

dennyscode:
test build

### DIFF
--- a/src/hooks/useSession.ts
+++ b/src/hooks/useSession.ts
@@ -1,16 +1,32 @@
 import { useEffect, useState } from 'react';
 import { uuidv7 } from 'uuidv7';
 
-export const useSession = (): string => {
-  const sessionIdFromStorage = sessionStorage?.getItem('session_id');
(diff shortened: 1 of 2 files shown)


src/hooks/useSession.ts
```diff
@@ -1,16 +1,32 @@
...
     if (sessionIdFromStorage) {
       setSession(sessionIdFromStorage);
     } else {
-      setSession(uuidv7());
-      sessionStorage?.setItem('session_id', session);
+      const newSessionId = uuidv7();
+      setSession(newSessionId);
+
+      // Ensure sessionStorage is available before setting item
+      if (
```
oktapodia, line 24: Hum, seems to be a duplication of line 8 because you are already checking if it exists or not, better to go out of the hook if not exists, wondering if 'use client' can help or not

oktapodia (changes requested):
just a commnt

oktapodia (approved):
LGTM

PR #1067, opened 2024-06-25
fix: clicking on element in welcomeScreen

oktapodia:


dennyscode (approved):
Works like a charm! LGTM

PR #1076, opened 2024-07-01
feat: always close welcome screen on safe

tcheee:


src/hooks/useWelcomeScreen.ts
```diff
@@ -21,7 +23,10 @@ export const useWelcomeScreen = (
 
   useEffect(() => {
     setState(cookie.welcomeScreenClosed);
-  }, [cookie.welcomeScreenClosed]);
+    if (isMultisigSigner) {
+      setState(true);
+    }
```
admsk2, line 28: maybe `setState(cookie.welcomeScreenClosed);` should be moved to else statement, so when `isMultisigSigner` is true we do not `setState` twice

PR #1087, opened 2024-07-04
chore: use uuid package

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-8961


chybisov:
Fixed as part of the #1133

PR #1089, opened 2024-07-04
fix: formatting issues on blog

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-9090

image

Compare: 
http://localhost:3000/learn/jumper-loyalty-pass/
http://jumper.exchange/learn/jumper-loyalty-pass/


src/components/Blog/CustomRichBlocks.tsx
```diff
@@ -195,18 +195,41 @@ export const CustomRichBlocks = ({
...
-                      italic={el.props.italic}
-                      strikethrough={el.props.strikethrough}
-                      underline={el.props.underline}
-                      bold={el.props.bold}
-                      key={`blog-paragraph-${index}`}
-                    >
-                      {el.props.text}
-                    </BlogParagraph>
-                  );
+                  if (el.props.text.indexOf('\n') === -1) {
```
oktapodia, line 198: Not really fan of this, what is the purpose of it exactly? you can use `nl2br` to do this

oktapodia (changes requested):

src/components/Blog/CustomRichBlocks.tsx
dennyscode: Thanks, Sebastien! This condition should be met if there is no line-break within the strapi-paragraph.
In such cases we just return our default <BlogParagraph>. Compared to, if there is a "\n" found, it is split into an array and returns multiple <BlogParagraph>-components.

Feel free to share some more info on how to apply nl2br.

oktapodia: something like this https://github.com/yosuke-furukawa/react-nl2br

dennyscode: Thanks, good idea! Added it now

admsk2 (approved):

src/components/Blog/CustomRichBlocks.tsx
```diff
@@ -199,18 +200,32 @@ export const CustomRichBlocks = ({
...
-                      bold={el.props.bold}
-                      key={`blog-paragraph-${index}`}
-                    >
-                      {el.props.text}
-                    </BlogParagraph>
+                  const nl2brText: Array<ReactElement | string> = nl2br(
+                    el.props.text,
                   );
+                  return nl2brText.map((line, lineIndex: number) => {
+                    if (line === '') {
```
oktapodia, line 207: why is that needed?

oktapodia (commented):
quick question

src/components/Blog/CustomRichBlocks.tsx
dennyscode: Thanks! I removed that part

oktapodia (approved):
LGTM

PR #1096, opened 2024-07-08
chore: add slash to urls by default

dennyscode:


tcheee:
does not seem to work 
image

dennyscode:
> It seems it does not work. image

I am assuming that you have switched to a different language within the menu without reloading the page.
The "language-switch" click-event triggers a javascript to adapt the url change but does not really reload the page. Also, translations are handled on the client. But because the click-event also sets the NEXT_LOCALE cookie, the changed language will persist on next pageload and add the "/".

It works well for me when navigating through the pages.

oktapodia:
I do not think this is useful overall as this is the default behavior within our nextjs settings

dennyscode:
@oktapodia This PR can be closed in favor of yours https://github.com/jumperexchange/jumper-exchange/pull/1145, right?

PR #1120, opened 2024-07-11
[WIP] feat: vertical navbar

admsk2:
**Ticket:**
https://lifi.atlassian.net/browse/LF-8831?atlOrigin=eyJpIjoiMzNiZDVkMmNhYjI5NDViNjkzYjUxNGQ1MGVlODQ0NzIiLCJwIjoiaiJ9

**Description:**
- vertical navbar added next to widget container
- horizontal navbar removed along with navbar blur

### DIFF
--- a/src/app/ui/app/App.tsx
+++ b/src/app/ui/app/App.tsx
@@ -1,5 +1,6 @@
 'use client';
 
+import { usePathname } from 'next/navigation';
 import { WelcomeScreen } from '@/components/WelcomeScreen/WelcomeScreen';
 import type { StarterVariantType } from '@/types/internal';
 import { WidgetContainer } from '@/components/Widgets';
@@ -7,8 +8,12 @@ import { TrackingAction, TrackingCategory } from '@/const/trackingKeys';
 import { EventTrackingTool } from '@/types/userTracking';
 import { useWelcomeScreen } from '@/hooks/useWelcomeScreen';
 import { useUserTracking } from '@/hooks/userTracking';
-import { Box } from '@mui/material';
+import { Box, Stack } from '@mui/material';
 import { StyledSlide } from './App.style';
+import { NavbarTabs } from 'src/components/Navbar';
+import { JUMPER_LEARN_PATH, JUMPER_LOYALTY_PATH } from 'src/const/urls';
+import { usePartnerTheme } from 'src/hooks/usePartnerTheme';
+import { useSuperfest } from 'src/hooks/useSuperfest';
 
 export interface AppProps {
   starterVariant: StarterVariantType;
@@ -17,6 +22,8 @@ export interface AppProps {
 }
 
 const App = ({ starterVariant, isWelcomeScreenClosed, children }: AppProps) => {
+  const pathname = usePathname();
+
   const { trackEvent } = useUserTracking();
 
   const welcomeScreen = useWelcomeScreen(isWelcomeScreenClosed);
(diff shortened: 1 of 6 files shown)


src/app/ui/app/App.tsx
```diff
@@ -62,12 +74,23 @@ const App = ({ starterVariant, isWelcomeScreenClosed, children }: AppProps) => {
...
+      <Stack
+        direction="row"
+        justifyContent="center"
+        alignItems="start"
+        spacing={4}
+        paddingTop={3.5}
       >
-        {children}
-      </WidgetContainer>
+        {!isLearnPage && !hasTheme ? (
```
tcheee, line 84: we might want to change the location of the navbar component to the `widget` to avoid these booleans. What do you think @oktapodia ?

src/components/Widgets/Widgets.style.tsx
```diff
@@ -17,7 +17,7 @@ export const WidgetContainer = styled(Box, {
     display: 'flex',
     margin: '0 auto 24px',
     overflow: !welcomeScreenClosed ? 'hidden' : 'inherit',
-    width: '100%',
+    width: 'auto',
```
tcheee, line 20: why do you have to put back to 'auto'? It make some themes bug.

admsk2: so i have put the navbar next to the widget (wrapped together within `<Stack>` component) and when widget takes 100% width, then the navbar is being pushed out of the screen. If widget has to have 100% of the width, then the navbar has to be positioned absolute i guess

src/app/ui/app/App.tsx
admsk2: we can technically do it, although widget component will not be a single responsibility component then which is usually good

oktapodia: Let's think about it later please because I am already changing a lot of structure in a PR soon, would love to avoid more conflicts than we will have

oktapodia (approved):
LGTM

dennyscode (approved):

src/components/Navbar/WalletButton.tsx
```diff
@@ -102,38 +121,80 @@ export const WallettButtons = () => {
...
+                variant={'lifiBodyMediumStrong'}
+                width={'auto'}
+                marginRight={1.1}
+                marginLeft={1}
+              >
+                {points}
+              </Typography>
+            )}
+            <XPIcon size={28} />
+          </WalletMenuButton>
```
dennyscode, line 164: Maybe this whole section could become it´s own styled-component?

dennyscode (changes requested):

src/components/Navbar/WalletButton.tsx
admsk2: we can do it, but then i would create it's own styled-component for connected button as well; @tcheee what is your take here

src/components/Menus/VerticalMenu/VerticalTabs.tsx
chybisov, line 1: Can we please remove the ripple effect when clicking?

chybisov (changes requested):
https://github.com/user-attachments/assets/5c9257bc-08d5-4c67-88d7-9d9f7e3557ef

Noticed small thing when we click on a back button in the browser we have a bit broken active button animation.

dennyscode (changes requested):
Can we make the transition of the indicator smooth?

tcheee:
added to `ft/beta-testing/1`

PR #1134, opened 2024-07-15
style: fix partner theme footer-image

dennyscode:
image

Partner-Theme footer image not overlapping the widget anymore.

_! The Widget-Container with width of 100% does not allow mouse-events across all screensizes._


src/components/Background.tsx
admsk2, line 1: So if `configTheme?.hasBackgroundGradient` is not true it essentially returns nothing?

src/components/BackgroundGradient/SuperfestBackground.tsx
admsk2, line 1: Same here, maybe that code would be easier to understand:

```
import { useSuperfest } from 'src/hooks/useSuperfest';
import { PartnerThemeFooterImage } from '../PartnerThemeFooterImage';
import { SuperfestBackgroundContainer } from './BackgroundGradient.style';

const SuperfestBackground = () => {
  const { isSuperfest } = useSuperfest();
  
  // Only render the container and contents if isSuperfest is false
  if (isSuperfest) {
    return null;
  }

  return (
    <SuperfestBackgroundContainer>
      <PartnerThemeFooterImage />
    </SuperfestBackgroundContainer>
  );
};

export default SuperfestBackground;
```

tcheee:
@dennyscode  we have lost the logo on page like: https://jumper-exchange-git-fix-partner-footer-l-3d2127-jumper-exchange.vercel.app/fr/superfest/rewards-from-aave-on-base/. Can you give it a look?

src/components/Background.tsx
dennyscode: @admsk2 Pretty much, in that case it will just be the "container" which is rendered. The gradients (top, bottom-left + bottom-right) are optional depending on "hasBackgroundGradient".

src/components/BackgroundGradient/SuperfestBackground.tsx
dennyscode: @admsk2 Good point! Here, we also return the container as default background and optionally add the partner-theme footer-image. So, returning null would not work afaik.

admsk2 (approved):

PR #1137, opened 2024-07-18
fix: prevent scrolling while widget not expanded

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-9007


src/components/FeatureCards/FeatureCards.tsx
dennyscode, line 1: This file contains the widget-expanded listener which has been moved to it´s own hook.

src/components/Widgets/Widget.style.ts
```diff
@@ -19,9 +21,15 @@ export const WidgetWrapper = styled(Box, {
...
-  // },
+
+  [theme.breakpoints.up('sm' as Breakpoint)]: {
+    // height: 200,
+    ...(!isExpanded && {
+      '& + div': {
+        height: 496,
+      },
+    }),
+  },
```
dennyscode, line 32: While widget is not expanded, overwrite height.

src/components/Widgets/Widgets.style.tsx
```diff
@@ -39,6 +39,7 @@ export const WidgetContainer = styled(Box, {
       },
 
       [theme.breakpoints.up('sm' as Breakpoint)]: {
+        height: 496,
```
dennyscode, line 42: Height of the default widget while not expanded.

src/components/Widgets/Widget.style.ts
```diff
@@ -19,9 +21,14 @@ export const WidgetWrapper = styled(Box, {
...
   }),
   zIndex: 2,
-  // [theme.breakpoints.up('sm' as Breakpoint)]: {
-  //   maxWidth: 416,
-  // },
+
+  [theme.breakpoints.up('sm' as Breakpoint)]: {
+    ...(!isExpanded && {
+      '& > div': {
+        height: 616,
```
dennyscode, line 28: This is the minimal widget height we should enable - without Wallet-Address field we can go as low as 496px:
image

src/components/Widgets/Widget.tsx
```diff
@@ -180,39 +179,34 @@ export function Widget({
...
     multisigSdkConfig,
-    multisigWidget,
-    setWalletSelectMenuState,
-    starterVariant,
-    theme.breakpoints,
-    theme.palette.accent1.main,
-    theme.palette.grey,
-    theme.palette.mode,
-    theme.palette.surface1.main,
-    theme.palette.surface2.main,
```
tcheee, line 201: @dennyscode why so many changes inside this file?

src/components/Widgets/Widget.style.ts
```diff
@@ -19,9 +21,14 @@ export const WidgetWrapper = styled(Box, {
     },
   }),
   zIndex: 2,
-  // [theme.breakpoints.up('sm' as Breakpoint)]: {
-  //   maxWidth: 416,
-  // },
+
+  [theme.breakpoints.up('sm' as Breakpoint)]: {
+    ...(!isExpanded && {
```
admsk2, line 26: i think it is quite difficult to read that part, maybe that could be refactored like:

```
export const WidgetWrapper = styled(Box, {
  shouldForwardProp: (prop) => !['welcomeScreenClosed', 'isExpanded'].includes(prop),
})<WidgetWrapperProps>(({ theme, welcomeScreenClosed, isExpanded }) => {
  const baseStyle = {
    width: '100%',
    position: 'relative',
    margin: theme.spacing(0, 'auto'),
    zIndex: 2,
  };

  const hoverEffect = !welcomeScreenClosed ? {
    '&:hover': {
      marginTop: 0,
    }
  } : {};

  const responsiveHeight = theme.breakpoints.up('sm') as Breakpoint;
  const expandedStyle = !isExpanded ? {
    '& > div': {
      height: 616,
    }
  } : {};

  return {
    ...baseStyle,
    ...hoverEffect,
    [responsiveHeight]: {
      ...expandedStyle
    }
  };
});
```

src/components/Widgets/Widget.tsx
dennyscode: For that, I just updated the dependency array and it cleaned up some.

src/components/Widgets/Widget.style.ts
dennyscode: Thanks for the input @admsk2 !

Maybe we can have this discussion with @oktapodia and tcheee. 
Personally, I do not find the approach you suggested more readable but we should decide within our team what we see as best practise for such "more sophisticated" styled-components. Wdyt?

admsk2 (approved):

chybisov (changes requested):
Found this while testing

https://github.com/user-attachments/assets/b55d721d-0218-4e19-b203-9063d8ca81de

dennyscode:
> Found this while testing
> 
>  Screen.Recording.2024-08-20.at.12.49.56.PM.mov

[bug-not-reproducable-screen-capture.webm](https://github.com/user-attachments/assets/d342ba2a-d910-4728-b75b-9936251d135e)

Thanks Eugene! I believe this issue must have occured before we got to fix and merge into develop. I cannot reproduce it myself. Let me know if I am wrong and we still have the "flickering" or "jumping" of the widget.

PR #1150, opened 2024-07-19
fix: fix bg colours widget default

tcheee:


oktapodia (changes requested):
Strongly against it, that will create some side effects for sure

oktapodia (approved):
Approved

chybisov (approved):

dennyscode (approved):

PR #1160, opened 2024-07-24
refactor: strapi url generation

dennyscode:


oktapodia (changes requested):
Unfortunately adding more complexity than before without benefits :( Let's have a talk about it and your approach

PR #1172, opened 2024-07-28
fix: Discord link (prod)

admsk2:
```
Hi, we need to urgently replace the Discord link on our Jumper UI burger menu to the new discord invite link: https://discord.gg/jumperexchange. 

It was changed from our previous ‘ /LIFI’ and that currently redirects to another server which is not ours.
```

https://discord.com/channels/849912621360218112/1267141147248169013

### DIFF
--- a/src/const/urls.ts
+++ b/src/const/urls.ts
@@ -3,7 +3,7 @@ import type { SitemapPage } from '@/types/sitemap';
 export const JUMPER_URL = 'https://jumper.exchange';
 export const LIFI_URL = 'https://li.fi';
 export const EXPLORER_URL = 'https://explorer.li.fi';
-export const DISCORD_URL = 'https://discord.gg/lifi';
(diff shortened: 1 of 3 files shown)


chybisov (approved):

putzhammer (approved):

src/const/urls.ts
```diff
@@ -3,7 +3,7 @@ import type { SitemapPage } from '@/types/sitemap';
 export const JUMPER_URL = 'https://jumper.exchange';
 export const LIFI_URL = 'https://li.fi';
 export const EXPLORER_URL = 'https://explorer.li.fi';
-export const DISCORD_URL = 'https://discord.gg/lifi';
+export const DISCORD_URL = 'https://discord.gg/jumperexchange';
```
Addminus, line 6: Do we want to use this export here everywhere instead of redefining the URL everytime?

Addminus (approved):
Quick improvement for the future that does not block deployment

PR #1173, opened 2024-07-28
fix: Discord link (dev)

admsk2:
```
Hi, we need to urgently replace the Discord link on our Jumper UI burger menu to the new discord invite link: https://discord.gg/jumperexchange. 

It was changed from our previous ‘ /LIFI’ and that currently redirects to another server which is not ours.
```

https://discord.com/channels/849912621360218112/1267141147248169013


tcheee (approved):

src/const/urls.ts
```diff
@@ -3,7 +3,7 @@ import type { SitemapPage } from '@/types/sitemap';
 export const JUMPER_URL = 'https://jumper.exchange';
 export const LIFI_URL = 'https://li.fi';
 export const EXPLORER_URL = 'https://explorer.li.fi';
-export const DISCORD_URL = 'https://discord.gg/lifi';
+export const DISCORD_URL = 'https://discord.gg/jumperexchange';
```
chybisov, line 6: Can we please use this const everywhere we have this link, as per Adrian's comment on the other PR?

chybisov (changes requested):

src/const/urls.ts
admsk2: @chybisov `DISCORD_URL` is being already used, although there is another link with invite -- i just added the other one as `const` and replaced the link with that `const`

chybisov (approved):

PR #1183, opened 2024-07-30
feat: implement backend

oktapodia:


tcheee (approved):

src/hooks/useLoyaltyPass.ts
```diff
@@ -41,52 +35,28 @@ export const useLoyaltyPass = (): UseLoyaltyPassProps => {
...
-        // filter to remove loyalty pass from pda
-        const pdasWithoutLoyalty = pdas.filter((pda: PDA) => {
-          if (pda.dataAsset.title === 'LI.FI Loyalty Pass') {
-            points = pda.dataAsset.claim.points;
-            tier = pda.dataAsset.claim.tier;
-            return false;
-          }
-          return true;
-        });
+      const data = await res.json();
```
admsk2, line 44: let's add a pre-check to make sure `res` can be jsoned, something like this should be sufficient:

```
  if (!res.ok) {
    // handle error
 } 

  const data = await res.json();
  ```

PR #1239, opened 2024-08-13
style: fix blog formatting of quotes + instruction items

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-9653


src/components/Blog/CustomRichBlocks.tsx
```diff
@@ -214,7 +225,7 @@ export const CustomRichBlocks = ({
                         strikethrough={el.props.strikethrough}
                         underline={el.props.underline}
                         bold={el.props.bold}
-                        key={`blog-paragraph-line-${index}-${lineIndex}`}
+                        key={generateKey('blog-paragraph-line')}
```
oktapodia, line 228: This key might conflicts with another one as it does not have any variable in anymore

oktapodia (changes requested):
Just a quick comment

src/components/Blog/CustomRichBlocks.tsx
dennyscode: Thanks Sebastien! I have undone those changes - but out of curiosity: Does the generateKey not create a random key anyways, already preventing a clash with any existing key?

dennyscode:
**Handle "Quote" rich-text elements:**

Previously:
image

Fixed here:
image

dennyscode:
**Handle Instruction-Item colors on dark-mode:**

Previously:
image

Fixed here:
image

src/components/Blog/CustomRichBlocks.tsx
```diff
@@ -102,6 +102,17 @@ export const CustomRichBlocks = ({
           );
       }
     },
+    quote: ({ children }: any) => {
```
oktapodia, line 105: Would be good to type it

src/components/Blog/CustomRichBlocks.tsx
oktapodia: Hum, you are actually right

admsk2 (approved):

src/components/Blog/CustomRichBlocks.tsx
dennyscode: Thanks! I added some more types now for the quotes + paragraph elements.

oktapodia (approved):
LGTM

PR #1253, opened 2024-08-19
chore: use actual theme typography within widget config

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-9486

image

### DIFF
--- a/src/components/Widgets/useWidgetTheme.ts
+++ b/src/components/Widgets/useWidgetTheme.ts
@@ -1,12 +1,12 @@
-import type { Breakpoint } from '@mui/material';
-import { useTheme } from '@mui/material';
-import { useTheme as useNextTheme } from 'next-themes';
-import type { PartnerTheme } from 'src/types/strapi';
-import { useEffect } from 'react';
-import { deepmerge } from '@mui/utils';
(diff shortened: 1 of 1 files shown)


admsk2 (approved):

src/components/Widgets/useWidgetTheme.ts
```diff
@@ -21,10 +21,10 @@ export const useWidgetTheme = (): PartnerTheme => {
     config: {
       appearance: theme.palette.mode,
       theme: {
-        //todo: fix typography
-        // typography: {
-        //   fontFamily: theme.typography,
-        // },
+        // @ts-expect-error
```
chybisov, line 24: What error do we have here?

chybisov (approved):

src/components/Widgets/useWidgetTheme.ts
dennyscode: @chybisov 
image

It´s complaining because of our custom typographies which is likely not included in a partner-theme.
> brandHeaderXLarge, bodyXLarge, bodyLargeStrong, bodyLarge, and 6 more

PR #1260, opened 2024-08-26
refactor: Effigy icon fetching

admsk2:
Effigy fallback is using svg image from public folder instead of third-party url


src/components/ProfilePage/AddressBox/AddressBox.tsx
```diff
@@ -52,12 +52,12 @@ export const AddressBox = ({ address, isEVM, imageLink }: AddressBoxProps) => {
     ? imageLink
     : address && isEVM
       ? `https://effigy.im/a/${address}.png`
```
tcheee, line 54: @admsk2 can you implement a logic so when this url is not reponsive we show the default one pls?

oktapodia, line 54: Seems to be duplicated with the above, would be good to have its own component or helper

PR #1272, opened 2024-08-28
chore: useSession hook; events url change;

admsk2:


src/hooks/useSession.ts
```diff
@@ -2,22 +2,24 @@
...
+  const [session, setSession] = useState<string>(() => {
     if (
       typeof window !== 'undefined' &&
       typeof sessionStorage !== 'undefined'
     ) {
-      return sessionStorage.getItem('session_id') || undefined;
+      return sessionStorage.getItem('session_id') || uuidv7();
     }
-    return undefined;
+    return uuidv7();
```
tcheee, line 13: why?

admsk2: to ensure we always have a value in session

PR #1279, opened 2024-09-02
feat: Add portfolio

oktapodia:


next.config.mjs
```diff
@@ -23,6 +23,36 @@ const nextConfig = {
         port: '',
         pathname: '/a/**',
       },
+      {
+        protocol: 'https',
+        hostname: 'raw.githubusercontent.com',
```
oktapodia, line 28: Might lead to a security issue if are not strict enough (mainly used to get icons from angleprotocol, sushiswap and trustwallet)

tcheee: We must be stricted.

PR #1281, opened 2024-09-03
chore: restore welcome screen

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-9576


src/components/Menus/ThemeSubMenu/useThemeMenuContent.tsx
```diff
@@ -33,7 +36,15 @@ export const useThemeMenuContent = () => {
         [TrackingEventParameter.SwitchedTemplate]: theme,
       },
     });
+    if (!(theme === 'system' || theme === 'light' || theme === 'dark')) {
```
oktapodia, line 39: better to use an include to for readability

src/components/Widgets/Widget.style.ts
```diff
@@ -13,15 +13,98 @@ export const WidgetWrapper = styled(Box, {
...
-    '&:hover': {
-      marginTop: 0,
-    },
-  }),
   zIndex: 2,
-  // [theme.breakpoints.up('sm' as Breakpoint)]: {
-  //   maxWidth: 416,
-  // },
+
+  '> div': {
```
oktapodia, line 18: Not fan of this as it is only related to welcomeScreen, can you add it to the welcomescreen component itself?

src/hooks/useWelcomeScreen.ts
```diff
@@ -1,25 +1,42 @@
...
+    () =>
+      activeTheme === 'default' ||
+      activeTheme === 'light' ||
+      activeTheme === 'dark' ||
+      activeTheme === 'system' ||
+      cookie.theme === 'default' ||
+      cookie.theme === 'light' ||
+      cookie.theme === 'dark' ||
+      cookie.theme === 'system' ||
+      !cookie.theme,
```
oktapodia, line 36: Would be better to simplify it

```suggestion
() => {
  const validThemes = ['default', 'light', 'dark', 'system'];
  return validThemes.includes(activeTheme) || validThemes.includes(cookie.theme) || !cookie.theme;
},
```

src/hooks/useWelcomeScreen.ts
```diff
@@ -29,6 +46,9 @@ export const useWelcomeScreen = (
   }, [cookie.welcomeScreenClosed, isMultisigSigner]);
 
   const updateState = (closed: boolean) => {
+    if (!enabled) {
```
oktapodia, line 49: Is this check needed?

src/components/Menus/ThemeSubMenu/useThemeMenuContent.tsx
dennyscode: Good idea!

src/hooks/useWelcomeScreen.ts
dennyscode: It´s meant to ensure that state changes on the welcome-screen are only enabled if the theme supports the welcome-screen.

src/components/Widgets/Widget.style.ts
dennyscode: I don´t think we can move this easily to the WelcomeScreen itself as they´re not directly related in the tree as children/parents.

This ugly hack allows to directly style the widget via the widget-wrapper. It´s mainly needed for positioning.
If you know of a cleaner way, I´d love to hear. Maybe the "container" prop within the widgetTheme is a better choice??

admsk2 (approved):

tcheee (approved):

PR #1283, opened 2024-09-03
chore: add track events to quest-card + mission-cta

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-9873


dennyscode:
1.) Click on quest card
image

2.) Click on a mission cta: There is also null values being sent, which is why I added the title (?).
Anyways, I guess if we could add id´s to the CTA, it´s better to track: (check strapi screenshot)
image
image

admsk2 (approved):

tcheee (approved):

PR #1301, opened 2024-09-11
fix: nl2br warning

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-10008

### DIFF
--- a/src/utils/newlineToBr.ts
+++ b/src/utils/newlineToBr.ts
@@ -1,8 +1,9 @@
+import type { ReactElement } from 'react';
 import { createElement, Fragment } from 'react';
 
-const newlineToBr = (str: any) => {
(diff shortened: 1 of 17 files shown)


src/utils/newlineToBr.ts
```diff
@@ -0,0 +1,16 @@
+import { createElement, Fragment } from 'react';
+
+const newlineToBr = (str: any) => {
```
oktapodia, line 3: use unknown instead

oktapodia (approved):
LGTM

src/utils/newlineToBr.ts
dennyscode: Yup! Much better, thanks !!!

tcheee (approved):

PR #1309, opened 2024-09-13
fix: fingerprint

oktapodia:


tcheee (approved):

dennyscode:
Thanks Sebastien for that fix! Reducing the number of dependencies in the dep-array makes sense and I should have done this in the first place. But did we not tackle this to reduce the number of re-renderings: 

Context:
- the approach I took with useContext only renders once ( + 1 because of strict).
Screenshot 2024-09-16 at 08 52 24

Hook:
- with the updated hook [setFingerprint] it renders 5 times ( + 5 because of strict)
Screenshot 2024-09-16 at 08 51 25

Definitely both are an improvement compared to what we had before:
image


Why do you prefer the hook over the context?

dennyscode (approved):

PR #1316, opened 2024-09-18
feat: switch Strapi

tcheee:


oktapodia (changes requested):
Can you use the env variable instead of hardcoding it please? 🙏

admsk2 (approved):

oktapodia (approved):
LGTM

PR #1325, opened 2024-09-19
chore: adjust TAI tokens

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-9999


src/components/Widgets/Widget.tsx
```diff
@@ -1,7 +1,6 @@
 'use client';
 import { ClientOnly } from '@/components/ClientOnly';
 import { MultisigWalletHeaderAlert } from '@/components/MultisigWalletHeaderAlert';
-import { widgetConfig } from '@/config/widgetConfig';
```
oktapodia, line 4: You cannot do that otherwise the widget won't be initialized anymore

src/components/Widgets/Widget.tsx
```diff
@@ -199,8 +202,7 @@ export function Widget({
       buildUrl: true,
       // insurance: true,
       integrator: integratorStringByType,
-      tokens:
-        partnerName === ThemesMap.Memecoins && tokens ? { allow: tokens } : {},
```
oktapodia, line 203: Are you sure about removing this? is that part of the story? memecoins page won't work anymore

src/config/widgetConfig.ts
```diff
@@ -12,25 +10,3 @@ createConfig({
...
-  tokens: {
-    allow: [
-      // {
-      //   // Wrapped SOL
-      //   chainId: ChainId.SOL,
-      //   address: 'So11111111111111111111111111111111111111112',
-      // },
-      {
-        // USDC
-        chainId: ChainId.SOL,
```
oktapodia, line 26: Is that expected to remove sol?

oktapodia (changes requested):
Couple of comments

tcheee (approved):
seems ok to me

src/components/Widgets/Widget.tsx
dennyscode: Is that the case? Here, we have just imported the `widgetConfig`. The widgetConfig only included a set of tokens and the problem with that was, that tokens are being overwritten again in the Widget.tsx.

If "not importing" the widgetConfig does stop the initialization of the widget, we must find a solution. But it seemed to work just fine.

src/components/Widgets/Widget.tsx
dennyscode: This is now being handled differently: The tokens which we want to add (such as for TAI) go into the` tokens.ts`. 
Those meme-coin tokens will only be added for certain pages and we still have that if condition. Please take a look. Let me know, if that should be done in a useMemo or somehow else.

src/config/widgetConfig.ts
dennyscode: Yes, I tried it with that and it is different from what we have in prod - because it was overwritten anyways. We can take another look and see if we want to add **only** USDC for Solana chain tokens. The way it is right now in prod is to see all @tcheee Wdyt?

admsk2 (approved):

oktapodia (approved):
LGTM

PR #1346, opened 2024-09-27
fix: Effigy fallback;

admsk2:


src/hooks/useImageStatus.ts
```diff
@@ -0,0 +1,26 @@
+import { useState, useEffect } from 'react';
+
+const useImageStatus = (imageUrl: string) => {
+  const [isImageValid, setIsImageValid] = useState<boolean | null>(null);
+
+  useEffect(() => {
+    const checkImage = async () => {
+      try {
+        const response = await fetch(imageUrl);
+        if (response.status <= 299) {
```
oktapodia, line 10: Not working in the case of 400, 401, 403, 404, 500

src/components/ProfilePage/AddressBox/AddressBox.tsx
```diff
@@ -14,6 +14,9 @@ import {
   PassImageBox,
   ProfileIconButton,
 } from './AddressBox.style';
+import useImageStatus from 'src/hooks/useImageStatus';
+
+const DEFAULT_IMAGE = '/default_effigy.svg';
```
oktapodia, line 19: Would be good to refactorize it to avoid duplication

tcheee: +1

tcheee (approved):

tcheee (approved):

PR #1352, opened 2024-09-30
fix: navbar blur + error fixes

dennyscode:
For our beta-testing branch I have opened a new PR which fixes the blur of the top-navigation. I came across the following errors:

1.) ✅ 
> MenuDesktop.tsx:56 [Intervention] Ignored attempt to cancel a touchstart event with cancelable=false, for example because scrolling is in progress and cannot be interrupted.

--> To reproduce: This happens whenever I am opening a menu: Like opening MainMenu via hamburger icon or opening WalletMenu

image

2.) ✅ 
> redirect-boundary.js:57 Uncaught Error: Invalid src prop (https://assets.gravity.xyz/token_logo.png) on `next/image`, hostname "assets.gravity.xyz" is not configured under images in your `next.config.js`
> See more info:

--> To reproduce: Open wallet-menu. I keep on stumbling over this and whenever I check out the branch, I have to enable to url in the `next.config.mjs`. But I am wondering why only I run into this? Is it an extension on my behalf that causes it?

`      {
        protocol: 'https',
        hostname: 'assets.gravity.xyz',
        port: '',
        pathname: '/**',
      },`
image

3.) ✅ 
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

> Check the render method of `ForwardRef(Tooltip)`.
>     at TotalBalanceIconButton (webpack-internal:///(app-pages-

Screenshot 2024-09-30 at 09 52 52

4.) ✅ 
> app-index.js:33 MUI: The children component of the Tooltip is not forwarding its props correctly.
> Please make sure that props are spread on the same element that the ref is applied to.


dennyscode:
Couple more I had seen in my logs:

image

tcheee (approved):

PR #1355, opened 2024-09-30
chore: Vertical navbar refactor; Navbar XP only desktop;

admsk2:


tcheee:
please add tooltip on the left of exchange and gas when hover

PR #1399, opened 2024-10-11
chore: show sol address and improve empty states

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-10420

### DIFF
--- a/src/utils/getAddressLabel.ts
+++ b/src/utils/getAddressLabel.ts
@@ -1,0 +1,23 @@
+import type { GetEnsNameReturnType } from 'wagmi/actions';
+import { walletDigest } from './walletDigest';
+interface getAddressLabelProps {
+  isSuccess: boolean;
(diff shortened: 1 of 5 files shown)


src/utils/getAddressLabel.ts
```diff
@@ -0,0 +1,23 @@
...
+  address,
+}: getAddressLabelProps): string => {
+  if (isSuccess && ensName) {
+    return String(ensName).length > 20
+      ? `${String(ensName).slice(0, 13)}...eth`
+      : ensName;
+  } else if (address) {
+    return walletDigest(address);
+  } else {
+    return '0x0000...0000';
```
oktapodia, line 21: please use walletDigest function for this, seems to be missing 1 char

dennyscode: Good idea! Please give it another look as I refactored that part

tcheee (approved):

PR #1406, opened 2024-10-15
fix: sorting + filtering on similar and recent posts

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-10322


tcheee (approved):

src/app/[lng]/(infos)/learn/[slug]/page.tsx
```diff
@@ -61,11 +61,10 @@ export default async function Page({ params }: { params: { slug: string } }) {
   const { activeThemeMode } = getCookies();
 
   const currentTags = (
-    article.data as BlogArticleData
+    article?.data.data[0] as BlogArticleData
```
Comment (admsk2) [inline src/app/[lng]/(infos)/learn/[slug]/page.tsx:64]: how about `article?.data?.data[0]` and below `article?.data?.data[0]?`

admsk2 (approved):

PR #1407, opened 2024-10-15
chore: UI fix

tcheee:


src/components/Menus/WalletMenu/WalletCard.style.ts
```diff
@@ -13,19 +13,19 @@ export const WalletAvatar = styled(Avatar)(({ theme }) => ({
...
     objectFit: 'contain',
-    backgroundColor: theme.palette.white.main,
   },
   // mask: avatarMask32,
 }));
 
 export const WalletCardContainer = styled(Container)(({ theme }) => ({
   boxShadow: '0px 1px 4px 0px rgba(0, 0, 0, 0.04)',
-  padding: '24px',
+  padding: theme.spacing(2, 1, 2, 1),
```
dennyscode:
```suggestion
  padding: theme.spacing(2, 1),
```

oktapodia (approved):
LGTM

src/components/Portfolio/CircularProgress/RefreshIcon.tsx
```diff
@@ -0,0 +1,96 @@
...
+        }}
+      >
+        <Box
+          sx={{
+            display: 'grid',
+            position: 'relative',
+            placeItems: 'center',
+            width: 24,
+            height: 24,
+          }}
```
dennyscode, line 66: Let´s put this right into a styled component?

src/components/Portfolio/Portfolio.styles.ts
```diff
@@ -90,37 +92,27 @@ export const SmallAvatar = styled(Avatar)(({ theme }) => ({
...
+export const NoTokenImageBox = styled(Box)(({ theme }) => ({
+  backgroundColor: 'grey',
+  borderRadisu: '50%',
+  height: 40,
+  width: 40,
+  display: 'flex',
+  justifyContent: 'center',
+  alignContent: 'center',
+  alignItems: 'center',
+}));
```
dennyscode, line 118: Not 100% knowing, but could this also be replaced by a Skeleton?

src/components/Portfolio/PortfolioTokensSkeleton.tsx
```diff
@@ -0,0 +1,83 @@
...
+                <Skeleton variant="circular" width={40} height={40} />
+              </Badge>
+            </Box>
+            <Box
+              sx={{
+                width: '100%',
+                display: 'flex',
+                justifyContent: 'space-between',
+                alignContent: 'center',
+                alignItems: 'center',
```
dennyscode, line 46: Looking at this quickly, sure I´d ask to put this into a styled-component.

But on top of that, I am wondering if it makes sense to come up with a "default container" that can be re-used? But maybe it´s over the top?? See, I found multiple components which share a similar setup:

image
image
image
...sure there is more.

Wdyt? Is it any useful to create a "default" styled-component that we can inherit from and refactor?

src/utils/optionalStrShortener.ts
dennyscode, line 1: I would remove the "optional" from naming. It´s already pretty clear - we are open to use it or not, so that "optional" is not adding more value and does not make it more clear - instead it seems confusing.

ChatGPT came up with some alternative wording:
- truncateString
- truncateWithEllipsis
- stringShortener
- limitStringLength

Wdyt?
ellipsisIfExceeds

dennyscode (changes requested):

src/components/Menus/WalletMenu/WalletCard.style.ts
tcheee: wip.

src/components/Portfolio/CircularProgress/RefreshIcon.tsx
tcheee: copy paste from widget repo tbh. But yes let's do it.

src/components/Portfolio/Portfolio.styles.ts
tcheee: it is not a skeleton but a Box when no image.

src/components/Portfolio/PortfolioTokensSkeleton.tsx
tcheee: Ok I'll make a styled component. agree with you on the component. Will create a ticket.

src/utils/optionalStrShortener.ts
tcheee: I mean i add the optional because it is either or given the length, I'll rename it.

dennyscode: Ahh, yes makes sense to me - but still. let´s rename

src/components/Portfolio/CircularProgress/RefreshIcon.tsx
dennyscode: There is this one time, we can do better than Eugene! Let´s take the chance 😄

src/components/Menus/WalletMenu/WalletCard.style.ts
```diff
@@ -13,25 +13,25 @@ export const WalletAvatar = styled(Avatar)(({ theme }) => ({
...
   display: 'flex',
   background: theme.palette.surface2.main,
   borderRadius: '16px',
-  [theme.breakpoints.down('sm' as Breakpoint)]: {
-    padding: '10px',
-  },
+  // padding: theme.spacing(2, 1, 2, 1),
+  // [theme.breakpoints.down('sm' as Breakpoint)]: {
+  //   padding: '10px',
+  // },
```
dennyscode, line 34: Let´s straight up remove this

dennyscode (changes requested):

dennyscode (changes requested):
Ughhh, found another one!

image

PR #1495, opened 2024-11-07
feat: Make profile page shareable

oktapodia:
Example:
image


src/app/api/profile/[walletAddress]/route.tsx
```diff
@@ -0,0 +1,190 @@
...
+    address: walletAddress,
+    points: data.sum,
+    level: data.level,
+    pdas: data.walletRewards,
+  };
+}
+
+export const maxDuration = 60;
+
+export async function GET(
```
Comment (tcheee) [inline src/app/api/profile/[walletAddress]/route.tsx:52]: file name is weird, let's try to rename

src/components/ProfilePage/MerkleRewards.tsx
```diff
@@ -0,0 +1,52 @@
+import {
```
tcheee, line 1: why?

src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx
```diff
@@ -0,0 +1,32 @@
...
+    },
+  };
+}
+
+export default async function Page({
+  params,
+}: {
+  params: { walletAddress: string };
+}) {
+  return <ProfilePage walletAddress={params.walletAddress} isPublic />;
```
Comment (tcheee) [inline src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx:31]: why not params?.walletAddress?

src/providers/ProfileProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export const ProfileProvider: React.FC<
+  PropsWithChildren<Pick<ProfileProps, 'walletAddress' | 'isPublic'>>
+> = ({ children, walletAddress, isPublic }) => {
+  return (
+    <ProfileContext.Provider
+      value={{
+        walletAddress,
+        isPublic: isPublic ?? false,
+        isEVMAddress: isEVMAddress(walletAddress),
```
tcheee, line 25: it should work with Solana address as well, why do we have this boolean?

admsk2 (approved):

src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx
Comment (oktapodia) [reply src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx:31]: this page is rendered only if params.walletAddress exists in the path (it is not optional in the url)

src/providers/ProfileProvider.tsx
oktapodia: Because I had to extract this information from the useAccount hook (used on some components, mainly effigy related + `useLoyaltyPass` hook (actually surprising that it is used there)

tcheee (approved):

PR #1496, opened 2024-11-07
feat: Remove path params when swapping tabs

oktapodia:


src/components/Menus/VerticalMenu/useVerticalTabs.tsx
```diff
@@ -17,11 +17,7 @@ export const useVerticalTabs = () => {
   const { t } = useTranslation();
 
   const handleClickTab = (tab: string) => () => {
-    const searchParams = new URLSearchParams(window.location.search);
-    let path = searchParams.toString();
-    path = path.startsWith('?') ? path.substring(1) : path;
-
-    router.push(`/${tab}?${path}`);
+    router.push(`/${tab}`);
```
tcheee, line 20: can you keep everything and just remove the "toToken" from the path  pls @oktapodia

dennyscode (approved):

dennyscode (changes requested):
Not sure! Although I like the approach you took, this might be misleading to our users!

Not all chains are supported on the Refuel-Widget: If we keep the SearchParams, it also suggests those ones. I think either removing it, or keeping searchParams elsewhere (zustand temporary or persistent) for each tab would be the cleaner way. Wdyt?

image

dennyscode (approved):

PR #1516, opened 2024-11-14
chore: add tokenSearch and routeSelected tracking

tcheee:


src/utils/isValidEvmOrSvmAddress.ts
```diff
@@ -0,0 +1,51 @@
...
+      addressType: 'EVM',
+    };
+  } else if (isValidSVMAddress(value)) {
+    return {
+      isValid: true,
+      addressType: 'SVM',
+    };
+  }
+
+  return { isValid: false };
```
admsk2, line 50: `addressType` is undefined in that case, but this should not be a problem i guess

admsk2 (approved):

dennyscode (approved):

PR #1524, opened 2024-11-18
feat: Use backend to fetch fc users

oktapodia:


tcheee (approved):
lgtm

admsk2 (approved):

src/hooks/usePersonalizedFeatureCards.ts
```diff
@@ -1,50 +1,62 @@
...
   (): UsePersonalizedFeatureCardsProps => {
     const { account } = useAccount();
 
+    const { data: fcCardData, isSuccess: fcCardDataIsSuccess } = useQuery({
+      queryKey: ['jumperUser', account?.address],
+      queryFn: async () => {
+        const response = await fetch(
+          getFeatureCardsEndpoint(account?.address!),
+        );
+        const result = await response.json();
```
admsk2, line 26: i think there should be some `try/catch` or check if `response.ok` in case backend throws an unexpected error

admsk2 (approved):

PR #1527, opened 2024-11-19
feat: Add redirectUrl to strapi blog article

oktapodia:


tcheee:
I think as the blog are not live anymore and won't be. We need to have a redirection when arriving on the page like inside a useEffect or equivalent, no?

src/components/Blog/BlogArticleCard/BlogArticleCard.tsx
```diff
@@ -52,7 +52,10 @@ export const BlogArticleCard = ({
   };
   return (
     <Link
-      href={`${JUMPER_LEARN_PATH}/${article.attributes.Slug}`}
+      href={
```
tcheee, line 55: won't be enough imho.

tcheee (changes requested):

tcheee (approved):

dennyscode (approved):

PR #1566, opened 2024-12-03
feat: zap page

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-11136


src/components/Zap/ZapAction/Zapwidget.tsx
dennyscode, line 1: @oktapodia @tcheee This is a copy of the "Widget.tsx" component -> The only change is that we do not want the box-shadow within the widget-theme. First, I started out to create a partner theme to be used for the "zap" layout but ran into the problem that partner-themes do not support multiple themeModes (light + dark).

src/components/Zap/ZapAction/Zapwidget.tsx
```diff
@@ -0,0 +1,268 @@
...
+    container: {
+      ...widgetTheme.config.theme?.container,
+      boxShadow: 'unset',
+      '@media (min-width:600px)': {
+        // @ts-ignore or @ts-expect-error
+        ...widgetTheme.config.theme?.container['@media (min-width:600px)'],
+        boxShadow: 'unset',
+      },
+    },
+  };
```
dennyscode, line 114: This is the part where I am overriden the boxShadow. It returned a ts-error which I am ignoring by using `// @ts-ignore or @ts-expect-error`. Wdyt?

src/app/[lng]/zap/[slug]/page.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const questData = data.data[0] || undefined;
+    if (!questData) {
+      throw new Error();
+    }
+
+    const openGraph: Metadata['openGraph'] = {
+      title: 'Jumper | Berachain',
+      description: 'Dive into the Berachain universe!',
+      siteName: siteName,
+      url: `${getSiteUrl()}/berachain/explore/${params.slug}`,
```
Comment (oktapodia) [inline src/app/[lng]/zap/[slug]/page.tsx:26]: Are you sure about this url? it seems to be /zap/[slug]

src/app/[lng]/zap/[slug]/page.tsx
```diff
@@ -0,0 +1,67 @@
...
+      description: `Description of ${questData.attributes.Title}`,
+      alternates: {
+        canonical: `${getSiteUrl()}/berachain/explore/${params.slug}`,
+      },
+      twitter: openGraph,
+      openGraph,
+    };
+  } catch (err) {
+    return {
+      title: `Jumper Learn | ${sliceStrToXChar(params.slug.replaceAll('-', ' '), 45)}`,
```
Comment (oktapodia) [inline src/app/[lng]/zap/[slug]/page.tsx]: Not sure this is a learn page

src/app/[lng]/zap/[slug]/page.tsx
```diff
@@ -0,0 +1,67 @@
...
+      description: `This is the description for the article "${params.slug.replaceAll('-', ' ')}".`,
+    };
+  }
+}
+
+export default async function Page({ params }: { params: { slug: string } }) {
+  const { data } = await getQuestBySlug(params.slug);
+  const questData = (data.data[0] as ExtendedQuest) || undefined;
+
```
Comment (oktapodia) [inline src/app/[lng]/zap/[slug]/page.tsx:58]: Do not forget to return notFound() if it does not exists

src/app/[lng]/zap/[slug]/page.tsx
```diff
@@ -0,0 +1,67 @@
...
+    return {
+      title: `Jumper Learn | ${sliceStrToXChar(params.slug.replaceAll('-', ' '), 45)}`,
+      description: `This is the description for the article "${params.slug.replaceAll('-', ' ')}".`,
+    };
+  }
+}
+
+export default async function Page({ params }: { params: { slug: string } }) {
+  const { data } = await getQuestBySlug(params.slug);
+  const questData = (data.data[0] as ExtendedQuest) || undefined;
```
Comment (oktapodia) [inline src/app/[lng]/zap/[slug]/page.tsx:57]: Would be better to type is within getQuestBySlug

src/components/AccordionFAQ/AccordionFAQ.style.ts
```diff
@@ -1,85 +1,150 @@
...
+    gap: theme.spacing(1),
+  }),
+);
+
+export const Accordion = styled(MuiAccordion)(({ theme }) => ({
+  fontFamily: 'inherit',
+  borderRadius: '16px',
+  padding: theme.spacing(2, 3),
+  width: '100%',
+  background: alpha(theme.palette.text.primary, 0.08),
```
oktapodia, line 75: Would be good to extract it to the theme to avoid duplication

src/components/AccordionFAQ/AccordionFAQ.style.ts
```diff
@@ -1,85 +1,150 @@
...
+  '&.Mui-expanded': {
+    margin: 0,
+  },
   '&:last-of-type': {
-    marginBottom: show ? theme.spacing(2) : 0,
+    borderRadius: '16px',
+  },
+
+  '&:hover': {
+    background: alpha(theme.palette.text.primary, 0.16),
```
oktapodia, line 89: Would be good to extract it to the theme to avoid duplication

src/components/Widgets/Widget.style.ts
```diff
@@ -12,106 +12,115 @@ const DEFAULT_WIDGET_TOP_HOVER_OFFSET = 24;
...
         },
-      },
-      // positioning of widget on mobile-screens from 900px height
-      [`@media screen and (min-height: 900px)`]: {
-        marginTop: `calc( ${DEFAULT_WIDGET_TOP_OFFSET_VAR} - ${HeaderHeight.MD}px)`,
-        '&:hover': {
-          marginTop: `calc( ${DEFAULT_WIDGET_TOP_OFFSET_VAR} - ${HeaderHeight.MD}px - ${DEFAULT_WIDGET_TOP_HOVER_OFFSET}px )`,
+
+        // positioning of widget on mobile-screens from 700px height
+        [`@media screen and (min-height: 700px)`]: {
```
oktapodia, line 61: Would be good to use breakpoints value instead to avoid regresions

src/components/Zap/BackButton/BackButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <ArrowBackIcon
+          sx={{ color: '#FFFFFF', width: '16px', height: '16px' }}
+        />
+        <Typography
+          fontSize="14px"
+          lineHeight="14px"
+          fontWeight={500}
+          marginLeft={'8px'}
+        >
+          {String(title || t('navbar.navbarMenu.profile')).toUpperCase()}
```
oktapodia, line 30: Not 100% sure about it but if the translation does not exists, it will be null or undefined, toUpperCase() will then crash the app

src/components/Zap/ZapAction/ZapAction.tsx
```diff
@@ -0,0 +1,288 @@
...
+          }}
+        >
+          <Tabs
+            data={tabs}
+            value={tab}
+            ariaLabel="zap-switch-tabs"
+            containerStyles={containerStyles}
+            tabStyles={tabStyles}
+          />
+          {tab === 0 ? (
```
oktapodia, line 269: A switch would be better, else case is not needed

src/components/Zap/ZapAction/ZapActionFaqAccordionHeader.tsx
```diff
@@ -0,0 +1,10 @@
+import { Typography } from '@mui/material';
+import { AccordionHeader } from 'src/components/AccordionFAQ';
+
+export const ZapActionFaqAccordionHeader = () => {
+  return (
+    <AccordionHeader sx={{ margin: 0, marginBottom: '4px', marginLeft: '8px' }}>
+      <Typography variant="title2XSmall">How to participate</Typography>
```
oktapodia, line 7: Would need to be a translation

src/components/Zap/ZapAction/Zapwidget.tsx
oktapodia: Would be much better to refactorize the Widget.tsx instead of fully duplicating it :(

oktapodia (changes requested):
Couple of comments

src/components/Zap/ZapAction/Zapwidget.tsx
dennyscode: Good idea running that extra mile instead of taking the shortcut I did!
I added a `useWidgetConfig`-hook -> We had such in the past as well, it now also takes a "customWidgetTheme" prop and is merged with the "source" widgetTheme using `deepmerge` by mui -> Wdyt?

src/components/Zap/ZapAction/Zapwidget.tsx
dennyscode: It´s deprecated! Now, that magic happens within "useWidgetConfig"

src/components/Widgets/Widget.style.ts
dennyscode: This is magic voodoo happening. Please note, it´s applying styles to the `min-height` which we usually don´t have as default breakpoints. It´s needed to position the widget accordingly while the welcomeScreen is open.

src/components/Zap/BackButton/BackButton.tsx
dennyscode: Good point! Not sure if it´s a good manner, but I did it like so now: 'Jumper Profile' as last resort!

`String(title || t('navbar.navbarMenu.profile') || 'Jumper Profile',).toUpperCase()}`

src/components/Zap/ZapAction/ZapActionFaqAccordionHeader.tsx
dennyscode: Content on that page is being pulled from strapi which is also not multi-language, yet.

src/components/Zap/ZapAction/ZapAction.tsx
dennyscode: Alright, added a switch statement

src/components/AccordionFAQ/AccordionFAQ.style.ts
dennyscode: This can be a topic for another day, a different ticket to clean this up and manage the general usage of such within the theme.

src/components/AccordionFAQ/AccordionFAQ.style.ts
dennyscode: Same, I think this is to be done later. Wdyt? @tcheee

src/app/[lng]/zap/[slug]/page.tsx
Comment (dennyscode) [reply src/app/[lng]/zap/[slug]/page.tsx]: @tcheee Can we align on the meta information?

src/app/[lng]/zap/[slug]/page.tsx
Comment (dennyscode) [reply src/app/[lng]/zap/[slug]/page.tsx]: Done, please have another look.

src/app/[lng]/zap/[slug]/page.tsx
Comment (tcheee) [reply src/app/[lng]/zap/[slug]/page.tsx]: should be good by now

src/components/AccordionFAQ/AccordionFAQ.style.ts
tcheee: agree

src/components/AccordionFAQ/AccordionFAQ.style.ts
tcheee: agree

src/components/Zap/ZapAction/ZapActionFaqAccordionHeader.tsx
tcheee: agree

src/app/[lng]/quests/[slug]/page.tsx
```diff
@@ -1,22 +1,63 @@
...
+      description: `This is the description for the quest "${params.slug.replaceAll('-', ' ')}".`,
+    };
+  }
+}
 
 export default async function Page({ params }: { params: { slug: string } }) {
   const { data, url } = await getQuestBySlug(params.slug);
-
-  if (!data?.data?.[0]) {
+  if (!data) {
```
Comment (oktapodia) [inline src/app/[lng]/quests/[slug]/page.tsx:59]: Might not be enough, data can be returned and data.data -> [] but will still not go through

src/app/[lng]/superfest/[slug]/page.tsx
```diff
@@ -44,10 +44,8 @@ import { notFound } from 'next/navigation';
 
 export default async function Page({ params }: { params: { slug: string } }) {
   const { data, url } = await getQuestBySlug(params.slug);
-
-  if (!data?.data?.[0]) {
+  if (!data) {
```
Comment (oktapodia) [inline src/app/[lng]/superfest/[slug]/page.tsx:47]: Same as above

src/app/[lng]/zap/[slug]/page.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const { data } = await getQuestBySlug(params.slug, 'ExtendedQuest');
+
+  if (!data) {
+    return notFound();
+  }
+
+  const questData = data;
+
+  if (questData) {
+    const protocolDetails = zapMarkets.filter(
```
Comment (oktapodia) [inline src/app/[lng]/zap/[slug]/page.tsx:68]: Dangerous, if find does not find any protocolDetails, it will be null, [0] will crash the app

src/app/lib/getQuestBySlug.ts
```diff
@@ -18,5 +33,12 @@ export async function getQuestBySlug(slug: string) {
 
   const data = await res.json(); // Extract data from the response
 
-  return { data, url: apiBaseUrl }; // Return a plain object
+  if (!data || !Array.isArray(data.data) || data.data.length === 0) {
```
oktapodia, line 36: This part seems weird and does not seem to achieve anything, let's have a talk about it

src/components/Quests/QuestPage/BackButton/BackButton.tsx
```diff
@@ -35,7 +35,9 @@ export const BackButton = ({ path, title }: BackButtonProps) => {
           fontWeight={500}
           marginLeft={'8px'}
         >
-          {String(title || t('navbar.navbarMenu.profile')).toUpperCase()}
+          {String(
```
oktapodia, line 38: String casting should not be needed

src/components/Zap/BackButton/BackButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+  path?: string;
+  title?: string;
+}
+
+export const BackButton = ({ path, title }: BackButtonProps) => {
+  const { t } = useTranslation();
+
+  return (
+    <Link
+      href={path ?? '/profile'}
```
oktapodia, line 17: Use the urls const instead

src/components/Zap/ZapAction/ZapAction.tsx
```diff
@@ -0,0 +1,265 @@
...
+}
+
+export const ZapAction = ({ market, detailInformation }: ZapActionProps) => {
+  const [tab, setTab] = useState(0);
+  const { setSnackbarState } = useMenuStore((state) => state);
+  const { t } = useTranslation();
+  const baseUrl = getStrapiBaseUrl();
+  const theme = useTheme();
+
+  const containerStyles = {
```
oktapodia, line 52: Why no styled co,ponent?

src/components/Zap/ZapAction/Zapwidget.tsx
```diff
@@ -0,0 +1,21 @@
+'use client';
+import { Widget } from 'src/components/Widgets/Widget';
+import type { WidgetProps } from 'src/components/Widgets/Widget.types';
+
+export function ZapWidget({ starterVariant }: WidgetProps) {
+  const customWidgetTheme = {
+    container: {
+      boxShadow: 'unset',
+      '@media (min-width:600px)': {
```
oktapodia, line 9: Use theme.breakpoint instead

src/hooks/useWidgetConfig.ts
```diff
@@ -0,0 +1,231 @@
...
+  // load environment config
+  const config: WidgetConfig = useMemo((): WidgetConfig => {
+    let rpcUrls = {};
+    try {
+      rpcUrls = {
+        ...JSON.parse(process.env.NEXT_PUBLIC_CUSTOM_RPCS),
+        ...publicRPCList,
+      };
+    } catch (e) {
+      if (process.env.DEV) {
```
oktapodia, line 104: Dangerous there, DEV is not a standard env variable, error will be obfuscated all time and not throwing anything (making it harder to debug), just remove this if

src/components/Zap/BackButton/BackButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+import Link from 'next/link';
+import { useTranslation } from 'react-i18next';
+import { BackButtonStyles } from './BackButton.style';
+
+interface BackButtonProps {
+  path?: string;
+  title?: string;
+}
+
+export const BackButton = ({ path, title }: BackButtonProps) => {
```
oktapodia, line 12: Why not using the back button already existing above?

oktapodia (changes requested):
Couple of comments

src/app/[lng]/quests/[slug]/page.tsx
Comment (dennyscode) [reply src/app/[lng]/quests/[slug]/page.tsx:59]: While adding the proper types on our fetch function `getQuestBySlug` I have also tweaked the return statement to the following (see screen). If there is no returned data, it should actually be undefined -->

image

src/app/[lng]/superfest/[slug]/page.tsx
Comment (dennyscode) [reply src/app/[lng]/superfest/[slug]/page.tsx:47]: Yes, I have updated this everywhere where `getQuestBySlug` is being called. Check my answer above.

src/app/[lng]/zap/[slug]/page.tsx
Comment (dennyscode) [reply src/app/[lng]/zap/[slug]/page.tsx:68]: Are you sure? Please take a look on my updated `getQuestBySlug` function.

src/components/Quests/QuestPage/BackButton/BackButton.tsx
dennyscode: Oh! I copied that from a different file. I suppose we can also remove it over there. Will do!

src/components/Zap/BackButton/BackButton.tsx
dennyscode: Yes, good idea!

src/components/Zap/ZapAction/ZapAction.tsx
dennyscode: You have brought this up before, but I really do not know how to apply that. Maybe we can have a chat on what you would image the solution to look like: 

The <Tabs> functional-component currently passes those style-props accordingly as an sx to the actual styled-components. Although, I´d love to not pass on such a CSSObject, I don´t know how. 

Maybe you have an example how to solve?

src/components/Zap/ZapAction/Zapwidget.tsx
dennyscode: Much better! Thanks

src/components/Zap/BackButton/BackButton.tsx
dennyscode: Good question! I did actually try as it seems odd to add a third "BackButton" component. Unfortunately, the layout on the QuestMissionPage and SuperfestMissionPage does not allow this out of the box because of the way how the layout was being applied.
I will add some comments on all BackButton components to return and clean, once we have dedicated time for it.

@tcheee @oktapodia

src/hooks/useWidgetConfig.ts
dennyscode: Can you explain?
image

Should we:
- remove the whole try-catch block?
- remove only the "if-condition"? ..or
- replace the env-var?

@tcheee @oktapodia

src/components/Zap/ZapAction/ZapAction.tsx
dennyscode: @tcheee Discussing it in the next dedicated call.

src/hooks/useWidgetConfig.ts
dennyscode: Can you confirm and take another look if that is what you had in mind?

image

PR #1578, opened 2024-12-09
chore: adjust bg color on scan page

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-11330


src/components/Menus/MainMenu/useMainMenuContent.tsx
```diff
@@ -70,7 +69,7 @@ export const useMainMenuContent = () => {
 
   let mainMenu: any[] = [];
 
-  if (configTheme?.hasThemeModeSwitch) {
+  if (configTheme?.hasThemeModeSwitch && !disableThemeSwitch) {
```
tcheee, line 72: won't be enough as the dark theme does not exist so we have to force the light theme on this page

tcheee:
image
Issue on the menu if the user is coming from a dark theme

tcheee (changes requested):

dennyscode:
@tcheee I suggest to close this PR. I tried going an extra mile by trying to reduce complexity and avoiding partner-theme for our default themes such as "light" or "dark". But reality is, that our partner-themes are working well and that´s our current solution.

A minor adjustment within the "scan" partner-theme already did the trick and we may go on with that. Here is the current scan page, where we already achieved our goal to have the proper background-color:
image

PR #1582, opened 2024-12-12
feat: use SNS name

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-10979

### DIFF
--- a/src/hooks/useENSSNS.ts
+++ b/src/hooks/useENSSNS.ts
@@ -1,0 +1,52 @@
+import { useReverseLookup } from '@bonfida/sns-react';
+import { useConnection } from '@solana/wallet-adapter-react';
+import { PublicKey } from '@solana/web3.js';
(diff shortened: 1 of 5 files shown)


src/hooks/useENSSNS.ts
```diff
@@ -0,0 +1,64 @@
...
+  //   new PublicKey('4SthxbjwRJGW4uugoA111j6NKjBye1pZb16y4PHZ3TZC'),
+  // );
+
+  // SNS via useQuery:
+  const {
+    data: snsName,
+    isSuccess: isSuccessSns,
+    isError: isErrorSns,
+  } = useBonfidaSNS({
+    walletAddress: address,
```
dennyscode, line 40: By using one of the addresses above (e.g. - "google" or "alex") and replacing the address here, it can be tested locally without impersonator.

PR #1616, opened 2025-01-10
refactor: remove washtrade

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-11633


src/providers/WalletProvider/SVMProvider.tsx
```diff
@@ -30,7 +29,7 @@ export const SVMProvider: FC<PropsWithChildren> = ({ children }) => {
   return (
     <ConnectionProvider endpoint={endpoint}>
       <WalletProvider wallets={wallets} autoConnect>
-        <UmiContextApp>{children}</UmiContextApp>
```
dennyscode, line 33: Umi context is sitting inside the wash-folder, so I removed it as well. But since it´s part of the SVMProvider in here, I want to double check! Can we remove `UmiContextApp`?

tcheee (approved):
lgtm

src/providers/WalletProvider/SVMProvider.tsx
dennyscode: > Feel free to remove it

PR #1627, opened 2025-01-15
style: rename profile

tcheee:


dennyscode:
Or should we also rename the files accordingly to be consistent? Or is it over the top? Wdyt? I do not have a strong opinion on that matter btw.

PR #1629, opened 2025-01-15
feat: add traits section + page

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-11272


dennyscode:
Couple of things to figure out:

- [ ] let´s align on a specific tooltip, currently ChatGPT helped me to come up with a first shot
- [ ] on the Figma layout there is one example called "Transactor" which comes with a chain-logo, I did not implement this Avatar because I was not able to reproduce the API response

PR #1651, opened 2025-01-21
fix: revert get-tags changes and fix hydration errors

dennyscode:


src/app/lib/getTags.ts
```diff
@@ -44,8 +44,10 @@ const sortBlogArticlesByPublishedDate = (tags: TagAttributes[]) => {
 
           return dateB - dateA;
         });
+      return tag;
```
oktapodia, line 47: By the name of it, this is a sorting method, weird that it is not returning -1/0/1

src/components/Blog/BlogArticlesCollections/BlogArticlesTabs.tsx
```diff
@@ -39,6 +39,10 @@ export function BlogArticlesTabs({
   const [pageTab, setPageTab] = useState(pagination.page);
   const chunkedPages = chunkArray(data, pagination.pageSize);
 
+  if (!chunkedPages) {
```
oktapodia, line 42: if data is undefined, that means that the whole component should not be rendered.

src/utils/chunkArray.ts
```diff
@@ -1,4 +1,7 @@
 export function chunkArray<T>(array: T[], chunkSize: number): T[][] {
+  if (!array) {
```
oktapodia, line 2: that won't solve the issue, array might not be an array and not undefined

oktapodia (changes requested):
Couple of comments

src/app/lib/getTags.ts
dennyscode: Yes, it´s actually mapping + sorting

src/components/Blog/BlogArticlesCollections/BlogArticlesTabs.tsx
dennyscode: Thanks, to be safe - I added another check outside of this component within `<BlogArticlesCollections>` now

src/utils/chunkArray.ts
dennyscode: Checking if it´s typeof array now and also its length on top

src/components/Blog/BlogArticlesCollections/BlogArticlesTabs.tsx
oktapodia: ?

dennyscode: image

src/components/Blog/BlogArticlesCollections/BlogArticlesCollections.tsx
```diff
@@ -29,6 +29,11 @@ export const BlogArticlesCollections = ({
         ),
         total: tag.attributes?.blog_articles?.data.length,
       };
+
+      if (tag.attributes?.blog_articles.data.length === 0) {
```
oktapodia, line 33:
```suggestion
      if (!tag.attributes?.blog_articles?.data || tag.attributes?.blog_articles?.data.length === 0) {
```

oktapodia (approved):
LGTM

tcheee (approved):

PR #1674, opened 2025-01-27
refactor: use mui v6 grid

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-11995


src/components/Berachain/components/BerachainTransactionDetails/Recipe.tsx
dennyscode, line 1: <Recipe> being used within BerachainTransactionDetails.tsx but this one is not being called in any other file.
Can we remove it?

src/components/Blog/BlogArticlesCollections/BlogArticlesTabs.style.ts
```diff
@@ -1,9 +1,8 @@
 import type { Breakpoint } from '@mui/material';
-import { Grid } from '@mui/material';
 
 import { styled } from '@mui/material/styles';
 
-export const ArticlesGrid = styled(Grid)(({ theme }) => ({
+export const ArticlesGrid = styled('div')(({ theme }) => ({
```
oktapodia, line 5: Would be better to keep the Grid to avoid breaking something

oktapodia (changes requested):
just a comment

PR #1735, opened 2025-02-07
feat: use custom hook to sort by latest activity

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-12152

### DIFF
--- a/src/hooks/useAccountByLatestActivity.ts
+++ b/src/hooks/useAccountByLatestActivity.ts
@@ -1,0 +1,38 @@
+import { useAccount } from '@lifi/wallet-management';
+import { useMemo } from 'react';
+import { useChainTokenSelectionStore } from 'src/stores/chainTokenSelection';
+import { useChains } from './useChains';
+
+export const useAccountByLatestActivity = () => {
+  const { accounts, account } = useAccount();
+  const { sourceChainToken } = useChainTokenSelectionStore();
+  const { getChainById } = useChains();
+
+  const sortedAccounts = useMemo(() => {
+    if (!sourceChainToken || !sourceChainToken.chainId || !accounts.length) {
+      return accounts;
+    }
+
+    const sourceChain = getChainById(sourceChainToken.chainId);
+    if (!sourceChain) {
+      return accounts;
+    }
+
+    const sourceChainType = sourceChain.chainType;
+
+    return [...accounts].sort((a, b) => {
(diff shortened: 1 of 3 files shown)


src/hooks/useAccountByLatestActivity.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    const sourceChain = getChainById(sourceChainToken.chainId);
+    if (!sourceChain) {
+      return accounts;
+    }
+
+    const sourceChainType = sourceChain.chainType;
+
+    return [...accounts].sort((a, b) => {
```
oktapodia, line 23: `accounts` should already be an array

src/hooks/useAccountByLatestActivity.ts
```diff
@@ -0,0 +1,38 @@
...
+import { useChainTokenSelectionStore } from 'src/stores/chainTokenSelection';
+import { useChains } from './useChains';
+
+export const useAccountByLatestActivity = () => {
+  const { accounts, account } = useAccount();
+  const { sourceChainToken } = useChainTokenSelectionStore();
+  const { getChainById } = useChains();
+
+  const sortedAccounts = useMemo(() => {
+    if (!sourceChainToken || !sourceChainToken.chainId || !accounts.length) {
```
oktapodia, line 12: `!sourceChainToken?.chainId` checks the same

oktapodia, line 12: !accounts.length might crash the app, check if it is an array or length > 0 (depending on what you would like to achieve)

oktapodia (dismissed):
IMO it should be part of the wallet-management lib as it is deeply changing its behavior

src/hooks/useAccountByLatestActivity.ts
dennyscode: Righhhht! Thanks

src/hooks/useAccountByLatestActivity.ts
dennyscode: Right, cleaner that way!

dennyscode: Yes, refactored that part. Thanks!

chenchen1398:
> Ticket: https://lifi.atlassian.net/browse/LF-12152

Thanks⬇️

PR #1772, opened 2025-02-18
feat: Added gcp sm in github actions staging

adityalifi:
Integrated gcp secrets manager with github actions in staging.

### DIFF
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -1,4 +1,4 @@
-name: Build docker
+name: Build Docker
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-docker.yaml
```diff
@@ -9,61 +9,103 @@ on:
...
+        gcloud auth login --brief --cred-file="${{ steps.gcpauth.outputs.credentials_file_path }}"
+        gcloud auth configure-docker europe-docker.pkg.dev
+
+    - name: Convert JSON .env to key-value format
+      run: |
+        if [ -f ".env.json" ]; then
+          echo "✅ .env.json file found! Converting to .env format..."
+          
+          # Convert JSON to key=value format
+          jq -r 'to_entries | map("\(.key)=\(.value | tostring)") | .[]' .env.json > .env
```
kalote, line 79: should it be `.env.development`, `.env.production`, ... instead of `.env` ? You can check how it is use in the dockerfile or app

kalote (approved):

PR #1777, opened 2025-02-18
refactor: useWidgetConfig hook usage

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-12395

### DIFF
--- a/src/hooks/useWidgetConfig.ts
+++ b/src/hooks/useWidgetConfig.ts
@@ -1,10 +1,11 @@
 import { EVM } from '@lifi/sdk';
-import { useWalletMenu } from '@lifi/wallet-management';
-import type { WidgetTheme } from '@lifi/widget';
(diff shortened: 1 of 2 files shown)


src/hooks/useWidgetConfig.ts
```diff
@@ -129,16 +168,9 @@ export const useWidgetConfig = ({
...
-      widgetTheme.config.theme,
-      customWidgetTheme,
-    );
-
     return {
       ...formParameters,
-      variant:
-        configTheme.variant ??
-        (starterVariant === 'refuel' ? 'compact' : 'wide'),
+      variant: configTheme.variant ?? 'wide',
```
dennyscode, line 173: @tcheee @oktapodia If we agree to add this change, I can go ahead and apply the partner-themes + check widgets in all places.

PR #1785, opened 2025-02-19
fix: error handling spindl + callRequest

dennyscode:


src/hooks/feature-cards/spindl/useSpindlCards.ts
```diff
@@ -30,9 +30,8 @@ export const useSpindlCards = () => {
 
   const fetchSpindlData = useCallback(
     async ({ country, chainId, tokenAddress, address }: SpindlFetchParams) => {
-      if (!showSpindle) {
-        // console.log('User is not part of the Spindl A/B test group.');
-        return; // Exit early if the user is not in the A/B test group
+      if (!showSpindle || !(spindlConfig?.apiUrl && spindlConfig?.headers)) {
```
oktapodia, line 33: As discussed when implementing this feature, config is checked into the getSpindlConfig itself and throw an error if not, so that extra part of the if statement will never be reachable

src/hooks/feature-cards/spindl/trackSpindl.ts
```diff
@@ -25,12 +25,15 @@ export async function trackSpindl(
   };
 
   try {
+    if (!(spindlConfig?.apiUrl && spindlConfig?.headers)) {
```
oktapodia, line 28: As discussed when implementing this feature, config is checked into the getSpindlConfig itself and throw an error if not, so that extra part of the if statement will never be reachable

src/hooks/feature-cards/spindl/spindlConfig.ts
```diff
@@ -3,9 +3,10 @@ export const getSpindlConfig = () => {
     !process.env.NEXT_PUBLIC_SPINDL_API_KEY ||
     !process.env.NEXT_PUBLIC_SPINDL_API_URL
   ) {
-    throw new Error(
+    console.error(
```
oktapodia, line 6: Why is that change?

oktapodia (changes requested):
Couple of comments

src/hooks/feature-cards/spindl/spindlConfig.ts
dennyscode: @oktapodia @tcheee 

We actually ran into an issue while the necessary variables for Spindl were not accessible at that time. We wanted to prevent the crash of a page if such problems occur again in the future. That´s why we chose to log a "silent-error" instead.

src/hooks/feature-cards/spindl/useSpindlCards.ts
dennyscode: That´s 100% right! We had it in our previous version as we discussed but due to the error described in my next comment, we adjusted the spindlConfig to return `null` if the necessary env-variables are not provided. This is the extra check I added here.

_Goal: Avoid to throw an error which leads to crash the page._
Open for any improvements and would love to hear or jump on a call!

src/hooks/feature-cards/spindl/trackSpindl.ts
dennyscode: See comment above

PR #1797, opened 2025-02-26
fix: glitch color on first load

oktapodia:


src/utils/formatTheme.ts
```diff
@@ -155,4 +155,4 @@ export function formatTheme(theme: PartnerThemesAttributes) {
 }
 
 export const isDarkOrLightThemeMode = (theme: PartnerThemesAttributes) =>
-  theme.darkConfig ? 'dark' : 'light';
+  theme.lightConfig ? 'light' : 'dark';
```
chybisov, line 158: Is this change just favoring dark mode over light mode, or there is a real glitch? 👀

tcheee: there is a small glitch on some config

tcheee (approved):

PR #1819, opened 2025-03-05
style: adjust rank fontSize on profile

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-12652


admsk2 (approved):

src/theme/theme.ts
```diff
@@ -123,6 +123,7 @@ declare module '@mui/material/styles' {
     titleSmall: React.CSSProperties;
     titleXSmall: React.CSSProperties;
     title2XSmall: React.CSSProperties;
+    titleMedium: React.CSSProperties;
```
oktapodia, line 126: Not used variable

src/components/ProfilePage/LeaderboardCard/LeaderboardCard.style.ts
```diff
@@ -60,23 +60,32 @@ export const CardButton = styled(ButtonSecondary)(({ theme }) => ({
...
+})<LeaderboardUserPositionButtonProps>(({ theme, isGtMillion }) => ({
+  padding: theme.spacing(0, 1),
+  textDecoration: 'none',
+  position: 'relative',
+  marginTop: theme.spacing(1),
+  height: 64,
+  background: 'transparent',
+  borderRadius: '16px',
+
+  [theme.breakpoints.up('xs')]: {
```
oktapodia, line 78: That will only solve the issue on mobile

src/components/ProfilePage/LeaderboardCard/LeaderboardCard.tsx
```diff
@@ -33,21 +33,20 @@ export const LeaderboardCard = ({ address }: { address?: string }) => {
...
-                ? t('format.decimal2Digit', { value: position })
-                : 'N/A'}
-            </LeaderboardUserTitle>
-          </LeaderboardUserPositionButton>
-        ) : (
-          <LeaderboardUserTitle variant="titleLarge">N/A</LeaderboardUserTitle>
-        )}
+        <LeaderboardUserPositionButton
+          isGtMillion={isGtMillion}
+          disabled={!position}
```
oktapodia, line 38: Not sure that those changes are related to the story + will improve the view

src/components/ProfilePage/LeaderboardCard/LeaderboardCard.tsx
```diff
@@ -9,7 +9,6 @@ import {
   CardButton,
   CardButtonContainer,
   LeaderboardUserPositionButton,
-  LeaderboardUserTitle,
```
oktapodia, line 12: Not used anymore, can be removed

oktapodia (changes requested):
couple of comments

FYI [fitty](https://github.com/rikschennink/fitty) is good for width text auto sizing

src/components/ProfilePage/LeaderboardCard/LeaderboardCard.style.ts
dennyscode: Not sure about that:
image

src/components/ProfilePage/LeaderboardCard/LeaderboardCard.tsx
dennyscode: Great find, thanks

src/components/ProfilePage/LeaderboardCard/LeaderboardCard.tsx
dennyscode: It should simplify the components by using:
- one single button
- if no position (because likely being disconnected), it´s just disabled
- if no position, just the label of the button changes instead of using a seperate component

src/theme/theme.ts
dennyscode: I have removed it now, thanks. Along the way I tried that typography and did not end up using it.

But personally, I would prefer adding all figma variables (colors + typography) to the theme although some may be unused at that time. That way a developer does not constantly have to touch the theme and could just use it.
But I agree, adding a single typography definition does not help so much here! 😃

dennyscode:
> couple of comments
> 
> FYI [fitty](https://github.com/rikschennink/fitty) is good for width text auto sizing

Good to know fitty!! Thanks for sharing!
I am not sure we´ll need it here because those minor changes did the trick for me. But let me know what you think

src/components/ProfilePage/LeaderboardCard/LeaderboardCard.style.ts
oktapodia: Sorry my comment wasn't explainatory enough, `theme.breakpoints.up('xs')` is taking in consideration every screen size above xs, not sure it is gonna do the expected behavior.

+ you are only doing something within this breakpoint IF `isGtMillion` is defined so the default object is declared for no reasons.

src/components/ProfilePage/LeaderboardCard/LeaderboardCard.tsx
oktapodia: But then that means that the hover is still there on the leaderboard position but no click possible, make the UX weird
CleanShot 2025-04-14 at 09 48 31@2x

dennyscode: But no pointer event still.. but I get it, okay

src/components/ProfilePage/LeaderboardCard/LeaderboardCard.style.ts
dennyscode: To simplify the fontSize will now be adjusted throughout all breakpoints

oktapodia (approved):
LGTM

PR #1823, opened 2025-03-07
fix: use hook to retrieve chainTokenSelection

dennyscode:


src/components/Widgets/Widget.tsx
```diff
@@ -272,7 +271,7 @@ export function Widget({
       ],
       requiredUI:
         // if AGW connected and destinationChainToken is ABS, require toAddress
-        !isConnectedAGW && destinationChainToken.chainId === ChainId.ABS
+        !isConnectedAGW && destinationChainToken.chainId === String(ChainId.ABS)
```
chybisov, line 274: It might be better to parse chain id to number when getting it from the query params

PR #1826, opened 2025-03-07
chore: fix subvariant split on agw via formref

dennyscode:


src/hooks/useWidgetFormObserver.tsx
```diff
@@ -0,0 +1,160 @@
...
+    });
+    observer.observe(wrapperRef.current, {
+      childList: true,
+      subtree: true,
+    });
+    return () => observer.disconnect();
+  }, [allowToChains, configTheme?.chains?.to?.allow, formRef, wrapperRef]);
+
+  useEffect(() => {
+    const observer = new MutationObserver(() => {
```
chybisov, line 98: I'm not sure what this is for, but we don't have a disconnect in effect clean-up function

PR #1842, opened 2025-03-18
chore: remove abtest for split subvariant

dennyscode:


oktapodia:
As @admsk2 suggested, can't we disable it only? Seems to touch a lot of files which will make it harder to re-enable

src/hooks/useABTest.ts
```diff
@@ -1,60 +1,46 @@
...
+    queryKey: ['abtest', feature, address],
     queryFn: async () => {
-      if (!feature || !user) {
-        return { isEnabled: false, isSuccess: true, isLoading: false };
-      }
-
-      const res = await fetch(
-        `${apiBaseUrl}/posthog/feature-flag?key=${feature}&distinctId=${user}`,
+      const response = await fetch(
+        `${apiBaseUrl}/posthog/feature-flag?key=${feature}&distinctId=${address}`,
```
oktapodia, line 24: caching is not by default since nextjs15, would be good to add 5/15mins cache (to avoid spamming backend)

dennyscode: Good hint, thanks!

oktapodia (approved):
LGTM

PR #1926, opened 2025-05-12
feat: contribution

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-13083

### DIFF
--- a/src/components/Widgets/FeeContribution/FeeContribution.style.ts
+++ b/src/components/Widgets/FeeContribution/FeeContribution.style.ts
@@ -1,0 +1,21 @@
+import { alpha, Button, ButtonProps, darken, styled } from '@mui/material';
+
+interface ContributionButtonProps extends ButtonProps {
+  active: boolean;
+}
+
+export const ContributionButton = styled(Button, {
+  shouldForwardProp: (prop) => prop !== 'active',
+})<ContributionButtonProps>(({ theme, active }) => ({
+  width: '100%',
+  fontSize: '12px',
+  lineHeight: '16px',
+  fontWeight: 700,
+  height: '32px',
+  color: alpha(theme.palette.text.primary, 0.84),
+  transition: 'background-color 250ms',
+  backgroundColor: active ? '#F0E5FF' : theme.palette.grey[100],
+  '&:hover': {
+    backgroundColor: active ? darken('#F0E5FF', 0.04) : theme.palette.grey[300],
+  },
+}));
--- a/src/components/Widgets/FeeContribution/FeeContribution.tsx
+++ b/src/components/Widgets/FeeContribution/FeeContribution.tsx
@@ -1,0 +1,480 @@
+import { ChainType } from '@lifi/sdk';
+import { useAccount } from '@lifi/wallet-management';
+import CheckIcon from '@mui/icons-material/Check';
+import {
+  Box,
+  Card,
+  CircularProgress,
+  darken,
+  Drawer,
+  Grid,
+  Input,
+  Typography,
+} from '@mui/material';
+import { useEffect, useMemo, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { TrackingAction, TrackingCategory } from 'src/const/trackingKeys';
+import { useGetTokenBalance } from 'src/hooks/useGetTokenBalance';
+import { useUserTracking } from 'src/hooks/userTracking/useUserTracking';
+import { Transfer, useTxHistory } from 'src/hooks/useTxHistory';
+import { useRouteStore } from 'src/stores/route/RouteStore';
+import { formatInputAmount } from 'src/utils/format';
+import { parseUnits } from 'viem';
+import { useWaitForTransactionReceipt, useWriteContract } from 'wagmi';
+import { ContributionButton } from './FeeContribution.style';
+
+const checkContributionByTxHistory = (
+  transfers: Transfer[] | undefined,
+): boolean => {
+  const transactionCount = transfers?.length ?? 0;
+  return (
+    transactionCount === 0 ||
+    (transactionCount > 0 && transactionCount % 3 === 0)
+  );
+};
+
+const DEFAULT_CONTRIBUTION_AMOUNTS = [0.5, 1, 2];
+
+const getContributionAmounts = (volume: number) => {
+  if (volume >= 100000) {
+    return [5, 10, 15];
+  } else if (volume >= 10000) {
+    return [2, 10, 15];
+  } else if (volume >= 1000) {
+    return [1, 5, 10];
+  } else {
+    return DEFAULT_CONTRIBUTION_AMOUNTS;
+  }
+};
+
+const FeeContribution = () => {
+  const { account } = useAccount();
+  const { trackEvent } = useUserTracking();
+  const [address, setAddress] = useState<string | undefined>(undefined);
+  const [contributionAmounts, setContributionAmounts] = useState<number[]>(
+    DEFAULT_CONTRIBUTION_AMOUNTS,
+  );
+  // Show Contribution for ~10% of users
+  const isContributionAbEnabled = useMemo(() => {
+    return Math.random() < 0.1;
+  }, []);
+  const [amount, setAmount] = useState<string | undefined>();
+  const [inputAmount, setInputAmount] = useState<string | undefined>(undefined);
+  const [isSending, setIsSending] = useState(false);
+  const completedRoute = useRouteStore((state) => state.completedRoute);
+  const { data } = useTxHistory(address, completedRoute?.id);
+  const { data: tokenBalanceData } = useGetTokenBalance(
+    address,
+    completedRoute?.toToken,
+  );
+  const [showContribution, setShowContribution] = useState(false);
+  const containerRef = useRef<HTMLDivElement>(null);
+  const {
+    status: txStatus,
+    data: writeTxHash,
+    isPending: isTxPending,
+    isError: isTxError,
+    error: txError,
+    writeContract,
+  } = useWriteContract();
+
+  const {
+    data: txReceipt,
+    isLoading: isTxConfirming,
+    isSuccess: isTxConfirmed,
+    isError: isTxConfirmError,
+  } = useWaitForTransactionReceipt({
+    chainId: completedRoute?.toChainId,
+    hash: writeTxHash,
+    confirmations: 1,
+    pollingInterval: 1_000,
+  });
+
+  const hasTrackedImpressionRef = useRef(false);
+  const hasTrackedConfirmationRef = useRef(false);
+  const { t } = useTranslation();
+
+  function onChangeValue(event: React.ChangeEvent<HTMLInputElement>) {
+    const { value } = event.target;
+    // Remove dollar sign if present and format the amount
+    const valueWithoutDollar = value.replace('$', '');
+
+    // Calculate max amount in USD based on token balance
+    const maxTokenAmount = tokenBalanceData
+      ? Number(tokenBalanceData.amount) /
+        Math.pow(10, tokenBalanceData.decimals)
+      : 0;
+    const maxUsdAmount =
+      maxTokenAmount * Number(completedRoute?.toToken?.priceUSD || 0);
+    const formattedAmount = formatInputAmount(
+      valueWithoutDollar,
+      2,
+      false,
+      maxUsdAmount,
+    );
+    setInputAmount(formattedAmount);
+    setAmount(formattedAmount);
+  }
+
+  // Set the address of the user from last tx
+  useEffect(() => {
+    setAddress(completedRoute?.fromAddress);
+  }, [completedRoute]);
+
+  // Check if contribution is enabled
+  useEffect(() => {
+    if (data?.transfers) {
+      const isContributionEnabledByTxHistory = checkContributionByTxHistory(
+        data.transfers,
+      );
+      const txUsdAmount = Number(completedRoute?.toAmountUSD || 0);
+      if (
+        txUsdAmount >= 10 &&
+        isContributionAbEnabled &&
+        isContributionEnabledByTxHistory &&
+        account?.chainType === ChainType.EVM
+      ) {
+        // if contribution is enabled:
+        // - based on chain type === EVM
+        // - based on past tx´s
+        // - and based on ab test
+        // - and based on the contribution amount > 10 USD
+        setShowContribution(true);
+        setContributionAmounts(getContributionAmounts(txUsdAmount));
+      }
+    }
+  }, [data]);
+
+  // Handle contribution button click
+  const handleButtonClick = (selectedAmount: number) => {
+    if (!!amount && selectedAmount === parseFloat(amount)) {
+      setAmount(undefined);
+    } else {
+      setAmount(selectedAmount.toString());
+    }
+  };
+
+  // Handle custom contribution amount
+  const handleCustom = () => {
+    if (!!inputAmount && parseFloat(inputAmount) > 0) {
+      setAmount(inputAmount);
+    }
+  };
+
+  // Track contribution impression
+  useEffect(() => {
+    if (showContribution && !hasTrackedImpressionRef.current) {
+      trackEvent({
+        action: TrackingAction.ContributeImpression,
+        category: TrackingCategory.Widget,
+        label: 'fee_contribution_impression',
+        data: {
+          donator: account?.address || '',
+          original_tx_id: completedRoute?.id || '',
+          original_amount: completedRoute?.toAmount || '',
+          original_amount_usd: completedRoute?.toAmountUSD || '',
+          original_token_symbol: completedRoute?.toToken?.symbol || '',
+          original_donation_chain: completedRoute?.toChainId || 0,
+          original_tx_token_address: completedRoute?.toToken?.address || '',
+        },
+      });
+      hasTrackedImpressionRef.current = true;
+    }
+  }, [showContribution]); // Only depend on showContribution
+
+  // Track contribution success
+  useEffect(() => {
+    if (isTxConfirmed && !hasTrackedConfirmationRef.current) {
+      trackEvent({
+        action: TrackingAction.ContributeSuccess,
+        category: TrackingCategory.Widget,
+        label: 'fee_contribution_success',
+        data: {
+          donator: account?.address || '',
+          original_tx_id: completedRoute?.id || '',
+          donation_amount_usd: amount || 0,
+          donation_token_symbol: completedRoute?.toToken?.symbol || '',
+          donation_token_address: completedRoute?.toToken?.address || '',
+          donation_chain: completedRoute?.toChainId || 0,
+          donation_tx_hash: txReceipt?.transactionHash || '',
+        },
+      });
+      hasTrackedConfirmationRef.current = true;
+    }
+  }, [isTxConfirmed]); // Only depend on isTxConfirmed
+
+  // Handle contribution confirmation to sign the tx
+  const handleConfirm = async () => {
+    if (
+      isTxConfirmed ||
+      !amount ||
+      !completedRoute?.toToken ||
+      !account?.address
+    )
+      return;
+
+    setIsSending(true);
+    try {
+      // amount is in USD, convert to token amount
+      const tokenPriceUSD = Number(completedRoute.toToken.priceUSD);
+      if (!tokenPriceUSD || tokenPriceUSD <= 0)
+        throw new Error('Invalid token price');
+
+      const usdAmount = parseFloat(amount);
+      // Use token's actual decimals for precision
+      const tokenAmount = (usdAmount / tokenPriceUSD).toFixed(
+        completedRoute.toToken.decimals,
+      );
+
+      // Ensure we have a non-zero amount after conversion
+      if (parseFloat(tokenAmount) === 0) {
+        throw new Error(t('contribution.error.amountTooSmall'));
+      }
+
+      // Track click event
+      trackEvent({
+        action: TrackingAction.ClickContribute,
+        category: TrackingCategory.Widget,
+        label: 'click_fee_contribution',
+        data: {
+          donator: account?.address || '',
+          original_tx_id: completedRoute?.id || '',
+          donation_amount_usd: usdAmount || 0,
+          donation_amount_token: tokenAmount || 0,
+          donation_token_symbol: completedRoute?.toToken?.symbol || '',
+          donation_token_address: completedRoute?.toToken?.address || '',
+          donation_chain: completedRoute?.toChainId || 0,
+        },
+      });
+
+      // Convert to token units with proper precision
+      const amountInTokenUnits = parseUnits(
+        tokenAmount,
+        completedRoute.toToken.decimals,
+      );
+
+      await writeContract({
+        address: completedRoute.toToken.address as `0x${string}`, // todo: change to actual wallet address !
+        abi: [
+          {
+            name: 'transfer',
+            type: 'function',
+            stateMutability: 'nonpayable',
+            inputs: [
+              { name: 'to', type: 'address' },
+              { name: 'amount', type: 'uint256' },
+            ],
+            outputs: [{ name: '', type: 'bool' }],
+          },
+        ],
+        functionName: 'transfer',
+        args: [account.address as `0x${string}`, amountInTokenUnits],
+        chainId: completedRoute.toChainId,
+      });
+    } catch (error) {
+      console.error('Error sending contribution:', error);
+      // Add user-friendly error message
+      if (
+        error instanceof Error &&
+        error.message === 'Amount too small for this token'
+      ) {
+        alert(
+          'The contribution amount is too small for this token. Please try a larger amount.',
+        );
+      }
+    } finally {
+      setIsSending(false);
+    }
+  };
+
+  const isCustomActive = useMemo(() => {
+    return (
+      !!amount &&
+      !contributionAmounts.includes(parseFloat(amount)) &&
+      parseFloat(amount) > 0
+    );
+  }, [amount]);
+
+  if (!showContribution) {
+    return null;
+  }
+
+  return (
+    <Box
+      sx={{
+        position: 'relative',
+        transition: 'height 250ms',
+        height: showContribution ? '156px' : '0px',
+      }}
+    >
+      <Box
+        ref={containerRef}
+        sx={{
+          position: 'absolute',
+          bottom: 0,
+          left: 0,
+          right: 0,
+          top: 0,
+          overflow: 'hidden',
+          borderRadius: '24px',
+        }}
+      >
+        <Drawer
+          open={showContribution}
+          anchor="top"
+          hideBackdrop={true}
+          container={() => containerRef.current}
+          sx={{ position: 'absolute' }}
+          ModalProps={{
+            disablePortal: true,
+            disableEnforceFocus: true,
+            disableAutoFocus: true,
+          }}
+          slotProps={{
+            paper: {
+              sx: {
+                position: 'absolute',
+                bottom: 0,
+                left: 0,
+                right: 0,
+                width: '100%',
+                borderRadius: '24px',
+              },
+            },
+            backdrop: {
+              sx: {
+                position: 'absolute',
+                backgroundColor: 'rgba(0, 0, 0, 0.5)', // Adjust opacity as needed
+                borderRadius: '24px',
+              },
+            },
+          }}
+        >
+          <Card
+            sx={(theme) => ({
+              display: 'flex',
+              flexDirection: 'column',
+              gap: theme.spacing(2),
+              border: 'unset',
+              padding: theme.spacing(2),
+              borderRadius: '16px',
+              boxShadow: '0px 2px 8px 0px rgba(0, 0, 0, 0.04)',
+            })}
+          >
+            <Typography
+              sx={{ fontSize: '14px', lineHeight: '20px', fontWeight: 700 }}
+            >
+              {t('contribution.title')}
+            </Typography>
+            <Grid
+              container
+              spacing={2}
+              columnSpacing={1}
+              justifyContent={'space-between'}
+            >
+              {contributionAmounts.map((contributionAmount) => (
+                <Grid size={3} key={contributionAmount}>
+                  <ContributionButton
+                    active={
+                      !!amount && parseFloat(amount) === contributionAmount
+                    }
+                    onClick={() => handleButtonClick(contributionAmount)}
+                    size="small"
+                  >
+                    ${contributionAmount}
+                  </ContributionButton>
+                </Grid>
+              ))}
+              <Grid size={3}>
+                <Input
+                  value={inputAmount ? `$${inputAmount}` : ''}
+                  aria-autocomplete="none"
+                  onChange={onChangeValue}
+                  onClick={handleCustom}
+                  disableUnderline
+                  placeholder={t('contribution.custom')}
+                  sx={(theme) => ({
+                    height: '32px',
+                    '& .MuiInputBase-input::placeholder': {
+                      color: theme.palette.text.primary,
+                      opacity: 1,
+                      textAlign: 'center',
+                    },
+                  })}
+                  slotProps={{
+                    input: {
+                      sx: (theme) => ({
+                        height: '32px',
+                        border: 'none',
+                        borderRadius: '16px',
+                        padding: 0,
+                        fontSize: '12px',
+                        lineHeight: '16px',
+                        fontWeight: 700,
+                        textAlign: 'center',
+                        backgroundColor: isCustomActive
+                          ? '#F0E5FF'
+                          : theme.palette.grey[100],
+                        color: theme.palette.text.primary,
+                        transition: 'background-color 250ms',
+                        '&:hover': {
+                          backgroundColor: isCustomActive
+                            ? darken('#F0E5FF', 0.04)
+                            : theme.palette.grey[300],
+                        },
+                      }),
+                    },
+                  }}
+                />
+              </Grid>
+            </Grid>
+
+            {!!amount ? (
+              <ContributionButton
+                active={false}
+                onClick={handleConfirm}
+                disabled={isSending || isTxPending || isTxConfirming}
+                sx={() => ({
+                  height: '40px',
+                  fontSize: '14px',
+                  lineHeight: '16px',
+                  fontWeight: 700,
+                  gap: '8px',
+                  ...(isTxConfirmed && {
+                    backgroundColor: '#D6FFE7',
+                    color: '#00B849',
+                    '&:hover': {
+                      backgroundColor: darken('#D6FFE7', 0.04),
+                    },
+                  }),
+                })}
+              >
+                {isTxConfirmed ? <CheckIcon /> : null}
+                {isSending || isTxPending || isTxConfirming ? (
+                  <CircularProgress size={20} color="inherit" />
+                ) : isTxConfirmed ? (
+                  t('contribution.contributionSent')
+                ) : (
+                  t('contribution.confirm')
+                )}
+              </ContributionButton>
+            ) : (
+              <Typography
+                sx={(theme) => ({
+                  fontSize: '12px',
+                  lineHeight: '16px',
+                  fontWeight: 500,
+                  color: theme.palette.grey[700],
+                })}
+              >
+                {t('contribution.description')}
+              </Typography>
+            )}
+          </Card>
+        </Drawer>
+      </Box>
+    </Box>
+  );
+};
+
+export default FeeContribution;
--- a/src/components/Widgets/Widget.tsx
+++ b/src/components/Widgets/Widget.tsx
@@ -29,6 +29,7 @@ import { useWelcomeScreen } from 'src/hooks/useWelcomeScreen';
 import { getWidgetThemeV2 } from 'src/providers/ThemeProvider/utils';
 import { useActiveTabStore } from 'src/stores/activeTab';
 import { themeAllowChains, WidgetWrapper } from '.';
+import FeeContribution from './FeeContribution/FeeContribution';
 import type { WidgetProps } from './Widget.types';
 
 export function Widget({
@@ -45,16 +46,14 @@ export function Widget({
   autoHeight,
 }: WidgetProps) {
   const theme = useTheme();
-  const [configTheme] = useThemeStore((state) => [
-    state.configTheme,
-  ]);
+  const [configTheme] = useThemeStore((state) => [state.configTheme]);
 
-    const { mode } = useColorScheme();
-    const prefersDarkMode = useMediaQuery('(prefers-color-scheme: dark)');
-
+  const { mode } = useColorScheme();
+  const prefersDarkMode = useMediaQuery('(prefers-color-scheme: dark)');
 
   const widgetTheme = getWidgetThemeV2(
-    mode === 'system' || !mode ? prefersDarkMode ? 'dark' : 'light' : mode);
+    mode === 'system' || !mode ? (prefersDarkMode ? 'dark' : 'light') : mode,
+  );
   const { destinationChainToken, toAddress } = useUrlParams();
   const widgetEvents = useWidgetEvents();
   const formRef = useRef<FormState>(null);
@@ -336,6 +335,9 @@ export function Widget({
           integrator={config.integrator}
           config={config}
           formRef={formRef}
+          feeConfig={{
+            _vcComponent: () => <FeeContribution />,
+          }}
         />
       </ClientOnly>
     </WidgetWrapper>
--- a/src/hooks/useTxHistory.ts
+++ b/src/hooks/useTxHistory.ts
@@ -1,0 +1,122 @@
+import { useQuery } from '@tanstack/react-query';
+
+interface Token {
+  address: string;
+  chainId: number;
+  symbol: string;
+  decimals: number;
+  name: string;
+  coinKey: string;
+  logoURI: string;
+  priceUSD: string;
+}
+
+interface ToolDetails {
+  key: string;
+  name: string;
+  logoURI: string;
+  webUrl: string;
+}
+
+interface IncludedStep {
+  tool: string;
+  toolDetails: ToolDetails;
+  fromAmount: string;
+  fromToken: Token;
+  toToken: Token;
+  toAmount: string;
+  bridgedAmount: string | null;
+}
+
+interface TransactionDetails {
+  txHash: string;
+  txLink: string;
+  token: Token;
+  chainId: number;
+  gasPrice: string;
+  gasUsed: string;
+  gasToken: Token;
+  gasAmount: string;
+  gasAmountUSD: string;
+  amountUSD: string;
+  value: string;
+  includedSteps?: IncludedStep[];
+  amount: string;
+  timestamp: number;
+}
+
+export interface Transfer {
+  transactionId: string;
+  sending: TransactionDetails;
+  receiving: TransactionDetails;
+  lifiExplorerLink: string;
+  fromAddress: string;
+  toAddress: string;
+  tool: string;
+  status: string;
+  substatus: string;
+  substatusMessage: string;
+  metadata: {
+    integrator: string;
+  };
+  feeCosts: any[];
+}
+
+interface TransferResponse {
+  transfers: Transfer[];
+}
+
+export interface TxHistoryProps {
+  data: TransferResponse | null;
+  total: number;
+  isLoading: boolean;
+  isError: boolean;
+  error: unknown;
+}
+
+export async function getTxHistoryQuery(walletAddress?: string) {
+  if (!walletAddress) {
+    return;
+  }
+
+  const toTimestamp = Math.floor(Date.now() / 1000); // Current timestamp in seconds
+  const fromTimestamp = toTimestamp - 30 * 24 * 60 * 60; // 30 days ago in seconds
+
+  const integrator = 'dev.jumper.exchange';
+
+  const response = await fetch(
+    `${process.env.NEXT_PUBLIC_LIFI_API_URL}/analytics/transfers?integrator=${integrator}&wallet=${walletAddress}&fromTimestamp=${fromTimestamp}&toTimestamp=${toTimestamp}`,
+  );
+
+  if (!response.ok) {
+    throw new Error('Failed to fetch transfers');
+  }
+
+  return response.json();
+}
+
+export const useTxHistory = (
+  walletAddress?: string,
+  completedRouteId?: string,
+): TxHistoryProps => {
+  const {
+    data: transfers,
+    isLoading,
+    isError,
+    error,
+  } = useQuery({
+    queryKey: ['transfers', walletAddress, completedRouteId],
+    queryFn: () => getTxHistoryQuery(walletAddress),
+    enabled: !!walletAddress,
+    staleTime: 5 * 60 * 1000, // Consider data fresh for 5 minutes
+    gcTime: 30 * 60 * 1000, // Keep data in cache for 30 minutes
+  });
+
+  return {
+    data: transfers || null,
+    total: transfers?.transfers.length || 0,
+    isLoading,
+    isError,
+    error,
+  };
+};
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -27,3 +27,93 @@ export function formatTokenPrice(amount?: string, price?: string) {
   }
   return parseFloat(amount) * parseFloat(price);
 }
+
+export function formatInputAmount(
+  amount: string,
+  decimals: number | null = null,
+  returnInitial = false,
+  maxValue?: number,
+) {
+  if (!amount) {
+    return amount;
+  }
+
+  // Remove any characters that are not numbers, dots, or commas
+  let formattedAmount = amount.replace(/[^0-9.,]/g, '').trim();
+
+  // Replace all commas with dots for consistent decimal handling
+  formattedAmount = formattedAmount.replaceAll(',', '.');
+
+  // Handle duplicate decimal points
+  if (
+    formattedAmount.endsWith('.') &&
+    formattedAmount.slice(0, -1).includes('.')
+  ) {
+    const hasPreDecimalFigure = formattedAmount.at(-3);
+    formattedAmount = hasPreDecimalFigure ? formattedAmount.slice(0, -2) : '0';
+  }
+
+  // Allow removal of leading "0." and empty input
+  if (formattedAmount === '' || formattedAmount === '.') {
+    return formattedAmount;
+  }
+
+  // Special handling for inputs starting with "0." or just "0"
+  if (formattedAmount === '0' || formattedAmount.startsWith('0.')) {
+    if (returnInitial) {
+      return formattedAmount;
+    }
+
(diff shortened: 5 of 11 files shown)


src/components/Widgets/FeeContribution/FeeContribution.style.ts
```diff
@@ -0,0 +1,138 @@
...
+}));
+
+interface ContributionCustomInputProps extends InputProps {
+  active: boolean;
+}
+
+export const ContributionCustomInput = styled(Input, {
+  shouldForwardProp: (prop) => prop !== 'active',
+})<ContributionCustomInputProps>(({ theme, active }) => ({
+  height: '32px',
```
oktapodia, line 60: active is unused

src/components/Widgets/FeeContribution/FeeContribution.style.ts
```diff
@@ -0,0 +1,138 @@
...
+interface ContributionButtonProps extends ButtonProps {
+  active: boolean;
+  mode: 'light' | 'dark' | 'system' | undefined;
+}
+
+export const ContributionButton = styled(Button, {
+  shouldForwardProp: (prop) => prop !== 'active',
+})<ContributionButtonProps>(({ theme, active, mode }) => {
+  console.log('active', active);
+  console.log('theme.palette.mode', theme.palette.mode);
```
oktapodia, line 78: console.logs

src/components/Widgets/FeeContribution/FeeContribution.style.ts
```diff
@@ -0,0 +1,138 @@
...
+  return {
+    width: '100%',
+    fontSize: '12px',
+    lineHeight: '16px',
+    fontWeight: 700,
+    height: '32px',
+
+    transition: 'background-color 250ms',
+    color: (theme.vars || theme).palette.text.primary,
+    backgroundColor: active
```
oktapodia, line 88: can you use a variant instead please

src/components/Widgets/FeeContribution/FeeContribution.style.ts
```diff
@@ -0,0 +1,138 @@
...
+      ? 'rgba(101, 59, 163, 0.84) !important'
+      : (theme.vars || theme).palette.grey[200] + ' !important',
+    '&:hover': {
+      backgroundColor: active
+        ? '#653BA3 !important'
+        : (theme.vars || theme).palette.grey[300] + ' !important',
+    },
+    ...(mode === 'light' && {
```
oktapodia, line 96: please use `...theme.applyStyles('light'` instead

src/components/Widgets/FeeContribution/FeeContribution.style.ts
```diff
@@ -0,0 +1,138 @@
...
+      backgroundColor: active ? '#F0E5FF' : theme.palette.grey[100],
```
oktapodia, line 97: theme.vars missing

src/components/Widgets/FeeContribution/FeeContribution.style.ts
```diff
@@ -0,0 +1,138 @@
...
+      '&:hover': {
+        backgroundColor: active
+          ? darken('#F0E5FF', 0.08)
+          : theme.palette.grey[300],
```
oktapodia, line 101: theme.vars missing

src/components/Widgets/FeeContribution/FeeContribution.style.ts
```diff
@@ -0,0 +1,138 @@
...
+
+export const ContributionDescription = styled(Typography)(({ theme }) => ({
+  fontSize: '12px',
+  lineHeight: '16px',
+  fontWeight: 500,
+  color: alpha('#fff', 0.84),
+  height: '40px',
+
+  ...theme.applyStyles('light', {
+    color: theme.palette.grey[700],
```
oktapodia, line 136: theme.vars

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+    pollingInterval: 1_000,
+  });
+
+  const hasTrackedImpressionRef = useRef(false);
+  const hasTrackedConfirmationRef = useRef(false);
+
+  function onChangeValue(event: React.ChangeEvent<HTMLInputElement>) {
+    const { value } = event.target;
+    // Remove dollar sign if present and format the amount
+    const valueWithoutDollar = value.replace('$', '');
```
oktapodia, line 122: you can have a leading $ instead of having it as part of the input, will be cleaner and more resilient

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+      : 0;
+    const maxUsdAmount =
+      maxTokenAmount * Number(completedRoute?.toToken?.priceUSD || 0);
+    const formattedAmount = formatInputAmount(
+      valueWithoutDollar,
+      2,
+      false,
+      maxUsdAmount,
+    );
+    setInputAmount(formattedAmount);
```
oktapodia, line 137: formatting should be displayed but only value should be store

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+    }
+  }, [completedRoute?.fromAddress]);
+
+  // Check if contribution should be shown based on:
+  // - Transaction history
+  // - AB test
+  // - Transaction amount >= $10
+  // - Chain type is EVM
+  // - Valid contribution fee address exists for the chain
+  useEffect(() => {
```
oktapodia, line 153: Prefer useMemo

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+
+  // Track contribution success
+  useEffect(() => {
+    if (isTxConfirmed && !hasTrackedConfirmationRef.current) {
+      trackEvent({
+        action: TrackingAction.ContributeSuccess,
+        category: TrackingCategory.Widget,
+        label: 'fee_contribution_success',
+        data: {
+          donator: account?.address || '',
```
oktapodia, line 223: Same as above

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+        );
+      }
+
+      // Track click event
+      trackEvent({
+        action: TrackingAction.ClickContribute,
+        category: TrackingCategory.Widget,
+        label: 'click_fee_contribution',
+        data: {
+          donator: account?.address || '',
```
oktapodia, line 278: Same as above

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+          donation_amount_usd: usdAmount || 0,
+          donation_amount_token: tokenAmount || 0,
+          donation_token_symbol: completedRoute?.toToken?.symbol || '',
+          donation_token_address: completedRoute?.toToken?.address || '',
+          donation_chain: completedRoute?.toChainId || 0,
+        },
+      });
+
+      // Convert to token units with proper precision
+      const amountInTokenUnits = parseUnits(
```
oktapodia, line 289: kinda weird as you've done the opposite just above https://github.com/jumperexchange/jumper-exchange/pull/1926/files#diff-12b87611174a4a9e1245869977b2bd9d199d9eb2e735e8b06752753ccecc53e0R255

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+      // Ensure we have a non-zero amount after conversion
+      if (parseFloat(tokenAmount) === 0) {
+        throw new Error(t('contribution.error.amountTooSmall'));
+      }
+
+      // Get the contribution fee address for the chain
+      const feeAddress = getContributionFeeAddress(completedRoute.toChainId);
+      if (!feeAddress) {
+        throw new Error(
+          'No contribution fee address configured for this chain',
```
oktapodia, line 268: Would be good to have translation as above

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+      !account?.address
+    )
+      return;
+
+    setIsSending(true);
+    try {
+      // amount is in USD, convert to token amount
+      const tokenPriceUSD = Number(completedRoute.toToken.priceUSD);
+      if (!tokenPriceUSD || tokenPriceUSD <= 0)
+        throw new Error('Invalid token price');
```
oktapodia, line 251: Would be good to have translation as above

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+            },
+          ],
+          functionName: 'transfer',
+          args: [feeAddress as `0x${string}`, amountInTokenUnits],
+          chainId: completedRoute.toChainId,
+        });
+      }
+    } catch (error) {
+      console.error('Error sending contribution:', error);
+      if (
```
oktapodia, line 321: what if it is another type of error?

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+        error.message === t('contribution.error.amountTooSmall')
+      ) {
+        alert(t('contribution.error.amountTooSmall'));
+      }
+    } finally {
+      setIsSending(false);
+    }
+  };
+
+  const isCustomActive = useMemo(() => {
```
oktapodia, line 332: Maybe change the naming of it as it is difficult to understand it's goal

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+    return (
+      !!amount &&
+      !contributionAmounts.includes(parseFloat(amount)) &&
+      parseFloat(amount) > 0
+    );
+  }, [amount]);
+
+  // Use translations from props if provided, otherwise use i18n translations
+  const getTranslation = (key: TranslationKey): string => {
+    if (key === 'error.amountTooSmall') {
```
oktapodia, line 342: Please use the classic translation system, adding another one on top of it will only increase complexity :/

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+
+                        backgroundColor: isCustomActive
+                          ? 'rgba(101, 59, 163, 0.84)'
+                          : (theme.vars || theme).palette.grey[200],
+                        '&:hover': {
+                          backgroundColor: isCustomActive
+                            ? '#653BA3'
+                            : (theme.vars || theme).palette.grey[300],
+                        },
+                        ...(mode === 'light' && {
```
oktapodia, line 450: please use `...theme.applyStyles('light'` instead

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+                          backgroundColor: isCustomActive
+                            ? '#F0E5FF'
+                            : theme.palette.grey[100],
```
oktapodia, line 453: Use theme.vars as above

src/components/Widgets/FeeContribution/helper.ts
```diff
@@ -0,0 +1,84 @@
+import { ChainType, Route } from '@lifi/sdk';
+import { Account } from '@lifi/wallet-management';
+import { MIN_CONTRIBUTION_USD } from './constants';
+import {
+  checkContributionByTxHistory,
+  getContributionFeeAddress,
+} from './utils';
+
+interface TransferResponse {
+  transfers?: any[];
```
oktapodia, line 10: Please avoid any

src/components/Widgets/Widget.tsx
```diff
@@ -336,6 +335,22 @@ export function Widget({
           integrator={config.integrator}
           config={config}
           formRef={formRef}
+          feeConfig={{
+            _vcComponent: () => (
+              <FeeContribution
+                translations={{
```
oktapodia, line 341: Please do not pass those to the components instead of using the already implemented i18n lib

src/hooks/useTxHistory.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export async function getTxHistoryQuery(walletAddress?: string) {
+  if (!walletAddress) {
+    return;
+  }
+
+  const toTimestamp = Math.floor(Date.now() / 1000); // Current timestamp in seconds
+  const fromTimestamp = toTimestamp - 30 * 24 * 60 * 60; // 30 days ago in seconds
+
+  const integrator = 'dev.jumper.exchange';
```
oktapodia, line 85: please do not hardcode it, use the env variable instead

src/hooks/useTxHistory.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  const response = await fetch(
```
oktapodia, line 87: Please use the sdk function instead

src/utils/format.ts
```diff
@@ -27,3 +27,93 @@ export function formatTokenPrice(amount?: string, price?: string) {
   }
   return parseFloat(amount) * parseFloat(price);
 }
+
+export function formatInputAmount(
```
oktapodia, line 31: please use intl numberformat instead

oktapodia (changes requested):
Couple of comments

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+  useEffect(() => {
+    if (
+      !helper.isEligibleForContribution(
+        data,
+        completedRoute ?? undefined,
+        account,
+        isContributionAbEnabled,
+      )
+    ) {
+      setShowContribution(false);
```
admsk2, line 162: shouldn't it work the way that it is `false` by default and you don't have to set it `false` explicitly, but only in case it should be `true` you set it to `true` explicitly?

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+    completedRoute?.toAmountUSD,
+    account?.chainType,
+    isContributionAbEnabled,
+    completedRoute?.toChainId,
+  ]);
+
+  // Handle contribution button click
+  const handleButtonClick = (selectedAmount: number) => {
+    if (!!amount && selectedAmount === parseFloat(amount)) {
+      setAmount(undefined);
```
admsk2, line 181: what is this for?

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+
+      if (completedRoute.toToken.address === DEFAULT_WALLET_ADDRESS) {
+        await sendTransaction({
+          to: feeAddress as `0x${string}`,
+          value: amountInTokenUnits,
+        });
+      } else {
+        await writeContract({
+          address: completedRoute.toToken.address as `0x${string}`,
+          abi: [
```
admsk2, line 302: we will eventually move this to some const

src/components/Widgets/FeeContribution/FeeContribution.style.ts
dennyscode: @oktapodia Thanks and I agree, this would be neat to have, but as this FeeContribution component is being injected into the widget, it does not seem to pick up the current themeMode from our Jumoer ThemeProvider. It´s a workaround to pass the (theme)-mode from `useColorScheme` as a styled-component prop because `...theme.applyStyles('light', {...}) `does not work.

src/components/Widgets/Widget.tsx
dennyscode: I agree that this is the first choice and should be best practise. Unfortunately, the translations within the Widget are handled by it´s own i18n instance and can not pick up our custom translation messages. That´s why this rather serves as a workaround. Wdyt?

src/hooks/useTxHistory.ts
dennyscode: THANKS, well spotted

src/utils/format.ts
dennyscode: This is a function copied over from the widget and added little customization. It handles different scenarios when using an input field such as:
- decimals using comma (,) and dot (.)
- 0 values and decimals
- adding optional max value

It feels smooth and I suggest this as re-usable function to be used on input fields events (onChange). I have used this in a different PR as well it´s been proven to work from the widget. Wdyt?

src/hooks/useTxHistory.ts
dennyscode: Nice! Thanks

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: Yes, you are right, I got it twisted when testing around to re-enable and disable it.

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: It should show a default message until an amount is either entered into the input field or selected from the buttons which sets the amount and toggles between actual value or undefined when clicked again. If the amount is undefined, it should show to the default message, otherwise show the "Contribute" button to proceed and sign the transaction.

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: I tried to make those helpers: `createTokenTransactionConfig` and `createNativeTransactionConfig`.

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+                  isTxPending ||
+                  isTxConfirming ||
+                  isNativeTxPending
+                }
+              >
+                {isTxConfirmed || isNativeTxSuccess ? <CheckIcon /> : null}
+                {isSending ||
+                isTxPending ||
+                isTxConfirming ||
+                isNativeTxPending ? (
```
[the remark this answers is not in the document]
dennyscode: Yes, good call and refactored with a const `isTransactionLoading`.

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: Yes, I agree. I am using an adornment element now.

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: I agree but please look at the comment:
https://github.com/jumperexchange/jumper-exchange/pull/1926#discussion_r2094633345

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926#discussion_r2096481863

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926#discussion_r2096481863

src/components/Widgets/FeeContribution/FeeContribution.style.ts
dennyscode: Yes, I refactored this styled component

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: I refactored it to be taken from the translation object (from props).

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: I refactored it to be taken from the translation object (from props).

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: Renamed it to "**isCustomAmountActive**"

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: I agree, this is not useful! I am taking translation messages directly from the passed props. Unfortunately, our translations dont work on this component which is injected into the widget.

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: Yes, definitely. Not sure why I did not get it working at first with applyStyles it´s working now and looking much better.

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: I agree that it´s a clean way to seperate formatting from the actual value. But the Input value depends on onChange event which triggers the formatting. Formatting also applies the check that maxAmountUsd is not exceeded. As we need to store the latest input value (in case the user selects a fixed amount after and returns back to the custom input amount), we also need to set the actual amount which is finally used as the amount to donate.

I believe it´s complicating things to seperate that logic but would love to hear your ideas.

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+      2,
+      false,
+      maxUsdAmount,
+    );
+    setInputAmount(formattedAmount);
+    setAmount(formattedAmount);
+  }
+  // Set the address of the user from last tx
+  useEffect(() => {
+    if (completedRoute?.fromAddress) {
```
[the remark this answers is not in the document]
dennyscode: Yes, I am actually doing an additional check: Check if the previous transaction (fromAddress) still matches the wallet currently connected. This should cover a case when a user disconnects a wallet and would disable the contribution component. Do you think it´s too much?

src/components/Widgets/FeeContribution/FeeContribution.style.ts
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/f6abbe8ba41c1d591f63bfe04d234ae4421b481a

src/components/Widgets/FeeContribution/FeeContribution.style.ts
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/dfe7f9aa971da864fcd529aedafed95d93b43d0a

src/components/Widgets/FeeContribution/FeeContribution.style.ts
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/dfe7f9aa971da864fcd529aedafed95d93b43d0a

src/components/Widgets/FeeContribution/FeeContribution.style.ts
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/dfe7f9aa971da864fcd529aedafed95d93b43d0a

src/components/Widgets/FeeContribution/helper.ts
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/ac2352ee3898f6d24607a1dd600db17380d56bca

src/hooks/useTxHistory.ts
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/beb1e988023d0f3fa5ab5eeb4996666d5fac36e1

src/hooks/useTxHistory.ts
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/0d570d4aaff0de4dd39821d76cd8b1081b9e81e2

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: Wdyt? @admsk2 https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/14d0853a4d798aaa3a12a3d3a3e41898d6b9f4b3

src/components/Widgets/FeeContribution/FeeContribution.tsx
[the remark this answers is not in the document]
dennyscode: Wdyt? @admsk2  https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/3c3b28cfe82c68c9481a5430dbe5e3b848c24bec

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/dfa57da424c90f72ca1f73ef2d17263ecee57c55

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/b1d32bd81abccd0a2a86d9dcf875f77bb69a7ecf

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/86445d0026236295a801d7f262f8af6828965a8a

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+                            : (theme.vars || theme).palette.grey[300],
+                        },
+                        ...(mode === 'light' && {
+                          backgroundColor: isCustomActive
+                            ? '#F0E5FF'
+                            : theme.palette.grey[100],
+                          '&:hover': {
+                            backgroundColor: isCustomActive
+                              ? darken('#F0E5FF', 0.08)
+                              : theme.palette.grey[300],
```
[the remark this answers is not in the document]
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/86445d0026236295a801d7f262f8af6828965a8a

src/components/Widgets/FeeContribution/FeeContribution.style.ts
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/89eb959eba1df4687f6b15fe08b56343ca7bcf24

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/2dee8e1cf1c22997d12451b0b9068a57ec34d724

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+      // Convert to token units with proper precision
+      const amountInTokenUnits = parseUnits(
+        tokenAmount,
+        completedRoute.toToken.decimals,
+      );
+
+      if (completedRoute.toToken.address === DEFAULT_WALLET_ADDRESS) {
+        await sendTransaction({
+          to: feeAddress as `0x${string}`,
+          value: amountInTokenUnits,
```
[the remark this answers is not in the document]
dennyscode: Wdyt?
image

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/8d79b89c40291a61dbd6cd2664cfdb77e33101ae

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/b1d32bd81abccd0a2a86d9dcf875f77bb69a7ecf

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: This should catch any error within the contribution component and provide the general message: "Error sending contribution:". This include the error object and in addition it logs a more detailed error message in some cases.
Do you think we need more error handling?

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: The original useEffect approach seems okay because it handles both related state updates in a single effect with clear dependencies, and since these values are conceptually related, using state management made their mutable nature clear. useMemo might be over-optimizing here? The useEffect approach instead was more maintainable for this specific use case where the values are coupled and part of the component's state management.
Wdyt?

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: Shoot, the link does only navigate me to the "changed files" and not the specific example you are referring to. Can you make it more clear or try to edit the link? Thanks ahead!!

src/components/Widgets/FeeContribution/FeeContribution.tsx
oktapodia: Then it would need few functions as it is mixing rendering and logic. onChange should not trigger formatting but formatting should be applied separately with a simple Intl (we have a translation already for this) here https://github.com/jumperexchange/jumper-exchange/pull/1926/files#diff-12b87611174a4a9e1245869977b2bd9d199d9eb2e735e8b06752753ccecc53e0R394

src/utils/format.ts
oktapodia: I've seen the one from the widget and this is very different, maybe ask eugene to export the widget one so you can use it, but I am not a fan of having all of that only for formatting while it can be done in roughly 3lines of code

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,504 @@
...
+
+  // Track contribution impression
+  useEffect(() => {
+    if (showContribution && !hasTrackedImpressionRef.current) {
+      trackEvent({
+        action: TrackingAction.ContributeImpression,
+        category: TrackingCategory.Widget,
+        label: 'fee_contribution_impression',
+        data: {
+          donator: account?.address || '',
```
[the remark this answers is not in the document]
oktapodia: Sorry if I wasn't clear, I am talking about the optional `?` everywhere, it is not the case there for an example https://github.com/jumperexchange/jumper-exchange/blob/b3a68d5f6f2f673c101a318e1662457d0e416f7a/src/components/Widgets/WidgetEvents.tsx#L139

src/components/Widgets/FeeContribution/FeeContribution.tsx
oktapodia: CleanShot 2025-05-22 at 11 23 21@2x
FeeContribution.tsx file:255

src/utils/transaction.ts
```diff
@@ -0,0 +1,60 @@
...
+      abi: typeof ERC20_TRANSFER_ABI;
+      functionName: 'transfer';
+      args: [Address, bigint];
+      chainId: number;
+    };
+
+/**
+ * Creates a transaction configuration for sending native currency
+ */
+export function createNativeTransactionConfig(
```
oktapodia, line 34: What is the goal of that function?

src/utils/transaction.ts
```diff
@@ -0,0 +1,60 @@
+import { type Address } from 'viem';
+
+// ERC20 Transfer ABI - commonly used for token transfers
+export const ERC20_TRANSFER_ABI = [
```
oktapodia, line 4: You can get the abi directly from viem

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,506 @@
...
+    } catch (error) {
+      console.error(translations.error.errorSending, error);
+      if (error instanceof Error) {
+        const errorMessage = error.message;
+        if (errorMessage === translations?.error?.amountTooSmall) {
+          console.error(errorMessage);
+        } else if (errorMessage === translations?.error?.noFeeAddress) {
+          console.error(errorMessage);
+        } else if (errorMessage === translations?.error?.invalidTokenPrice) {
+          console.error(errorMessage);
```
oktapodia, line 317: they needs to be reported to sentry, just doing console.error won't help us to fix it :s

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,506 @@
...
+  }, [amount]);
+
+  if (!showContribution) {
+    return null;
+  }
+
+  return (
+    <ContributionWrapper showContribution={showContribution}>
+      <DrawerWrapper ref={containerRef}>
+        <Drawer
```
oktapodia, line 340: Would be good to use styled

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,506 @@
...
+              sx: {
+                position: 'absolute',
+                bottom: 0,
+                left: 0,
+                right: 0,
+                width: '100%',
+                borderRadius: '24px',
+              },
+            },
+            backdrop: {
```
oktapodia, line 362: if I understand correctly, you've disabled the backdrop so this might not be necessary

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,506 @@
...
+              <Grid size={3}>
+                <ContributionCustomInput
+                  value={inputAmount || ''}
+                  aria-autocomplete="none"
+                  onChange={onChangeValue}
+                  onClick={handleCustom}
+                  onFocus={handleCustom}
+                  placeholder={translations.custom}
+                  slotProps={{
+                    root: {
```
oktapodia, line 401: same as above

src/components/Widgets/FeeContribution/helper.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+export const isSameWalletTransfer = (
+  fromAddress?: string,
+  toAddress?: string,
+): boolean => {
+  return fromAddress === toAddress;
+};
+
+export const isWalletAccessValid = (
```
oktapodia, line 35: This function and the above one does the exact same thing

src/components/Widgets/FeeContribution/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+  CONTRIBUTION_AMOUNTS,
+  VOLUME_THRESHOLDS,
+  contributionFeeAddresses,
+} from './constants';
+
+/**
+ * Determines if contribution should be shown based on transaction history
+ * Shows contribution for:
+ * - First transaction (count = 0)
+ * - Every third transaction (count % 3 = 0)
```
oktapodia, line 12: Are you sure about this? We already apply a 10% rule as ab testing above https://github.com/jumperexchange/jumper-exchange/pull/1926/files#diff-12b87611174a4a9e1245869977b2bd9d199d9eb2e735e8b06752753ccecc53e0R70

So with this on top of it, it should appear only 3.33% (with current settings)

src/components/Widgets/FeeContribution/FeeContribution.tsx
oktapodia: are you sure about it? it is doing a lot of checks (async or not), then getting to the `isEligibleForContribution` function which is checking some more things that can be check at the top of the component (or hook), will permits us to avoid those async calls in a lot of case (isSameAddress, isEVMAddress, etc...)

tests/testData/settingsFunctions.ts
```diff
@@ -1,9 +1,7 @@
 import { expect, Locator, Page } from '@playwright/test';
```
oktapodia, line 1: please do not format files that you are not editing because it is making it harder to review

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: True dat! I have added some `captureException`: Some questions on that:

- is it not necessary to track further information like I did here, because the log may already reveal the file + line?
image
- I am not 100% sure which errors to track and which may be overhead to be sent to Sentry

src/components/Widgets/FeeContribution/helper.ts
dennyscode: That´s true! Meant for different reasons but technically not even necessary for such simple comparison. Removed both functions.

src/utils/transaction.ts
dennyscode: Great idea! Did not know that!

src/components/Widgets/FeeContribution/utils.ts
dennyscode: Well spotted!! I double checked with Marko and the rest of the guys and it´s okay to roll out this feature to a very small group of users first

src/utils/format.ts
dennyscode: You´re right! I have refactored it and seperated it into different concerns:
- logic: sanitizing input mostly using regex
- formatting: using Intl for formatting

src/components/Widgets/FeeContribution/helper.ts
```diff
@@ -0,0 +1,85 @@
+import { ChainType, Route, StatusResponse } from '@lifi/sdk';
+import { Account } from '@lifi/wallet-management';
+import { MIN_CONTRIBUTION_USD } from './constants';
+import {
+  checkContributionByTxHistory,
+  getContributionFeeAddress,
+} from './utils';
+
+export interface TransferResponse {
+  transfers: StatusResponse[] | undefined;
```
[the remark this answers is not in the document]
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/6db816b6bf860cd4527ffb569711d61c19011525

src/components/Widgets/FeeContribution/helper.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+export const hasValidContributionFeeAddress = (chainId: number): boolean => {
+  const contributionFeeAddress = getContributionFeeAddress(chainId);
+  return !!contributionFeeAddress;
+};
+
+// todo: Uncomment before merging this PR to re-enable checks
+export const isEligibleForContribution = (
+  data: TransferResponse | null | undefined,
```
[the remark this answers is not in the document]
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/6db816b6bf860cd4527ffb569711d61c19011525

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/9a5f9da0eb83d1bde59c84f23f2a96e312ea0299

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/b6a60cfbc3b6b8942d13b9ba50084830365f6d38

src/utils/transaction.ts
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/4541363f159e01bb6e417716ac530711f683b18d

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: I agree! I believe I have simplified that logic a lot! Great feedback!

https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/b5325b5f085440e7b0f4747b2fd71ce40b5dfb12

src/utils/format.ts
dennyscode: [z](https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/b5325b5f085440e7b0f4747b2fd71ce40b5dfb12)

tests/testData/settingsFunctions.ts
dennyscode: Shoot, yes!

src/components/Widgets/FeeContribution/helper.ts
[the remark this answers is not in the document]
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/6db816b6bf860cd4527ffb569711d61c19011525

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: Hmm, sorry. I am not sure I fully understand: Are you referring why I was miximg up "parseFloat(...)" and some other times used "Number(...)"?

I have refactored that and am only using Number(...). Is that what you would recommend?

Besides, the `const amountInTokenUnits = parseUnits(...)` is only being used to calculate the actual tokenAmount used for the tx.

dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/29d395e40d7d36f5c430840f1a882f54f92361ed

src/components/Widgets/FeeContribution/FeeContribution.tsx
dennyscode: I have given it another shot by adding some useMemo´s to handle such states: Ultimately, I still end up with a useEffect to handle the state "isEligible":
image

https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/31b4fb064965f391e8224fc3d192da35a231bddd

src/components/Widgets/FeeContribution/FeeContribution.tsx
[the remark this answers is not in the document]
dennyscode: I checked in with Ali and it should be okay to have that check in place. Let´s keep a close eye on it, but it should not cause much issues but rather prevent them.

src/components/Widgets/FeeContribution/helper.ts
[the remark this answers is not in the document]
oktapodia: This is not fixed

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,495 @@
...
+      completedRoute &&
+      showContribution &&
+      !hasTrackedImpressionRef.current
+    ) {
+      trackEvent({
+        action: TrackingAction.ContributeImpression,
+        category: TrackingCategory.Widget,
+        label: 'fee_contribution_impression',
+        data: {
+          donator: account.address || '',
```
oktapodia, line 209: Same comment as before, this should not be optional fields (`|| ''`)

src/components/Widgets/FeeContribution/FeeContribution.tsx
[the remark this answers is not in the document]
oktapodia: You still have the `|| ''` everywhere

tests/settings.spec.ts
```diff
@@ -15,29 +15,30 @@ const SETTINGS_MENU = {
   ROUTE_PRIORITY: {
```
oktapodia, line 15: Same comment as before regarding this, do not change files that you are not updating please

tests/testData/settingsFunctions.ts
```diff
@@ -16,27 +14,33 @@ export async function checkItemInSettingsMenu(
 }
 
 export async function checkSettingInSettingsMenu(
```
oktapodia, line 16: Same comment as before regarding this, do not change files that you are not updating please

src/components/Widgets/FeeContribution/FeeContribution.tsx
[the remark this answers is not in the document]
dennyscode: Sorry, I was taking you by the word -> "?" :D I cleaned it up

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,495 @@
...
+      // amount is in USD, convert to token amount
+      const tokenPriceUSD = Number(completedRoute.toToken.priceUSD);
+      const sentryHint = {
+        tags: {
+          component: 'FeeContribution',
+          action: 'handleConfirm',
+        },
+      };
+      if (!tokenPriceUSD || tokenPriceUSD <= 0) {
+        isProduction &&
```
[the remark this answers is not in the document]
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/37cdfbf1d69786df388688e1ce6352eef9d83ba3

src/components/Widgets/FeeContribution/FeeContribution.tsx
```diff
@@ -0,0 +1,521 @@
...
+          value: amountInTokenUnits,
+        });
+      } else {
+        // Type assertion to ensure we're using the token transaction config
+        const txConfig = createTokenTransactionConfig(
+          completedRoute.toToken.address as `0x${string}`,
+          feeAddress as `0x${string}`,
+          amountInTokenUnits,
+          completedRoute.toChainId,
+        ) as {
```
oktapodia, line 357: why casting it here while you've created a function to generate it?

src/components/Widgets/FeeContribution/helper.ts
```diff
@@ -0,0 +1,125 @@
...
+
+// Add validation function that handles max value check and special cases
+export const validateInputAmount = (
+  value: string,
+  decimals?: number,
+): string => {
+  if (!value) return '';
+
+  // Sanitize the input value
+  const { cleanValue } = sanitizeNumberField(value);
```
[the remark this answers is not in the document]
dennyscode: Thanks!

1.) I tried to wrap my head around a solution to solve with Intl as you proposed. It comes with the following issues:
- multiple leading zeros (not sanitized)
- adding decimal (e.g. " 0. " -> ending with a dot or comma) returns NaN
- no "locale aware formatting" needed as we have a strict a
--> This still requires more handling to serve a smooth UX

2.) Adding a `type="number"` seems great at first and remove necessity of sanitizing non-digit characters. The issue that comes a long is, it turns decimal seperator into a comma (e.g. `1,00`).

My suggestion:
a.) Since, we´re already using zod: Do you think creating a general zod input schema could be a better solution?
b.) Keep using regex, but optimizing (the state you reviewed may have room to be improved)

Wdyt?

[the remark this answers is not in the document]
dennyscode: Parts of it. Whereas it helps with formating numbers, it does not fully cover validation and sanitizing. In order to not lose quality of UX we need to add that on top. 

In the last approach I have added the `Intl.NumberFormat` to display the amount. I believe this is a good middle ground between both worlds: Handling formatting by Intl but still not lose quality on the user input validation + sanitizing. Wdyt?

https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/a851135dff50e64af62d9b8260c6949bd5b4e067

src/components/Widgets/FeeContribution/hooks.ts
```diff
@@ -0,0 +1,403 @@
...
+
+      // Convert to token units with proper precision
+      const amountInTokenUnits = parseUnits(
+        tokenAmount,
+        completedRoute.toToken.decimals,
+      );
+
+      if (completedRoute.toToken.address === DEFAULT_WALLET_ADDRESS) {
+        const nativeTxConfig = createNativeTransactionConfig(
+          feeAddress as `0x${string}`,
```
oktapodia: Would be good to extract this in its own type

dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1926/commits/2c2871aaddca99fa03d1e6bfc9adc38af920e5e4

vinzenzLIFI:
Tested fee contribution without checks on current feature branch for on-chain and cross chain swaps on EVM. The contribution works correctly and looks like the figma design. the custom field detects the maximum available amount and sets it when a value higher than that is entered. otherwise the buttons will be highlighted if a value of same amount as the presets is entered.  ✅

one thing to note is, that there are many chains that do not have a contribution address maintained, such as BOB and APE. The checks will filter them out. I could not verify that yet. Otherwise the fee contribution modal will be displayed for those chains, but the transaction cannot be triggered (obviously)

vinzenzLIFI:
Tested latest changes without activated checks. 

- Confirm button now requires a set donation amount, otherwise it will be inactive
- chain switch is fixed when the current chain is not the same as the destination chain

mmioana (commented):
Working really smooth, nice work on this one @dennyscode  🎉 Only the line-height value for the options should be updated and we're good to go.

https://github.com/user-attachments/assets/bd5e4396-ffd8-4729-a0d4-61284d3ea5d1

src/components/Widgets/FeeContribution/FeeContribution.style.ts
```diff
@@ -197,8 +197,8 @@ export const ContributionButtonCTA = styled(Button, {
 })<ContributionButtonCTAProps>(
   ({ theme }) => ({
     width: '100%',
-    fontSize: '14px',
-    lineHeight: '16px',
+    fontSize: '12px',
+    lineHeight: '12px',
```
mmioana: Please revert this

vinzenzLIFI (approved):
Tested fee contribution feature including all checks. ✅
The AB check was not e2e tested, because of feasibility. Instead the AB test was inspected with @dennyscode using console logs to see that the correct probability range is generated for the feature to appear.

src/components/Widgets/FeeContribution/FeeContribution.style.ts
```diff
@@ -0,0 +1,269 @@
...
+}));
+
+interface ContributionCustomInputProps {
+  isFieldActive?: boolean;
+  inputAmount?: string;
+  value?: string;
+}
+
+export const ContributionCustomInput = styled(TextField, {
+  shouldForwardProp: (prop) => !['isFieldActive'].includes(prop as string),
```
oktapodia, line 55: value might need to be added there as well

oktapodia (approved):
LGTM

PR #1949, opened 2025-05-16
refactor: cleanup apiAccessToken

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-13534

### DIFF
--- a/src/utils/strapi/StrapiApi.ts
+++ b/src/utils/strapi/StrapiApi.ts
@@ -42,21 +42,6 @@ class StrapiApi {
     }
   }
 
-  /* todo:
-   move this method to strapiHelper
-   use on apiAccessToken and replace getApiAccessToken() calls across the app
-   remove this method after migration
-   */
-  public getApiAccessToken(): string {
-    if (process.env.NEXT_PUBLIC_STRAPI_DEVELOP === 'true') {
-      // Use local-strapi-api token for development environment
-      return process.env.NEXT_PUBLIC_LOCAL_STRAPI_API_TOKEN || '';
-    } else {
-      // Use default STRAPI API token for other environments
-      return process.env.NEXT_PUBLIC_STRAPI_API_TOKEN || '';
-    }
-  }
-
   private getBaseUrl(): string {
     if (process.env.NEXT_PUBLIC_STRAPI_DEVELOP === 'true') {
       // Use local Strapi URL for development environment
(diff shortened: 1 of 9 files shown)


src/utils/strapi/StrapiApi.ts
```diff
@@ -42,21 +42,6 @@ class StrapiApi {
...
-    if (process.env.NEXT_PUBLIC_STRAPI_DEVELOP === 'true') {
-      // Use local-strapi-api token for development environment
-      return process.env.NEXT_PUBLIC_LOCAL_STRAPI_API_TOKEN || '';
-    } else {
-      // Use default STRAPI API token for other environments
-      return process.env.NEXT_PUBLIC_STRAPI_API_TOKEN || '';
-    }
-  }
-
   private getBaseUrl(): string {
```
oktapodia, line 45: If you remove the above, this one should also be removed

dennyscode: Thanks for spotting it!
The method `getBaseUrl()` is already being refactored to re-use `getStrapiBaseUrl()` from strapiHelper.ts.
In addition, this method provides better logging when instantiating. Wdyt?

I believe, although you´re right in a way, there might be some compromise?

dennyscode: All right, yes no half baked things!
So, I removed the `getBaseUrl` and `getApiAccessToken` method from the StrapiApi class and use helpers now.

src/utils/strapi/StrapiApi.ts
```diff
@@ -42,21 +42,6 @@ class StrapiApi {
     }
   }
 
-  /* todo:
-   move this method to strapiHelper
-   use on apiAccessToken and replace getApiAccessToken() calls across the app
-   remove this method after migration
-   */
-  public getApiAccessToken(): string {
```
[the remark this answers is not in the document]
dennyscode: For baseUrl and accessToken, we´re using the strapi helpers now.
Only for the apiUrl which is specific to the content-type is still needed.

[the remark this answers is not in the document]
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/1949/commits/d2bc6d767870e6f53c58c437c2f5357814d86d77

admsk2 (approved):

oktapodia (approved):
LGTM

PR #1959, opened 2025-05-21
feat: add e2e for hyperliquid chain where check BTC/SOL/SUI/EVM

callMeTheQA:
Added a checks for swap  / bridge action for hyperliquid chain where : 
EVM -> Hyperliquid should load routes 
SOL -> Hyperliquid should not load routes 
SUI -> Hyperliquid should not load routes 
BTC -> Hyperliquid should not load routes


oktapodia (approved):
LGTM

vinzenzLIFI (approved):
verified the test locally. looks good

PR #1964, opened 2025-05-29
chore: hide toAddress on HYP (Hypercore)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-13896


src/hooks/useTokenSelection.ts
mmioana, line 1: Really nice work on creating this hook @dennyscode 🙌 
Would it make more sense to not retrieve also the `toAddress`, given the hook's responsibility is to enhance the source and destination tokens details?
It might also be worthwhile to retrieve the `chainType` for each of these tokens, since in the future we may be interested in more than just whether they're EVM or not

oktapodia (approved):
LGTM

mmioana (approved):
Let's get this merged 🎉

PR #1985, opened 2025-06-06
feat: jumper improve blog article retrieval to not include the content for the article list

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-14023

### DIFF
--- a/src/utils/strapi/StrapiApi.ts
+++ b/src/utils/strapi/StrapiApi.ts
@@ -108,20 +108,24 @@ type ArticleField =
   | 'createdAt'
   | 'updatedAt';
 
+const mainArticleFields: ArticleField[] = [
+  'Title',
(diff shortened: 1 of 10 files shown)


src/utils/strapi/StrapiApi.ts
```diff
@@ -95,19 +95,73 @@ class StrapiApi {
...
 class ArticleParams {
   private apiUrl: URL;
 
+  private static defaultFields: ArticleField[] = [
+    'Title',
+    'Subtitle',
+    'Slug',
+    'RedirectURL',
+    'featured',
+    // 'WordCount',
```
mmioana, line 119: This needs to be uncommented before merging, but only after https://github.com/jumperexchange/strapi-cms/pull/54 goes in ⚠️

PR #1995, opened 2025-06-12
fix: override toAddress with poolName for zap widget

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-14093


oktapodia (approved):
LGTM

PR #1998, opened 2025-06-13
feat: create update mission card variants

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-14110

And added [Storybook](https://storybook.js.org/)
So please run `pnpm build-storybook` and then `pnpm storybook` to see the new components ✨

### DIFF
--- a/src/components/Badge/Badge.tsx
+++ b/src/components/Badge/Badge.tsx
@@ -1,5 +1,5 @@
 import { FC } from 'react';
-import { StyledBadge, StyledBadgeLabel } from './Badge.styled';
+import { StyledBadge, StyledBadgeLabel } from './Badge.styles';
 
 export interface BadgeProps {
   startIcon?: React.ReactElement;
--- a/src/components/Badge/Badge.tsx
+++ b/src/components/Badge/Badge.tsx
@@ -1,0 +1,18 @@
+import { FC } from 'react';
+import { StyledBadge, StyledBadgeLabel } from './Badge.styled';
+
+export interface BadgeProps {
+  startIcon?: React.ReactElement;
+  endIcon?: React.ReactElement;
+  label: string;
+}
+
+export const Badge: FC<BadgeProps> = ({ label, startIcon, endIcon }) => {
+  return (
+    <StyledBadge>
+      {startIcon && <>{startIcon}</>}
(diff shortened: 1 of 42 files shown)


src/components/Badge/Badge.tsx
dennyscode, line 1: I just discovered the <Chip> component provided by MUI. Do you think it could replace this custom component?
https://mui.com/material-ui/react-chip/

src/components/Cards/EntityCard/constants.ts
dennyscode, line 1: I like that those measures are to be found in one spot!

src/components/Cards/EntityCard/variants/CompactEntityCard.tsx
```diff
@@ -0,0 +1,120 @@
...
+  return (
+    <StyledEntityCard
+      sx={{
+        width: ENTITY_CARD_SIZES.COMPACT.CARD_WIDTH,
+      }}
+      onClick={onClick}
+    >
+      <StyledEntityCardImageContainer
+        width={ENTITY_CARD_SIZES.COMPACT.CARD_WIDTH}
+        height={ENTITY_CARD_SIZES.COMPACT.IMAGE_HEIGHT}
```
dennyscode, line 43: I am wondering if a "variant" of those styled-components (where custom ENTITY_CARD_SIZES sizes are passed) could be a cleaner way to maintain?

src/theme/theme.ts
```diff
@@ -1037,9 +1041,9 @@ export const themeCustomized: Omit<Theme, 'applyStyles'> & CssVarsTheme =
             dark: '#FFFFFF',
           },
           surface3: {
-            light: '#F6F0FF',
-            main: '#F6F0FF',
-            dark: '#F6F0FF',
+            light: '#E5E1EB',
```
oktapodia, line 1044: make sure to double check it as it can be widely used across the app

oktapodia (approved):
LGTM

src/theme/theme.ts
mmioana: Yes, I have checked and it's only used in the `NavbarButtons`, though now I am wondering why I have updated it in this PR - might have overlooked doing the change in the previous one

PR #2012, opened 2025-06-24
feat: add carousel for campaign banners on profile page

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-14267

### DIFF
--- a/src/components/ProfilePage/CampaignBanners/CampaignBanners.style.ts
+++ b/src/components/ProfilePage/CampaignBanners/CampaignBanners.style.ts
@@ -9,7 +9,6 @@ export const CampaignBannersContainer = styled(Box)(({ theme }) => ({
   display: 'flex',
   gap: theme.spacing(3),
   borderRadius: '24px',
-  // borderRadius: '32px',
   flexDirection: 'column',
   boxShadow: theme.shadows[1],
 


src/components/Blog/BlogCarousel/BlogCarousel.tsx
```diff
@@ -38,10 +38,7 @@ export const BlogCarousel = ({
 
   return (
     <BlogCarouselContainer>
-      <CarouselContainer
-        title={title || t('blog.recentPosts')}
-        trackingCategory={TrackingCategory.BlogCarousel}
```
dennyscode, line 43: `trackingCategory` in CarouselContainer was unused

src/components/Blog/BlogCarousel/CarouselContainer.tsx
```diff
@@ -42,6 +39,11 @@ export const CarouselContainer = ({
   const handleChange = useCallback((direction: 'next' | 'prev') => {
     if (carouselContainerRef.current) {
       const node: HTMLDivElement = carouselContainerRef.current;
+      const children = Array.from(carouselContainerRef.current.children);
+      const swipeDistance =
+        Array.isArray(children) && children.length > 0
+          ? children[0]?.clientWidth
+          : 420;
```
dennyscode, line 46: This should be a better solution to swipe the actual distance of one item. As a fallback, we may still use that 420. Wdyt?

PR #2028, opened 2025-06-30
fix: MEE Client Init;

admsk2:


oktapodia (approved):

src/components/ZapWidget/ZapWidget.tsx
```diff
@@ -518,13 +575,13 @@ export function ZapWidget({
         instructions,
       });
 
-      const { hash } = await meeClient.executeFusionQuote({
+      const { hash } = await meeClient!.executeFusionQuote({
```
mmioana, line 578: Here you could've used `client` as well instead of `meeClient!`

PR #2038, opened 2025-07-04
feat: mission supporting UI components

mmioana:
Resolves https://lifi.atlassian.net/browse/LF-14111

Please run `pnpm build-storybook` and then `pnpm storybook` to see the new components ✨

### DIFF
--- a/src/components/Alerts/BaseAlert/BaseAlert.tsx
+++ b/src/components/Alerts/BaseAlert/BaseAlert.tsx
@@ -23,7 +23,7 @@ interface BaseAlertProps extends PropsWithChildren {
 
 export const BaseAlert: FC<BaseAlertProps> = ({
   title,
-  variant = 'default',
+  variant = BaseAlertVariant.Default,
   headerAppend,
   headerPrepend,
   description,
@@ -39,14 +39,14 @@ export const BaseAlert: FC<BaseAlertProps> = ({
     }
 
     switch (variant) {
-      case 'warning': {
+      case BaseAlertVariant.Warning: {
         return <WarningRoundedIcon sx={{ height: 24, width: 24 }} />;
       }
-      case 'error': {
+      case BaseAlertVariant.Error: {
         return <ErrorRoundedIcon sx={{ height: 24, width: 24 }} />;
       }
-      case 'info':
-      case 'hint': {
+      case BaseAlertVariant.Info:
+      case BaseAlertVariant.Hint: {
         return (
           <InfoIcon
             sx={(theme) => ({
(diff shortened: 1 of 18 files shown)


src/components/Alerts/BaseAlert/BaseAlert.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const hasTitle = !!title;
+  const hasDescription = !!description;
+
+  const getHeaderAppend = () => {
+    if (headerAppend) {
+      return headerAppend;
+    }
+
+    switch (variant) {
+      case 'warning': {
```
oktapodia: Would be good to have an enum

src/components/Cards/TaskCard/TaskCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+  type?: string;
+  statusBadge?: ReactNode;
+  onClick?: () => void;
+  isLoading?: boolean;
+}
+
+export const TaskCard: FC<TaskCardProps> = ({
+  title,
+  description,
+  isActive,
```
oktapodia: Might be good to set some default value as fields are optional to avoid issues (e.g. isActive might be undefined which might create some side effects)

src/components/Cards/TaskCard/TaskCard.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+      />
+    ),
+  },
+};
+
+export const WithTaskTypes: Story = {
+  render: () => (
+    <Stack spacing={2}>
+      <TaskCard
+        title="Bridge ETH"
```
oktapodia, line 90: Would be good to use translation

oktapodia (changes requested):
Couple of comments

src/components/Cards/TaskCard/TaskCard.stories.tsx
mmioana: Do we want to use translations for the stories @oktapodia?

src/theme/theme.ts
```diff
@@ -517,6 +540,69 @@ const palette = {
...
+    100: '#FEF5FF',
+    200: '#FCEBFF',
+    300: '#FBE0FF',
+    400: '#FAD6FF',
+    500: '#F7C2FF',
+    600: '#C86FD6',
+    700: '#AA51B8',
+    800: '#702C7A',
+    900: '#37113D',
+  },
```
dennyscode, line 605: Are those color definitions based on figma?

mmioana: @dennyscode yes, these are the colour variables from Figma, though I haven’t yet semantically mapped them. The idea was to have access to them at least ftm, and in a separate task we should align all colours to the design system

dennyscode (approved):

src/components/Cards/TaskCard/TaskCard.stories.tsx
oktapodia: oh mb

oktapodia (approved):

PR #2041, opened 2025-07-04
feat: add widget variants for missions and zaps

mmioana:

This PR sets up the base configuration for a widget that we might use on various pages like mission, zap, and others. The configuration is organized across multiple hooks to keep things focused and make it easier to manage specific changes.

Depending on the context, we can apply overrides by enabling the relevant hooks. For example, the zap page requires some additional props, so its hook can include those extras. In addition to the hook-based setup, the widget is also configurable through the baseOverrides provided by context. This allows us to tweak specific props without needing to alter the entire configuration.

In the `missions` folder the widget component depends on the mission store, which further configures the widget context based on that data.

At this stage, only the Storybook stories have been added for the `base` widgets ✨ so go ahead and run `pnpm build-storybook` and then `pnpm storybook`

### DIFF
--- a/.storybook/withProviders.tsx
+++ b/.storybook/withProviders.tsx
@@ -5,6 +5,7 @@ import {
   MUIThemeProvider,
 } from '../src/providers/ThemeProvider';
 import TranslationsProvider from '../src/providers/TranslationProvider';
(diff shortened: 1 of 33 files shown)


oktapodia (approved):
LGTM

PR #2042, opened 2025-07-04
fix: initial ssr rendering for wallet tracking provider

mmioana:
Solves
Screenshot 2025-07-04 at 12 46 46


dennyscode (approved):
Thanks for resolving this!

PR #2043, opened 2025-07-04
feat: lifi sdk config store

mmioana:

### DIFF
--- a/src/providers/WalletProvider/WalletProvider.tsx
+++ b/src/providers/WalletProvider/WalletProvider.tsx
@@ -22,18 +22,17 @@ import { useTranslation } from 'react-i18next';
 import { EVMProvider } from './EVMProvider';
 import { SVMProvider } from './SVMProvider';
 import { UTXOProvider } from './UTXOProvider';
-import { createConfig } from '@lifi/sdk';
 import { SuiProvider } from './SuiProvider';
 import { WalletManagementThemeProvider } from '@/providers/ThemeProvider';
 import { useSdkConfigStore } from 'src/stores/sdkConfig/SDKConfigStore';
 import { ClientOnly } from 'src/components/ClientOnly';
 
 export const WalletProvider: FC<PropsWithChildren> = ({ children }) => {
-  const { config } = useSdkConfigStore();
+  const { initSdk } = useSdkConfigStore();
 
   useEffect(() => {
-    createConfig(config);
-  }, [config]);
+    initSdk();
+  }, []);
 
   return (
     <EVMProvider>
--- a/src/providers/WalletProvider/WalletProvider.tsx
+++ b/src/providers/WalletProvider/WalletProvider.tsx
@@ -24,16 +24,9 @@ import { SVMProvider } from './SVMProvider';
 import { UTXOProvider } from './UTXOProvider';
 import { SuiProvider } from './SuiProvider';
 import { WalletManagementThemeProvider } from '@/providers/ThemeProvider';
-import { useSdkConfigStore } from 'src/stores/sdkConfig/SDKConfigStore';
 import { ClientOnly } from 'src/components/ClientOnly';
 
 export const WalletProvider: FC<PropsWithChildren> = ({ children }) => {
-  const { initSdk } = useSdkConfigStore();
-
-  useEffect(() => {
-    initSdk();
-  }, []);
-
   return (
     <EVMProvider>
       <UTXOProvider>
--- a/src/providers/WalletProvider/WalletProvider.tsx
+++ b/src/providers/WalletProvider/WalletProvider.tsx
@@ -22,25 +22,18 @@ import { useTranslation } from 'react-i18next';
 import { EVMProvider } from './EVMProvider';
 import { SVMProvider } from './SVMProvider';
 import { UTXOProvider } from './UTXOProvider';
-import { createConfig, EVM, Solana, UTXO, Sui } from '@lifi/sdk';
-import { publicRPCList } from '@/const/rpcList';
-import getApiUrl from '@/utils/getApiUrl';
+import { createConfig } from '@lifi/sdk';
 import { SuiProvider } from './SuiProvider';
 import { WalletManagementThemeProvider } from '@/providers/ThemeProvider';
+import { useSdkConfigStore } from 'src/stores/sdkConfig/SDKConfigStore';
 import { ClientOnly } from 'src/components/ClientOnly';
 
 export const WalletProvider: FC<PropsWithChildren> = ({ children }) => {
-  createConfig({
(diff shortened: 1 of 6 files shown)


oktapodia (approved):
LGTM

dennyscode (approved):

src/providers/WalletProvider/WalletProviderZap.tsx
```diff
@@ -3,23 +3,13 @@ import { type FC, type PropsWithChildren, useEffect } from 'react';
...
-      apiUrl: process.env.NEXT_PUBLIC_LIFI_API_URL,
-      providers: [EVM(), Solana(), UTXO()],
-      integrator: process.env.NEXT_PUBLIC_WIDGET_INTEGRATOR,
-      rpcUrls: {
-        ...JSON.parse(process.env.NEXT_PUBLIC_CUSTOM_RPCS),
-        ...publicRPCList,
-      },
-      preloadChains: true,
-    });
+    setConfigType('zap');
```
chybisov: Why do we switch config type here but not the actual config in the SDK, like in the standard `WalletProvider`?

src/providers/WalletProvider/WalletProvider.tsx
```diff
@@ -22,25 +22,18 @@ import { useTranslation } from 'react-i18next';
...
-    rpcUrls: {
-      ...JSON.parse(process.env.NEXT_PUBLIC_CUSTOM_RPCS),
-      ...publicRPCList,
-    },
-    preloadChains: true,
-  });
+  const { config } = useSdkConfigStore();
+
+  useEffect(() => {
+    createConfig(config);
```
chybisov: If we want to change some properties of the config when switching between pages it's better to call `config.set`. Please see https://docs.li.fi/sdk/configure-sdk#update-sdk-configuration

Generally, it is better to just create it once and update from there.

src/stores/sdkConfig/SDKConfigStore.ts
```diff
@@ -0,0 +1,72 @@
...
+  apiKey: process.env.NEXT_PUBLIC_LIFI_API_KEY,
+  apiUrl: process.env.NEXT_PUBLIC_LIFI_API_URL,
+  providers: [EVM(), Solana(), UTXO()],
+  integrator: process.env.NEXT_PUBLIC_WIDGET_INTEGRATOR,
+  rpcUrls: {
+    ...JSON.parse(process.env.NEXT_PUBLIC_CUSTOM_RPCS),
+    ...publicRPCList,
+  },
+  preloadChains: true,
+};
```
chybisov: If the difference between them is only an API URL, maybe we should have one config and just change the URLs?

src/stores/sdkConfig/SDKConfigStore.ts
```diff
@@ -0,0 +1,72 @@
...
+  },
+  config: defaultConfig,
+
+  setConfigType: (type) => {
+    const configMap = get().configMap;
+    const nextConfig = configMap[type];
+    if (!nextConfig) throw new Error(`No config defined for type: ${type}`);
+
+    set({ configType: type, config: nextConfig });
+  },
```
chybisov: I'm not sure if we even need the config store, since the config itself is a singleton, which means we can't have two instances for now, but if we want to have some abstraction over it, I'd move the creation of the config and full management to this store, instead of having to manage it in some useEffects in many places, and call all config create/set functions here.

chybisov (dismissed):

src/providers/WalletProvider/WalletProviderZap.tsx
mmioana: The `WalletProvider` is invoked in the main app layout, which means that even on the zap page, both `WalletProvider` and `WalletProviderZap` will be instantiated. To avoid duplicating the configuration logic, it's better to have the setup in a single location. So simply changing the type will trigger the effect in `WalletProvider` to update the configuration accordingly.

src/providers/WalletProvider/WalletProvider.tsx
mmioana: I'm updating the store to have an init function for the sdk and directly update the sdk config when changing the type

src/stores/sdkConfig/SDKConfigStore.ts
mmioana: The `apiUrl` is actually more or less the same, but the list of providers is different, but you're right. It's better to clearly separate the differences and update the config accordingly

src/stores/sdkConfig/SDKConfigStore.ts
mmioana: I think having this config store makes sense, especially since we might need different SDK configs depending on the path or specific requirements — like one mission involving a zap, and another just a simple bridge or swap. It gives us better control and makes it easier to set or reset the config as needed.

src/stores/sdkConfig/SDKConfigStore.ts
```diff
@@ -0,0 +1,111 @@
...
+  rpcUrls: {
+    ...JSON.parse(process.env.NEXT_PUBLIC_CUSTOM_RPCS),
+    ...publicRPCList,
+  },
+  preloadChains: true,
+};
+
+const zapConfig: Partial<SDKConfig> = {
+  providers: [EVM(), Solana(), UTXO()],
+};
```
chybisov: Zap can also have Sui, it does not matter in this context. So probably should be removed.

src/providers/WalletProvider/WalletProvider.tsx
```diff
@@ -22,25 +22,17 @@ import { useTranslation } from 'react-i18next';
...
-    rpcUrls: {
-      ...JSON.parse(process.env.NEXT_PUBLIC_CUSTOM_RPCS),
-      ...publicRPCList,
-    },
-    preloadChains: true,
-  });
+  const { initSdk } = useSdkConfigStore();
+
+  useEffect(() => {
+    initSdk();
```
chybisov: I think we can just init the SDK during store creation, why do we need to use useEffeect for this.

src/stores/sdkConfig/SDKConfigStore.ts
mmioana: @admsk2 is there any reason Sui was not among the provider list? 🤔 If there isn't, surely I'll remove it 🙌

src/providers/WalletProvider/WalletProvider.tsx
mmioana: Fair fair fair, on it 🚀

mmioana:
After talking w/ @chybisov will close this PR in favor of a more simple approach to instantiate the SDK config at top level + cleanup

PR #2044, opened 2025-07-04
feat: mission details page

mmioana:

Currently the widget is not shown until https://github.com/jumperexchange/jumper-exchange/pull/2041

> [!NOTE]
> From the `/missions` page, given the quests already have a link attached to them, please copy a quest slug and go to `/missions/[slug]` (eg. `/missions/rewards-from-bonsai` or `/missions/golden-ticket`)
>
> Also, note some information as for eg. the taskType is not yet added in `Strapi`


src/i18n/translations/en/translation.json
```diff
@@ -152,6 +152,13 @@
       "hoursLeft_other": "{{count}} hours left",
       "daysLeft_one": "{{count}} day left",
       "daysLeft_other": "{{count}} days left"
+    },
+    "tasks": {
+      "type": "{{type}} task",
+      "status": {
+        "verify": "Verify",
+        "verified": "Verified"
+      }
```
dennyscode, line 161: If we only add the english translation like you did here should be good enough. The messages for other translations should update after the crowdin branch has been updated and merged.

mmioana:
Leaving this in DRAFT until I'll merge the strapi change

src/i18n/translations/en/translation.json
mmioana: Oh, so there is no need to add the empty translations for the other languages?

src/app/ui/mission/hooks.ts
```diff
@@ -0,0 +1,131 @@
...
+      //   if (taskType === TaskType.Zap && configType !== 'zap') {
+      //     setConfigType('zap');
+      //   } else if (taskType !== TaskType.Zap && configType === 'zap') {
+      //     setConfigType('default');
+      //   }
+
+      setCurrentActiveTask(task.uuid, taskType);
+
+      setCurrentTaskWidgetFormParams({
+        sourceChain: widgetFormParams.sourceChain ?? undefined,
```
oktapodia, line 77: ?? undefined should not be needed

src/types/strapi.ts
```diff
@@ -290,6 +290,8 @@ interface TaskVerification {
   description: string;
   CTALink: string;
   CTAText: string;
+  TaskType?: string;
+  TaskWidgetInformation?: any;
```
oktapodia: any 🙈

oktapodia (changes requested):
Couple of comments

src/app/ui/mission/hooks.ts
mmioana: I fixed the type, as all props coming in `TaskWidgetInformation` can be null, so I'd rather leave the `?? undefined` for these store props as we're going to use them in the lifi widget config. Wdyt?

Screenshot 2025-07-08 at 19 44 22

src/types/strapi.ts
mmioana: Forgot to actually use the correct type here 🫠 now it should be fixed

admsk2 (approved):

oktapodia (approved):

PR #2056, opened 2025-07-10
chore: Optimize workflow

oktapodia:


mmioana (approved):
Really nice work @oktapodia 🙌

PR #2060, opened 2025-07-11
feat: missions carousel

mmioana:
Resolves https://lifi.atlassian.net/browse/LF-14560

Todos:
- [x] Add strapi plugin for quests count so we query only for needed data - this PR will not work yet until https://github.com/jumperexchange/strapi-cms/pull/63 is deployed
- [x] Refactor a bit the components + placement
- [x] Disable animation when autoplay is stopped


dennyscode (approved):
Works great locally! Let´s see once https://github.com/jumperexchange/strapi-cms/pull/63 is merged

PR #2062, opened 2025-07-11
feat: Add custom route label

oktapodia:


src/components/Widgets/Widget.tsx
mmioana, line 1: I would have extracted this into a const, so it's easier to modify in the future, but nothing speaking against merging this 🙌

mmioana (approved):

PR #2064, opened 2025-07-14
feat: add campaign hero section

mmioana:
Contributes to https://lifi.atlassian.net/browse/LF-14576


dennyscode (approved):
Some feedback in dm, but looking great!

PR #2069, opened 2025-07-14
fix: mission pages layout

mmioana:
Based on latest [Figma](https://www.figma.com/design/MmAjz3cDZU2BbPDFeisAdi/Missions?node-id=212-51770&m=dev) 

- The page container should have `1080px` and a top margin of `48px`.
- The two column layout we need to have the main column of `640px`, while the widget part should have `408px`
- On **dark** mode we needed to interchange `surface1` with `surface2` as these colors are not properly aligned with the design system and changing them will impact the `Li.Fi` widget theme
- All content that needed a card wrapper will make use of `SectionCard` so we ensure we keep the style consistent
- Added the card wrapper for the `/mission` page for the missions section + the `Explore x missions`
- Expanded the mission cards on the mission/zap details section to full width as there was an issue when on tablet view
- And hope I didn't miss including anything in the description 😅


src/components/ZapWidget/DepositPoolCard/DepositPoolCard.tsx
```diff
@@ -104,76 +105,78 @@ export const DepositPoolCard: FC<DepositPoolCardProps> = ({
...
           />
+          <Typography
+            variant="bodyXLargeStrong"
+            sx={{ fontWeight: 700 }}
+          >{`${zapData?.meta.name} Pool`}</Typography>
+        </DepositPoolHeaderContainer>
+        <Grid container rowSpacing={3} columnSpacing={2}>
+          {apyValue && (
+            <DepositPoolCardItem
+              title={'Base APY'}
```
dennyscode: Those titles such as "Base APY", "TVL", "Pool token" are general terms that don´t need additional translation as done here, right?

dennyscode (commented):
Actually, I miss the times when we had german translations (which we ditched to avoid legal issues) but for some reason the German translation is being picked up and sneaking through in the Deposit section on the zap page. (We dont have it on other pages).
image

src/components/ZapWidget/DepositPoolCard/DepositPoolCardSkeleton.tsx
```diff
@@ -7,25 +8,27 @@ import Grid from '@mui/material/Grid';
...
-              sx={{ height: 32, width: '75%', marginTop: 0.5 }}
-            />
-          </Grid>
-        ))}
-      </Grid>
-    </DepositPoolCardContainer>
+    <SectionCardContainer>
+      <DepositPoolCardContainer>
+        <DepositPoolHeaderContainer>
+          <BaseSkeleton variant="circular" sx={{ height: 40, width: 40 }} />
```
dennyscode:
```suggestion
          <BaseSkeleton variant="circular" sx={{ height: 40, width: 40, flexShrink: 0 }} />
```

src/components/Cards/SectionCard/SectionCard.style.ts
```diff
@@ -4,8 +4,9 @@ import { styled } from '@mui/material/styles';
 export const SectionCardContainer = styled(Box)(({ theme }) => ({
   padding: theme.spacing(3),
   borderRadius: `${theme.shape.cardBorderRadius}px`,
-  boxShadow: theme.shadows[2],
-  backgroundColor: (theme.vars || theme).palette.surface2.main,
+  // elevation-4
```
dennyscode, line 7: Should we add to the ```shadows``` inside theme as the 4th elevation?

src/components/ZapWidget/DepositPoolCard/DepositPoolCard.tsx
dennyscode, line 1: Should this ```apyTooltip```also be added to translations?

dennyscode (commented):
Just as a note and to take on once we´ll find the time.

I think we have to revisit the translations as some of the contain grammatical issues or do not sound professional:
- "The displayed token is the token required to deposit into this market." (token is the token)
- "Additional APY you get from participating to this campaign inside Jumper. This APY will be paid in {{token}}." (“participating to” → “participating in”)

Generally, some of these are like a bullet-point where there is others that sentence-like, still all contain a dot at the end. We should find a general approach to phrase such.

src/components/Cards/SectionCard/SectionCard.style.ts
mmioana: We have an array in our theme defined after the first 3 definitions, so I wouldn't touch it for the moment tbh to avoid any unintended effect 😢

src/components/ZapWidget/DepositPoolCard/DepositPoolCard.tsx
mmioana: I was torn if to add it in now, or in another PR tbh

dennyscode (approved):

PR #2076, opened 2025-07-16
feat: add horizontal tabs

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-14557


src/components/HorizontalTabs/HorizontalTabs.stories.tsx
```diff
@@ -0,0 +1,195 @@
...
+    <HorizontalTabs
+      tabs={tabs}
+      value={value}
+      onChange={onChange}
+      onTabClick={onTabClick}
+      tabSx={{ width: '160px' }}
+      size={HorizontalTabSize.MD}
+    />
+  );
+};
```
mmioana: Do we still need this?

src/app/[lng]/test-tabs/page.tsx
Comment (mmioana) [inline src/app/[lng]/test-tabs/page.tsx:1]: Do we need this?

src/components/QuestPage/Banner/Banner.style.ts
```diff
@@ -150,24 +148,6 @@ export const BannerTitleTypography = styled(Typography)(({ theme }) => ({
...
-  '@keyframes spin': {
-    '0%': {
-      transform: 'rotate(360deg)',
-    },
-    '100%': {
-      transform: 'rotate(0deg)',
-    },
-  },
-}));
-
```
mmioana: Is this needed?

src/components/Navbar/NavbarButtons/NavbarButtons.tsx
mmioana, line 1: Please uncomment back these as we don't want to display the floating buttons yet

src/components/HorizontalTabs/HorizontalTabs.stories.tsx
mmioana, line 1: 😍

src/components/HorizontalTabs/HorizontalTabs.stories.tsx
mmioana: @dennyscode I would propose to use a render component that does what your current component does, but accepts the `args` as a parameter, so it's easier to manipulate the data passed to the component

```
const HorizontalTabsRenderer = (args: HorizontalTabsProps) => {
  const [value, setValue] = useState<string>(args.value ?? tabs[0].value);

  const handleChange = (event: React.SyntheticEvent, newValue: string) => {
    setValue(newValue);
    args.onChange?.(event, newValue);
  };

  const handleTabClick =
    (val: string) => (event: React.MouseEvent<HTMLDivElement>) => {
      event.preventDefault();
      setValue(val);
      args.onTabClick?.(val)?.(event);
    };

  return (
    <HorizontalTabs
      {...args}
      tabs={args.tabs ?? tabs}
      value={value}
      onChange={handleChange}
      onTabClick={handleTabClick}
    />
  );
};

export const InteractiveWithState: Story = {
  render: (args) => <HorizontalTabsRenderer {...args} />,
  args: {
    ...baseArgs,
    tabs,
    value: tabs[0].value,
  },
};
```

mmioana: And you'll need to export the `HorizontalTabsProps` from the `HorizontalTabs.tsx`***

src/components/HorizontalTabs/HorizontalTabs.stories.tsx
```diff
@@ -0,0 +1,195 @@
...
+      backgroundColor: theme.palette.alphaLight100.main,
+      ...(theme.applyStyles &&
+        theme.applyStyles('light', {
+          backgroundColor: (theme.vars || theme).palette.alphaDark100.main,
+        })),
+    }),
+  },
+};
+
+export const WithSubLinks: Story = {
```
mmioana: Is this the correct name for this story or is it even needed?

src/theme/theme.ts
```diff
@@ -650,6 +656,31 @@ export const themeCustomized: Omit<Theme, 'applyStyles'> & CssVarsTheme =
...
+            gap: theme.spacing(1),
+            minWidth: '56px',
+            minHeight: '40px',
+
+            '& .MuiTab-icon': {
+              marginBottom: 0,
+            },
+          }),
+        },
+      },
```
mmioana, line 683: Do we want to override this style globally?

src/components/Navbar/NavbarButtons/Links.tsx
```diff
@@ -1,29 +1,38 @@
...
+        // startAdornment: <PanoramaFishEyeIcon />,
+        // endAdornment: <PanoramaFishEyeIcon />,
       },
       {
-        href: AppPaths.Missions,
+        value: AppPaths.Missions,
         label: t('navbar.links.missions'),
+        // startAdornment: <PanoramaFishEyeIcon />,
+        // endAdornment: <PanoramaFishEyeIcon />,
       },
```
mmioana: Now that you have touched this component, can you please add the `subLinks: [AppPaths.Campaign]` this the `Missions` tab? 😬

mmioana (changes requested):
Hey @dennyscode really nice and clean work on this one! Left a few comments and once these are resolved, I'll happily approve ✨

src/components/HorizontalTabs/HorizontalTabs.stories.tsx
dennyscode: GREAT DEAL! Thanks, much better - still noobish with storybook but this makes sense and is more flexible now!!
https://github.com/jumperexchange/jumper-exchange/pull/2076/commits/149091ba2e814cf3425b1da99a2676d9f8d9027d

src/components/Navbar/NavbarButtons/Links.tsx
dennyscode: Ahhh, sure thing yes! Also got rid of the comments here

https://github.com/jumperexchange/jumper-exchange/pull/2076/commits/46093092f7d00b0f210dea1755fa3bbc198245b3

src/components/HorizontalTabs/HorizontalTabs.stories.tsx
dennyscode: THANKS! Naa, this seems to not be worth a story and can be considered junk, removed it:
https://github.com/jumperexchange/jumper-exchange/pull/2076/commits/c21fc39e0f4a109505d56c8b023bf7ebfa9ab315

src/theme/theme.ts
dennyscode: I kind of want to for now, yes. I believe it´s best to put stuff in the theme on that "lower level", if possible, but also I remember your points about the partner-themes (if we happen to revive them) - I don´t think it should be a problem either as we used to take our theming as a base and only applied partner-theme related changes on top.

It does not affect the VerticalTabs but maybe we can revisit once we´re also refactoring the component for VerticalTabs and adding those stories. Wdyt?

mmioana: Okay, we can leave it as is for now. However, changes like this at a global level should be made very rarely—and only if we're intentionally enforcing consistent styles across the entire app, regardless of styles introduced by any other theme. The reason I mention this is because removing it later could unintentionally break styling on other components throughout the app, making debugging and fixing much more difficult.

src/components/HorizontalTabs/TabsExample.tsx
mmioana, line 1: I'd still suggest to remove this as we're not using it, but feel free to decide how you want to proceed 🙌

mmioana (approved):
This PR is really nice, awesome work @dennyscode 💪

src/components/HorizontalTabs/HorizontalTab.tsx
```diff
@@ -0,0 +1,40 @@
...
+  label,
+  startAdornment,
+  endAdornment,
+  disabled = false,
+  size = HorizontalTabSize.MD,
+  value,
+  onClick,
+  sx,
+}: HorizontalTabProps) => {
+  const formattedLabel = (
```
oktapodia: Would be better to use forwardRef to avoid pasing all of those + Tab already natively support adornment by using `icon`

src/components/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export interface HorizontalTabsProps {
+  tabs: HorizontalTabItem[];
+  onChange: (event: React.SyntheticEvent, newValue: string) => void;
+  onTabClick: (
+    value: string,
+  ) => (event: React.MouseEvent<HTMLDivElement>) => void;
+  value?: string;
+  size?: HorizontalTabSize; //@todo: use enum instead
```
oktapodia: The comment might be deprecated

src/components/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export const HorizontalTabs = ({
+  tabs,
+  onChange,
+  onTabClick,
+  value,
+  size = HorizontalTabSize.LG,
+  sx,
+  tabSx,
```
oktapodia: Seems overkill to pass those sx through, is that really needed?

src/theme/theme.ts
```diff
@@ -80,8 +81,8 @@ declare module '@mui/material/styles' {
     azure: Pick<Color, 100 | 200 | 300 | 400 | 500 | 600 | 700 | 800 | 900>;
     scarlet: Pick<Color, 100 | 500>;
     orchid: Pick<Color, 100 | 200 | 300 | 400 | 500 | 600 | 700 | 800 | 900>;
-    lavenderLight: Pick<Color, 100 | 200 | 300 | 400>;
-    lavenderDark: Pick<Color, 100 | 200 | 300 | 400>;
+    lavenderLight: Pick<Color, 100 | 200 | 300 | 400> & { 0: string };
```
oktapodia: by logic, if it is 0 that's transparent or plain, do we really need this?

src/components/HorizontalTabs/HorizontalTabs.tsx
dennyscode: These optional sx´s props (general sx for container + tabSx for tab-items) offer flexibility to add custom styling, but yet with this <HorizontalTabs> component we still have a general and re-usable approach.
Do you see issues with that?

src/theme/theme.ts
dennyscode: Oh yes! Logically, you´re right but as we are trying to sync our theming with our Figma design system created by the design team, we did not ask such question and take it as granted. As long as we are aligned, I do not see any issues, but if you want - I can communicate with them. Wdyt?

src/components/HorizontalTabs/HorizontalTabs.tsx
dennyscode: https://github.com/jumperexchange/jumper-exchange/pull/2076/commits/1854f5f002796846f46a41400796b8f2d6a03a5a

src/components/HorizontalTabs/HorizontalTab.tsx
mmioana: @oktapodia the `Tab` only supports one icon at a time, which makes using something like `endAdornment` less clean as we'll still need to pass it in the way we do now. Thus I'd say to keep it simple and just use these props to be able to render the needed content. Maybe we could though switch to passing directly the children to the `StyledHorizontalTab`. Wdyt?

dennyscode: Unfortunately, the MUI Tab does not expect to take in children. 

As @mmioana stated - by default it only supports one icon at a time which brought us to unify the structure with a label of  `<startAdornment><label><endAdornment>` instead of choosing different ways to achieve the same goal for a more readible and cleaner approach.

oktapodia: understood

src/components/HorizontalTabs/HorizontalTabs.tsx
oktapodia: Yes, it is making things much harder to read + subject to errors/regression (this is the reason why we removed most of those implementation a year ago).

Benefits seems to be very limited as the tab itself should scale on the tabsContainer, if tab itself require different styling, we can edit it from the container sx (with class props) 

Especially as it is not even used :/ that's over-engineering

oktapodia (changes requested):
Couple of responses

src/components/HorizontalTabs/HorizontalTab.tsx
```diff
@@ -0,0 +1,42 @@
...
+  label,
+  startAdornment,
+  endAdornment,
+  disabled = false,
+  size = HorizontalTabSize.MD,
+  value,
+  onClick,
+  sx,
+}: HorizontalTabProps) => {
+  if (!startAdornment && !label && !endAdornment) {
```
oktapodia: Adding this block the fact that we can have start, label or endAdornment

dennyscode: The idea was to prevent any rendering of empty Tab in case none of these are given. As so often I may be on the wrong track, but maybe you are suggesting to add more checks for empty strings as well?

```
  const hasStartAdornment = startAdornment != null && startAdornment !== '';
  const hasLabel = label != null && label !== '';
  const hasEndAdornment = endAdornment != null && endAdornment !== '';
  
  if (!hasStartAdornment && !hasLabel && !hasEndAdornment) {
    return null;
  }
```

oktapodia: No, what I mean compare to what was there before [`476d1f7` (#2076)](https://github.com/jumperexchange/jumper-exchange/pull/2076/commits/476d1f7fd9d1a7f17fce37f4e3ac5538947ae880#diff-4b050ca98c5dffc3073310172bf6a02992177e86f9b03805a2b73644d42a965fR22) is that you could have startAdornment + label and it was working like a charm. (as an example)

But with this added check, it ONLY works if we have start/label AND end adornment

dennyscode: It is doing exactly what I was aiming for...
image
image

src/components/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -0,0 +1,51 @@
...
+  tabs,
+  onChange,
+  onTabClick,
+  value,
+  size = HorizontalTabSize.LG,
+  sx,
+}: HorizontalTabsProps) => {
+  return (
+    <HorizontalTabsContainer value={value} onChange={onChange} sx={sx}>
+      {tabs.map((tab) => (
```
mmioana: @dennyscode let's check here if the tab has a one of these properties `label`, `endAdornment` or `startAdornment` and if not use `null`, otherwise instantiate the `HorizontalTab`

dennyscode: Thanks, have moved the check to the `HorizontalTabs.tsx` now

src/components/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -0,0 +1,57 @@
...
+            label={tab.label}
+            size={size}
+            onClick={onTabClick(tab.value)}
+            disabled={tab.disabled}
+          />
+        );
+      })}
+    </HorizontalTabsContainer>
+  );
+};
```
mmioana: @dennyscode I think we can further simplify the implementation.
What I mean by simplifying is that
- you can control the selected value directly inside the `HorizontalTabs`
- then, if we don't pass the ref to the `HorizontalTab` custom component, given the MUI Tab's internals, it will not properly react to changes ( you can check your current implementation and see if `onChange` is actually called)
-  then you don't need to react to both the tab being clicked and the event being fired `onChange` as it will likely just duplicate the logic most probably in both event handlers
- also as a note when using an event handler it's best to try simplifying the layers of arguments as in not using `onTabClick: (
    value: string,
  ) => (event: React.MouseEvent<HTMLDivElement>) => void` but instead use a simpler approach `onTabClick (event: React.MouseEvent<HTMLDivElement>,  value: string) => void` - though now it can be safely removed

```
export const HorizontalTabs = ({
  tabs,
  onChange,
  value,
  size = HorizontalTabSize.LG,
  sx,
}: HorizontalTabsProps) => {
  const [internalValue, setInternalValue] = useState(value ?? tabs[0]?.value);

  useEffect(() => {
    if (value) {
      setInternalValue(value);
    }
  }, [value]);

  const handleChange = (event: React.SyntheticEvent, newValue: string) => {
    setInternalValue(newValue);
    onChange?.(event, newValue);
  };

  return (
    <HorizontalTabsContainer
      value={internalValue}
      onChange={handleChange}
      sx={sx}
    >
      {tabs.map((tab) => {
        if (!tab.label && !tab.startAdornment && !tab.endAdornment) {
          return null;
        }

        return (
          <HorizontalTabContainer
            disabled={tab.disabled}
            value={tab.value}
            key={tab.value}
            disableRipple
            sx={sx}
            label={
              <>
                {tab.startAdornment}
                {tab.label}
                {tab.endAdornment}
              </>
            }
            size={size}
          />
        );
      })}
    </HorizontalTabsContainer>
  );
};

```

You'll need to update the interfaces and usages accordingly though

src/components/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -0,0 +1,75 @@
...
+  return (
+    <HorizontalTabsContainer
+      value={internalValue}
+      onChange={handleChange}
+      sx={sx}
+    >
+      {tabs.map((tab) => {
+        if (!tab.label && !tab.startAdornment && !tab.endAdornment) {
+          return null;
+        }
```
mmioana: Could we actually first filter out all tabs that have none of these props, so we can then safely map them to the `HorizontalTabContainer` components, instead of returning `null`?

oktapodia (approved):

src/components/HorizontalTabs/HorizontalTabs.tsx
dennyscode: Wdyt? https://github.com/jumperexchange/jumper-exchange/pull/2076/commits/e832e41417ceb7c2fc07df42b3569ed92bd45332

mmioana (approved):
Seems we're ready to 🚢  nice work!

PR #2084, opened 2025-07-21
fix: add taskId and taskName to dependency list

mmioana:
Due to not having the dep list updated, the request was made w/ the wrong task id

Screenshot 2025-07-21 at 15 00 20
Screenshot 2025-07-21 at 15 00 51

Contributes to https://lifi.atlassian.net/browse/LF-14608


dennyscode (approved):

oktapodia (approved):
LGTM

PR #2090, opened 2025-07-22
fix: zap uplift init to context and sendCalls not being called

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-14736

- [x] refactored zap init hook into context
- [x] added logic to reinitialise meeClient and oNexus if there is a chain/walletAddress switch
- [x] add pending operations if clients are re-initialising
- [x] fixed issue w/ chain switching and `sendCalls` not being called
- [x] remove WalletProviderZap as the main layout already initialises the sdk config and adds the providers

**Testing**

1. Use a wallet standard account to make a transfer from the connected wallet's chain
2. Use a wallet standard account to make a transfer from a chain different than the wallet's current chain
3. Repeat steps 1-2 for a smart account though the transfer will not get through as we get this error 

Screenshot 2025-07-23 at 17 15 25

**Own testing**

https://github.com/user-attachments/assets/e7e2f14e-408a-432c-8df8-b4cee8b04619


src/providers/ZapInitProvider/ZapInitProvider.tsx
mmioana, line 1: Need to use the `EVMAddress` instead of `0x${string}`

src/providers/ZapInitProvider/ZapInitProvider.tsx
```diff
@@ -0,0 +1,801 @@
...
+      try {
+        resetInProgressRef.current = false;
+        initInProgressRef.current = true;
+        console.warn('Initializing oNexus with chains:', [
+          currentChain.id,
+          depositChain.id,
+        ]);
+        const oNexusInit = await toMultichainNexusAccount({
+          signer: walletClient,
+          accountAddress: walletClient.account.address,
```
mmioana: This might need to be removed as we're not dealing w/ embedded wallets

admsk2 (approved):

PR #2091, opened 2025-07-22
fix: campaign stats card colour and missing translations

mmioana:
Point 3 from https://lifi.atlassian.net/browse/LF-14672


src/hooks/campaigns/useCampaignDisplayData.ts
```diff
@@ -26,6 +26,10 @@ export const useCampaignDisplayData = (campaign: CampaignData) => {
         campaign.ProfileBannerCTA ||
         `${AppPaths.Campaign}/${campaign.Slug}` ||
         '',
+      statsCardVariant:
+        campaign.BenefitCardColorMode === BenefitCardColorMode.Dark
+          ? MissionHeroStatsCardVariant.Inverted
+          : MissionHeroStatsCardVariant.Default,
```
dennyscode, line 32: As in feature-card, I would suggest to be consistent and also keep the same naming as the field is defined in strapi to stick with "Light" and "Dark":

```suggestion
          ? MissionHeroStatsCardVariant.Dark
          : MissionHeroStatsCardVariant.Light,
```

dennyscode (approved):

src/hooks/campaigns/useCampaignDisplayData.ts
mmioana: Normally it would make better sense, but in the designs these variants are called as `default`/`inverted` so it would be more difficult to keep in mind which variant to use if you are using the designs as a reference, while at the same time using `default` and `inverted` in `Strapi` it would make it difficult for a content creator to have a mind map of what these mean. Wdyt?

PR #2093, opened 2025-07-23
feat: add feature flag to toggle new profile page

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-14681

> [!NOTE]
>
> If you don't see yet the new profile page with just the intro section, please run the app locally and set a different value at `line 31` in `src/app/[lng]/(infos)/profile/page.tsx`


src/components/ProfilePage/OldProfilePage.tsx
```diff
@@ -0,0 +1,93 @@
...
+export const OldProfilePage = ({ campaigns, quests }: ProfilePageProps) => {
+  const { walletAddress, isPublic } = useContext(ProfileContext);
+  const { isLoading, points, pdas } = useLoyaltyPass(walletAddress);
+  const { traits } = useTraits();
+
+  // const { isEnabled: isABTestEnabled } = useABTest({
+  //   feature: 'test_ab_1',
+  //   user: account?.address || '',
+  // });
+
```
laurentsenta, line 60:
```suggestion
```

dead code?

src/components/ProfilePage/OldProfilePage.tsx
```diff
@@ -0,0 +1,93 @@
...
+    userAddress: walletAddress,
+  });
+
+  const validBannerCampaigns = campaigns?.filter(isBannerCampaign) || [];
+
+  return (
+    <PageContainer className="profile-page">
+      {!isPublic && <MerklRewards />}
+      {/* <IntroSection /> */}
+      {/* {!isPublic && <RewardsSection />} */}
```
laurentsenta:
```suggestion
```

dead code?

src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx
```diff
@@ -61,14 +67,30 @@ export default async function Page({ params }: { params: Params }) {
...
-    <ProfilePage
-      quests={questsData.data}
+    <OldProfilePage
+      quests={questsExtended}
       campaigns={campaigns}
       walletAddress={sanitizedAddress}
-      isPublic
     />
   );
 }
```
Comment (laurentsenta) [inline src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx:96]: From what I understand we'll use the same init code every time we try to load the profile page component, could we wrap it, and expose a single `<ProfilePage/>` component so that the callers don't need to know about the ab test and the component internals?

Something like:

```ts
export const ProfilePage: React.FC.... () => {
    // load data

  if (isPageEnabled) {
    return (
      <Suspense fallback={<ProfilePageSkeleton />}>
        <ProfilePage2
          isPublic={true}
          campaigns={campaigns}
          quests={questsExtended}
        />
      </Suspense>
    );
  }

  return (
    <ProfilePage1
      isPublic={false}
      quests={questsExtended}
      campaigns={campaigns}
    />
  );
}
```

src/components/ProfilePage/OldProfilePage.tsx
laurentsenta, line 60:
```suggestion
```

src/components/ProfilePage/OldProfilePage.tsx
laurentsenta:
```suggestion
```

mmioana: This was left from my previous testing of updating those section, can easily be removed - though the OldProfilePage will be completely removed once we release

src/components/ProfilePage/OldProfilePage.tsx
mmioana: Will not do it, as I am not sure what that was and the old profile page component will be 🧹

src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx
Comment (mmioana) [reply src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx:96]: Would make sense if we actually want to do AB testing for this page, but it's more of intermediary solution until we release the new profile page, so don't see the benefit ftm

laurentsenta (approved):

PR #2097, opened 2025-07-23
fix: changes after design review

mmioana:
Figma comments 1-5, besides the colors and the tabbed navigation

https://www.figma.com/design/MmAjz3cDZU2BbPDFeisAdi/Missions?node-id=1-318&p=f&m=dev


src/hooks/tasksVerification/useGetVerifiedTasks.ts
```diff
@@ -29,11 +31,26 @@ export async function getVerifiedTasksQuery({
...
+  } catch (error) {
+    console.error('Failed to refetch with cache-busting:', error);
+  }
+};
+
 export const useGetVerifiedTasks = (address?: string) => {
   return useQuery({
     queryKey: ['task_verification', address!],
-    queryFn: getVerifiedTasksQuery,
+    queryFn: () => getVerifiedTasksQuery(address!),
```
laurentsenta, line 52: Since we're using `!` on both the key and fn, it looks like we always assume it's defined. What if we removed the optional on the function def, and let callers make that assumption?

src/hooks/quests/useFormatDisplayRewardsData.ts
```diff
@@ -2,30 +2,35 @@ import { CustomInformation, RewardGroup } from 'src/types/loyaltyPass';
...
+        chains: customInformation?.['chains'],
+      };
+    }, [JSON.stringify(customInformation ?? {})]);
 
-  //   const chains = customInformation?.['chains'] ?? [];
+  const chainIds = (chains ?? [])
+    .map((chain) => chain.chainId)
+    .filter((chainId) => chainId !== undefined);
 
   // @TODO maybe need to pass the chainIds
```
laurentsenta:
```suggestion
```

laurentsenta (approved):
Left a couple of non-blocking comments

src/hooks/tasksVerification/useGetVerifiedTasks.ts
mmioana: Actually, the address can be undefined, but in that case, the query isn't enabled, so when the query function is triggered, we can safely assume that the address is defined. Given this, I'd leave the hook param def as optional

PR #2100, opened 2025-07-25
feat: connect perks to the profile page

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-14684

Screenshot 2025-07-25 at 15 50 43

> [!NOTE]
>
> If you don't see yet the new profile page with just the intro section, please run the app locally and set a different value at `line 34` in `src/app/[lng]/(infos)/profile/page.tsx`

### DIFF
--- a/src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx
+++ b/src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx
@@ -4,6 +4,7 @@ import type { Metadata } from 'next';
 import { notFound } from 'next/navigation';
 import { Suspense } from 'react';
 import { getFeatureFlag } from 'src/app/lib/getFeatureFlag';
+import { getPerks } from 'src/app/lib/getPerks';
 import { getProfileBannerCampaigns } from 'src/app/lib/getProfileBannerCampaigns';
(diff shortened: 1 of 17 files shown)


src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx
```diff
@@ -67,26 +68,35 @@ export default async function Page({ params }: { params: Params }) {
...
+    getProfileBannerCampaigns(),
+    getQuestsWithNoCampaignAttached(),
+    getPerks(),
+    getFeatureFlag(
+      GlobalFeatureFlags.ProfilePage,
+      // Placeholder distinctId required by the API call.
+      // This global feature flag is not tied to any specific user.
+      'distinct-id',
+    ),
+  ]);
```
Comment (oktapodia) [inline src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx:81]: Not really fan of this duplication :s

oktapodia (approved):
LGTM

src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx
Comment (mmioana) [reply src/app/[lng]/(infos)/profile/[walletAddress]/page.tsx:81]: Yes, now that you're mentioning it, it actually makes sense to extract a separate handler for it, though my plan would be to remove the feature flags before the release

PR #2101, opened 2025-07-25
feat: add achievement cards list to profile

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-14683

> [!IMPORTANT]
> Ping me for a wallet address having achievements

> [!NOTE]
>
> If you don't see yet the new profile page with just the intro section, please run the app locally and set a different value at `line 34` in `src/app/[lng]/(infos)/profile/page.tsx`

### DIFF
--- a/src/components/ProfilePage/components/AchievementsList/AchievementsList.tsx
+++ b/src/components/ProfilePage/components/AchievementsList/AchievementsList.tsx
@@ -1,0 +1,70 @@
+import { useContext, useMemo, useEffect, useCallback, useState } from 'react';
+import { InfiniteScroll } from 'src/components/InfiniteScroll/InfiniteScroll';
+import { useLoyaltyPass } from 'src/hooks/useLoyaltyPass';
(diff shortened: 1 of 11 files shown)


src/components/ProfilePage/components/AchievementsList/AchievementsList.tsx
```diff
@@ -0,0 +1,70 @@
...
+  // Simulate pagination by truncating the existing data
+  // @TODO: remove this once we have a proper pagination solution
+  const [displayedPdas, setDisplayedPdas] = useState<PDA[]>([]);
+  const [currentPage, setCurrentPage] = useState(1);
+  const [isFetchingNextPage, setIsFetchingNextPage] = useState(
+    isWalletLoading || isLoading,
+  );
+  const pageSize = 10;
+
+  const totalPages = useMemo(() => {
```
oktapodia: I think that we have a reusable pagination component already which manage those, otherwise it would be good to extract it as we have pagination in many pages

oktapodia (approved):
LGTM

PR #2103, opened 2025-07-25
feat: add achievements perks section to profile page

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-14682

This adds the card with tab navigation on the profile page

> [!NOTE]
>
> If you don't see yet the new profile page with just the intro section, please run the app locally and set a different value at `line 31` in `src/app/[lng]/(infos)/profile/page.tsx`


oktapodia (approved):
LGTM

PR #2104, opened 2025-07-25
fix: resolve mode for generated img

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-14776

### DIFF
--- a/src/utils/image-generation/helpers.ts
+++ b/src/utils/image-generation/helpers.ts
@@ -24,7 +24,7 @@ export const getWidth = (type?: string, fullWidth?: boolean) => {
   }
 };
 
-export const getResolvedMode = (mode: Appearance | undefined): ImageTheme => {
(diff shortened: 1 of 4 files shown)


src/utils/image-generation/helpers.ts
```diff
@@ -1,3 +1,6 @@
+import { Appearance } from '@lifi/widget';
```
dennyscode, line 1: Not sure if this is best practise to be used from `@lifi/widget` or if we should create another dedicated type that includes `'light' | 'dark' | 'system'` within our repo.

oktapodia (approved):
LGTM

src/utils/image-generation/helpers.ts
mmioana: If these are the values we also use, should be fine! If in the future we want to extend this type we can create a new one `type OwnAppearance = Appearance & { <prop-key>: <prop-value>... }`

mmioana (approved):
🚢

PR #2105, opened 2025-07-25
fix: campaign design feedback

mmioana:
Feedback from https://www.figma.com/design/MmAjz3cDZU2BbPDFeisAdi/Missions?node-id=1104-120425&m=dev

And realised I actually updated wrongly the `Badge` styles previously, which wasn't obvious as I used the another size variant

### DIFF
--- a/src/app/ui/mission/MissionDetails.tsx
+++ b/src/app/ui/mission/MissionDetails.tsx
@@ -58,7 +58,7 @@ export const MissionDetails: FC<MissionDetailsProps> = ({ mission, tasks }) => {
       <Badge
         label={status}
         variant={BadgeVariant.Secondary}
-        size={BadgeSize.MD}
+        size={BadgeSize.LG}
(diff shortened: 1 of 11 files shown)


laurentsenta (approved):

mmioana:
Hey @laurentsenta thanks so much for your review 🙏 

I've added two more commits after discussing w/ Tim last evening. When you have a moment, I'd appreciate it if you could take another look 😬 

I've simplified the progress animation, as the swiper timer and the CSS animation were getting out of sync during pause/resume. The previous logic was a workaround for an issue in Firefox, where swiper events weren’t reliably triggered, which caused the state flag not to update correctly. Now that we're using the timeLeft approach, the animation should be properly in sync.

I also moved the borderRadius and boxShadow styles to apply them directly to the `swiper` container

admsk2 (approved):

PR #2112, opened 2025-07-28
chore: extract instructions building

mmioana:


src/providers/ZapInitProvider/Zapper.ts
```diff
@@ -17,15 +20,251 @@ export interface SendCallsExtraParams {
...
+        depositAddress,
+        runtimeERC20BalanceOf({
+          targetAddress: oNexus.addressOn(depositChainId, true) as EVMAddress,
+          tokenAddress: depositToken,
+          constraints,
+        }),
+      ],
+    });
+  }
+}
```
laurentsenta: So this looks like a function object, that's a pattern you would find in programming language without first-class functions but typescript has them/ Since we're not relying on implementation inheritance, what if we defined these as functions?

```
type ZapInstruction = (...) => ...

export deposit: ZapInstruction = (params) => {
   ....
}

export hyperwaveDeposit: ....
```

src/providers/ZapInitProvider/Zapper.ts
```diff
@@ -17,15 +20,251 @@ export interface SendCallsExtraParams {
   address: string | undefined;
 }
 
+export interface InstructionCommand {
+  buildInstruction(
+    oNexus: MultichainSmartAccount,
+    sendCallsExtraParams: SendCallsExtraParams,
+    zapper: ZapperStrategy,
+  ): Promise<any | null>;
```
laurentsenta: Is there anyway we could type this?

admsk2 (approved):

src/providers/ZapInitProvider/Zapper.ts
admsk2: I think yes, i will do it later

src/providers/ZapInitProvider/Zapper.ts
admsk2: Yeah, i like that too

admsk2 (approved):

PR #2115, opened 2025-07-29
feat: replace mission completed with task completed

mmioana:
Issue https://lifi.atlassian.net/browse/LF-14785

For a verified task check the `Task completed` screen is in place

### DIFF
--- a/src/hooks/tasksVerification/useEnhancedTasks.ts
+++ b/src/hooks/tasksVerification/useEnhancedTasks.ts
@@ -26,13 +26,20 @@ export const useEnhancedTasks = (
   }, [tasks, verifiedTaskIds]);
 
   const {
-    isMissionCompleted,
-    setIsMissionCompleted,
     setCurrentActiveTask,
+    setIsCurrentActiveTaskCompleted,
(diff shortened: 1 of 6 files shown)


src/hooks/tasksVerification/useEnhancedTasks.ts
```diff
@@ -26,14 +26,21 @@ export const useEnhancedTasks = (
...
     setCurrentTaskInstructionParams,
     currentActiveTaskId,
+    isCurrentActiveTaskCompleted,
   } = useMissionStore();
 
+  const checkIsTaskVerified = useCallback(
+    (task: TaskVerificationWithApy) => {
+      return task && task.hasTask && verifiedTaskIds.has(task.uuid);
+    },
+    [JSON.stringify(verifiedTasks)],
```
admsk2, line 41: is that OK to stringify on the fly?

mmioana: It's an optimisation, as it prevents unnecessary memo recalculations if `verifiedTasks`  has the same data inside.

admsk2 (approved):

PR #2122, opened 2025-07-31
feat(zaps): bypass widget calls on same token tx

laurentsenta:
Closes https://lifi.atlassian.net/browse/LF-14731

Ignore widget calls data when we're doing a same token deposit.

⚠️ Note this PR is targeting the LF-14699 branch.


mmioana (approved):
Cool cool @laurentsenta  🙌 

Will be nice to move `isSameToken` in a utils in the future

admsk2 (approved):

PR #2124, opened 2025-07-31
fix: add retrial logic for biconomy clients init

mmioana:
The biconomy clients might fail to initialise, so added retrial logic.

Attempts to solve the issue @chybisov encountered. 

image (2)
image (3)

> [!NOTE]
> We might need to track pending route executions by chain and initialise separate client instances for each chain, then route RPC method calls to the appropriate chain-specific clients.
> Multiple route executions can run simultaneously on different chains, but global clients only support one chain at a time. When background routes execute, they use the wrong chain's clients, causing the wallet methods to hang.


laurentsenta (approved):
Should we create a ticket to catch route changes? That sounds pretty important.

oktapodia (approved):
LGTM

mmioana:
@laurentsenta 100% yes, will add one now

PR #2125, opened 2025-07-31
fix: light grey 300 override for widget

mmioana:


src/theme/theme.ts
mmioana, line 1: As we already have the `surface4` colors in the light/dark palette, decided to add them to our theme config

oktapodia (approved):
LGTM

oktapodia (approved):

PR #2126, opened 2025-07-31
chore: update profile missing achievements component

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-14878

> [!NOTE]
> Took the word for it that the perks would always have at least one item. 
> Otherwise we will need to update the image url - luckily only an image needs to be loaded and in the `PerksList` component we should point the `imageUrl` for `NoDataPlaceholder`

**Testing steps**

1. Just connect a fresh new wallet and head to `/profile`
2. Select the `Achievements` tab
3. You should see the empty state card

Screenshot 2025-07-31 at 17 33 34

4. Switch to a mobile view and ensure it's responsive 

Screenshot 2025-07-31 at 17 33 54


src/components/ProfilePage/components/NoDataPlaceholder/NoDataPlaceholder.style.ts
```diff
@@ -1,17 +1,41 @@
...
+export const NoDataPlaceholderCard = styled(Box)(({ theme }) => ({
+  height: '100%',
+  display: 'flex',
+  flexDirection: 'column',
+  alignItems: 'center',
+  justifyContent: 'center',
+  gap: theme.spacing(4),
+  padding: theme.spacing(3),
+  borderRadius: theme.shape.cardBorderRadius,
+  backgroundColor: theme.palette.surface1.main,
```
mmioana: need to use `theme.vars || theme`

laurentsenta (approved):

PR #2129, opened 2025-08-04
fix: disable redirect for locked perks

mmioana:
This PR handles enabling the redirect for unlocked perks and disabling card interactivity for locked ones.

**Testing steps**
1. Use a fresh wallet to and navigate to the perks page 
2. The current perks should be locked and not clickable
3. Either switch to a wallet having a high enough level (min. 12) or create a perk inside strapi with a small unlock level (as a DRAFT)
4. Check perks with a level < current level are unlocked and clickable


callMeTheQA:
hi @mmioana  it looks good on your PR . 

Example video:

https://github.com/user-attachments/assets/166efc86-2f4d-43c9-b01f-a35828fcf538

callMeTheQA (approved):

PR #2130, opened 2025-08-04
feat: initialise biconomy clients for unfinalised route executions

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-14889

- [x] Use Zustand store for Biconomy clients and pending operations
- [x] Add persistence for pending operations with serialization/deserialization of initial args
- [x] Track current route on start and on process state STARTED updates (the latter is needed for history execution)
- [x] Execute pending operations only if they match the current route ID
- [x] Queue only operations requiring Biconomy clients; getCapabilities should run immediately
- [x] Move wallet client methods to plain JS functions (all data via params)
- [x] Create hook to initialize wallet client and dependent Biconomy clients, using current wallet as fallback
- [x] Remove `isInitialisedForCurrentChain` check for pending ops; only the pending operations execution effect should have it as a dependency to react to clients being initialised

> [!NOTE]
> Pending operations persistence might not actually be needed, given the wallet client methods will be called again once a route is restarted after page refresh
> This needs more testing

### 1. Deposit via Biconomy
- Start a deposit and confirm it goes through Biconomy and appears in history. 

### 2. Create Multiple Pending Operations
- Start another deposit without finalizing.
- Start a deposit on a different chain without finalizing.
- ~~Check both appear as pending in the store/history.~~ Unless there was an issue with the biconomy clients initialisation or the current route is not available, only the `waitForCalls`/`getCalls` will be stored

### 3. Switch Between Pending Routes
- Switch between the two unfinalized routes in history.
- Confirm the correct route state is resumed each time. 

### 4. Queue & Execution Verification
- Ensure only operations needing Biconomy are queued.
- Confirm getCapabilities executes immediately, not delayed. 

### 5. Refresh / Persistence Check
- Refresh the page with pending operations.
- Verify they are restored (if persistence is enabled) or re-executed via wallet client methods. 

### 6. Finalize and Clean Up
- Finalize a pending operation and confirm it is removed from the queue/history.
- Ensure remaining pending operations continue to work correctly. 

### 7. Edge Case: Rapid Switching
- Rapidly switch between pending routes.
- Confirm the correct route is always active and no wrong execution occurs.


admsk2 (approved):

src/providers/ZapInitProvider/ZapInitProvider.tsx
```diff
@@ -127,159 +125,218 @@ export const ZapInitProvider: FC<ZapInitProviderProps> = ({
...
+            args,
+            resolve,
+            reject,
+          );
         });
       }
 
-      return operation(meeClient, oNexus, sendCallsExtraParams);
+      return operation(
+        args as any,
```
laurentsenta: So close! Something is up with typescript and react, it should be able to resolve that type correctly, same for the `const operation ... as X | undefined` above.

mmioana: I know! It was heartbreaking, though I think the fact that one of the methods having the params set as `never` is breaking it

src/providers/ZapInitProvider/types.ts
```diff
@@ -74,25 +83,36 @@ export interface CallsStatusResponse {
...
+  >;
+  [WalletMethods.waitForCallsStatus]: WalletMethodDefinition<
+    WalletWaitForCallsStatusArgs,
+    CallsStatusResponse
+  >;
+  [WalletMethods.sendCalls]: WalletMethodDefinition<
+    WalletSendCallsArgs,
+    SendCallsResponse
+  >;
 }
```
laurentsenta: Ultra badass, it's almost completely typesafe and generic at the same time 🔥 

3 nits and a proposition:
- arg & responses names are inconsistent, it'd be easier to follow through with consistent naming
- WalletMethodArgs.method could be typed
- The type `WalletMethod(s)` should be singular (it contains a single method, like a `number` or `string`)
- You could build your type the other way around, this way you have one place to define the entire object, it trickles down to WalletMethod, and it's more readable than these `[MyEnum.ItsValue]:`.

```ts
export interface WalletMethodArgs {
  method: WalletMethod;
  params?: unknown[];
}

export interface WalletMethodsRef {
  wallet_getCapabilities: WalletMethodDefinition<
    GetCapabilitiesArgs,
    GetCapabilitiesResponse
  >;
  wallet_getCallsStatus: WalletMethodDefinition<
    GetCallsStatusArgs,
    GetCallsStatusResponse
  >;
  wallet_waitForCallsStatus: WalletMethodDefinition<
    WaitCallsStatusArgs,
    WaitCallsStatusResponse
  >;
  wallet_sendCalls: WalletMethodDefinition<
    SendCallsArgs,
    SendCallsResponse
  >;
};

export type WalletMethod = keyof WalletMethodsRef
```

src/providers/ZapInitProvider/ZapInitProvider.tsx
laurentsenta: There's a lot already, and it's almost type safe, so we might not revisit that code, but if you ever fix it, let me know :D

src/stores/zapPendingOperations/ZapPendingOperationsStore.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    for (const [key, value] of Object.entries(obj)) {
+      result[key] = deserializeWithTypes(value);
+    }
+
+    return result as T;
+  }
+
+  return obj as T;
+};
```
laurentsenta: Serialization / deserialization is alway a bit tricky to get right (dealing with collisions, weird inputs, etc), and it's always too late when you find out about a bug.

Since they are pure functions, could we move them to their module and unit-test them a bit? Or would it make sense to use a library like [superjson](https://github.com/flightcontrolhq/superjson)?

src/stores/zapPendingOperations/ZapPendingOperationsStore.ts
```diff
@@ -0,0 +1,199 @@
...
+      },
+
+      clearAll: () => {
+        set({ pendingOperations: {} });
+        get().promiseResolvers.clear();
+      },
+    }),
+    {
+      name: 'zap-pending-operations-storage',
+      version: 1,
```
laurentsenta: Do you know what type of changes would be breaking and require us to bump the version and forget about pending operation?

laurentsenta (approved):
code looks great, super precise typing,
I couldn't dig to deep into concurrency issues, I need more practice with the widget lifecycle 🙏

src/stores/zapPendingOperations/ZapPendingOperationsStore.ts
mmioana: That is an amazing point @laurentsenta let's go with `superjson` ❤️ though would love to do it in a separate PR

src/stores/zapPendingOperations/ZapPendingOperationsStore.ts
mmioana: Don't have one in mind, so I guess we can remove the `version` actually

src/providers/ZapInitProvider/ZapInitProvider.tsx
mmioana: Initially I was sure the `params: never` were triggering the issue, but it was actually caused by how ts handles generic function parameter types

This solved it ✅ 

```
     operation as WalletMethodDefinition<
          WalletMethodArgsType<T>,
          Awaited<WalletMethodReturnType<T>>
        >
```

PR #2131, opened 2025-08-04
feat: added test for perks / achievements and some refactoring [LF-14925]

callMeTheQA:
https://lifi.atlassian.net/browse/LF-14925


src/components/HorizontalTabs/HorizontalTabs.style.ts
mmioana, line 1: I would remove `id` from here as it's a standard HTML prop for an element, so no need to skip forwarding it

callMeTheQA: Thanks @mmioana  , removed !

PR #2135, opened 2025-08-06
LF-14699: hyperwave mint fix

laurentsenta:
- [x] drop special case for minting, it's hardcoded now
- [x] fix from amount
- [x] add slippage to deal with latency during tx execution

## Success on base usdc -> hyperwave

Success without slippage:
https://meescan.biconomy.io/details/0x3fc6af0e33c2b02dc700390e8c67be4ff42aacd298a2776584594991766a8a1b

Success with slippage (first version):
https://meescan.biconomy.io/details/0xd70174d5e3e187cecff7aa1f6e330fde8e57f8299f7e7bf45b5185df0f383a30

image

### DIFF
--- a/src/providers/ZapInitProvider/ModularZaps/HyperwaveZap.ts
+++ b/src/providers/ZapInitProvider/ModularZaps/HyperwaveZap.ts
@@ -29,7 +29,7 @@ const computeHyperwaveMinimumMint = async (
 
   const decimals = market.depositToken.decimals;
   const token = market.depositToken.address;
-  const amount = BigInt(zapper.currentRoute.fromAmount);
+  const amount = BigInt(zapper.currentRoute.toAmount);
 
   const getRateInQuoteSafe = zapper.zapData.abi.getRateInQuoteSafe;
 
--- a/src/providers/ZapInitProvider/ModularZaps/HyperwaveZap.ts
+++ b/src/providers/ZapInitProvider/ModularZaps/HyperwaveZap.ts
@@ -17,6 +17,13 @@ import { buildContractComposable } from '../utils';
 import { approve, transfer } from './DefaultZap';
 import { ZapDefinition, ZapExecutionContext, ZapInstruction } from './base';
 
+const MINT_SLIPPAGE_PERCENT = 2;
+
+const withSlippage = (amount: bigint, slippagePercent: number) => {
+  const slippageAmount = (amount * BigInt(slippagePercent)) / 100n;
+  return amount - slippageAmount;
+};
+
 const computeHyperwaveMinimumMint = async (
   zapper: ZapExecutionContext,
 ): Promise<bigint> => {
@@ -74,7 +81,11 @@ export const hyperwaveDeposit: ZapInstruction = async (
     greaterThanOrEqualTo(parseUnits('0.1', depositTokenDecimals)),
   ];
 
-  const minimumMint: bigint = await computeHyperwaveMinimumMint(context);
+  const minimumMint: bigint = withSlippage(
+    await computeHyperwaveMinimumMint(context),
+    MINT_SLIPPAGE_PERCENT,
+  );
+
   const depositInputs = integrationData.abi.deposit.inputs;
 
   const depositArgs = depositInputs.map((input: AbiParameter) => {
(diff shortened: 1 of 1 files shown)


src/providers/ZapInitProvider/ModularZaps/HyperwaveZap.ts
```diff
@@ -74,13 +81,15 @@ export const hyperwaveDeposit: ZapInstruction = async (
     greaterThanOrEqualTo(parseUnits('0.1', depositTokenDecimals)),
   ];
 
-  let minimumMint: bigint | null = await computeHyperwaveMinimumMint(context);
+  const minimumMint: bigint = withSlippage(
+    await computeHyperwaveMinimumMint(context),
+    MINT_SLIPPAGE_PERCENT,
+  );
+
```
laurentsenta, line 88: Please double check, it made sense to assume there's a bit of slippage due to the multiple txs run before we reach the deposit, but I'm not 100% sure since we encountered a different issue.

laurentsenta:
Note the flow was successful but I got dead links in the UI:

Link 1 (bridge tx confirmed) - error 500
https://basescan.org/tx/biconomy:0xd70174d5e3e187cecff7aa1f6e330fde8e57f8299f7e7bf45b5185df0f383a30

Link 2 (deposit completed) -  tx hash not found
https://hyperevmscan.io/tx/0xd70174d5e3e187cecff7aa1f6e330fde8e57f8299f7e7bf45b5185df0f383a30

Link 3 (send): 404
https://meescan.biconomy.io/address/0xb9bc23Fb2E50536CeD652297e2506a2E66f64340

src/providers/ZapInitProvider/ModularZaps/HyperwaveZap.ts
```diff
@@ -74,13 +81,15 @@ export const hyperwaveDeposit: ZapInstruction = async (
...
   const depositInputs = integrationData.abi.deposit.inputs;
+
   const depositArgs = depositInputs.map((input: AbiParameter) => {
     if (input.name === 'minimumMint') {
-      if (minimumMint === null || minimumMint <= 0) {
-        throw new Error('Minimum mint is not set');
-      }
       return minimumMint;
```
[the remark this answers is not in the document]
laurentsenta: We're using solely the hyperwave's API in that code path, no reason to assume negative or null mint anymore.

admsk2 (approved):

src/providers/ZapInitProvider/ModularZaps/HyperwaveZap.ts
mmioana: Does this mean we need to mint an amount taking into account the fees taken along the way?

laurentsenta: @admsk2 to correct me if I'm wrong,

We already take fees into account:

- With https://github.com/jumperexchange/jumper-exchange/pull/2135/commits/6f1fab3a392d5617bc25b4b2565a0785487d7d95: we'll use the correct estimate that takes into account the conversion's fee, so if you go from 5 USDC on base to 4.9USDT0 on hyperevm, we'll use 4.9.
- And with [that line](https://github.com/jumperexchange/jumper-exchange/blob/34bf9f617aa662fc7932588a7aea506ae6d26ac7/src/providers/ZapInitProvider/ZapInitProvider.tsx#L630) we compute the fees taken by biconomy (the quote returns a tokenAmount), that we add to the transaction approved by the user on metamask (so you'll approve spending 5.2 USDC for example).


The problem I addressed here is that we want to pass a minimum mint, which `getRateInQuoteSafe` gives us, but it might slip between the moment we compute the tx and the moment it gets executed by biconomy.

admsk2: so at this point there might be some slippage which is coming from the bridge/swap transaction, this is not a fee, but a slippage itself, which by default is set to 0.5% on jumper side, so setting extra slippage on minimum mint like 1-2% is totally fine and it includes 0.5% of previous transaction slippage (2% - 0.5% = 1.5% real slippage); This is minumum mint, meaning it will fail if outcome is less then that, but usually you get more

laurentsenta: Nice I haven't though about taking into consideration the previous tx's slippage. I'll add back the commit, thanks for the confirmation,

Won't do here, but I guess the perfect version is to use the Widget's configured slippage.

PR #2152, opened 2025-08-07
fix: chains should not be sorted alphabetically

chybisov:
Issue:

https://github.com/user-attachments/assets/22af3118-4613-44fa-bc92-707f923eaac3


oktapodia (approved):

mmioana (approved):
Good catch 💯

PR #2170, opened 2025-08-13
feat: wallet sorting order for MetaMask and Phantom

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-14470

## Testing steps

> [!IMPORTANT]
> We need to ensure for these 2 wallet we have this order
> MetaMask: EVM → UTXO → SVM → MVM - though from my research only **EVM** and **SVM** are available for MetaMask
> Phantom: SVM → EVM → UTXO → MVM

1. Try connecting from the navbar `Connect` button and confirm the ecosystem order for these 2 wallets is as described
2. Try connecting from the wallet management drawer 
2.1. Click on the connected wallet in the navbar and the wallet management drawer shows up
2.2. Click on the `Connect another wallet`
2.3. Confirm that once a wallet is connected on an ecosystem, that ecosystem shows up first, but the unconnected ecosystem are still sorted in the defined order
3. Try connecting through the widget and confirm the ecosystem order
3.1. On the `/` page with an unconnected wallet, press the `Connect wallet` button in the widget

https://github.com/user-attachments/assets/c6f6608e-add3-4013-b993-f08a6d0291cb


src/providers/WalletProvider/constants.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const walletEcosystemsOrder = Object.fromEntries(
+  Object.entries(walletPriorities).map(([wallet, priorities]) => [
+    wallet,
+    [
+      ...priorities,
+      ...allChainTypes.filter((type) => !priorities.includes(type)),
+    ],
+  ]),
+);
```
laurentsenta, line 15: I'm not saying we have to, but for what it's worth, that's one of the cases, where lodash makes the code super readable, it reads like english almost:

```ts
import mapValues from 'lodash/mapValues';
import uniq from 'lodash/uniq';

...

export const walletEcosystemsOrder = mapValues(
   walletPriorities,
   (priorities) => uniq([...priorities, ...allChainTypes])
)
```

laurentsenta (approved):

src/providers/WalletProvider/constants.ts
mmioana: Oh definitely a much nicer and cleaner solution, so pushing it rn 🚀 thank you @laurentsenta!

PR #2183, opened 2025-08-14
fix: try using the active connector

mmioana:


mmioana:
No longer needed, as getting the active connector is treated differently

PR #2184, opened 2025-08-14
feat: add vaults override with Gauntlet

laurentsenta:
Fixes https://lifi.atlassian.net/browse/LF-15086

Adds Katana's target token:

💣  Do not merge:

**Problem 1: The token will show up in other parts of the UI, like at as a source, we don't want that**

**Problem 2: We need to decide if we want to hardcode these now and release, or implement it "correctly"**

The "right" way would pull to the data from Strapi or the Backend, maybe through the `get-zaps-data` endpoint,
and update the Widget Config dynamically.

https://github.com/jumperexchange/jumper-backend/blob/15d6bf30784dfd0659c034e98f2aa634ddac964f/src/common/contracts/contracts.json#L208-L212

### DIFF
--- a/src/components/Widgets/variants/widgetConfig/hooks.ts
+++ b/src/components/Widgets/variants/widgetConfig/hooks.ts
@@ -1,31 +1,31 @@
-import { useMemo } from 'react';
 import { WidgetConfig } from '@lifi/widget';
 import merge from 'lodash/merge';
-
-import { ConfigContext } from './types';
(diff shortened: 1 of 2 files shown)


laurentsenta:
I might have found a better way to do it (single source of truth): https://github.com/jumperexchange/jumper-backend/pull/416

laurentsenta:
Superseded by https://github.com/jumperexchange/jumper-exchange/pull/2203

PR #2193, opened 2025-08-15
fix: Increasing timeout + more logs

oktapodia:


mmioana (approved):
I believe it's still going to be overridden like this, unless the timeout is undefined, but let's go 🚀

PR #2198, opened 2025-08-19
fix: error shown due to rpc rate limiting

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-15140

## Testing steps
1. Go to `/zap/morpho-katana-usdc`
2. Select `Base` as the source chain
3. Select any token and insert an amount > 5$
4. Deposit
5. Sign the tx
6. Ensure you don't get this error message

2370405a-1881-45d9-8718-d49a079017df

> [!IMPORTANT]
> Bonus testing points:
> Repeatedly refresh the page while your wallet is connected to Base.
> If an RPC URL causes an error during the Biconomy initialization, the next initialization attempt should automatically switch to a different RPC URL, since we shuffle them on each retry.

### DIFF
--- a/src/const/rpcList.ts
+++ b/src/const/rpcList.ts
@@ -30,3 +30,8 @@ export const publicRPCList = {
   '42161': shuffleArray(arbRPCList),
   '8453': shuffleArray(basRPCList),
 };
+
+export const mergedRPCList = {
+  ...JSON.parse(config.NEXT_PUBLIC_CUSTOM_RPCS ?? {}),
+  ...publicRPCList,
+};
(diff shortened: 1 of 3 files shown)


src/const/rpcList.ts
```diff
@@ -18,15 +18,22 @@ const arbRPCList = [
...
 const basRPCList = [
+  'https://base.llamarpc.com',
   'https://mainnet.base.org/',
-  'https://base.meowrpc.com',
+  // 'https://base.meowrpc.com',
   // 'https://base.drpc.org',
   `https://lb.drpc.org/ogrpc?network=base&dkey=${config.NEXT_PUBLIC_DKEY}`,
   'https://base-pokt.nodies.app',
+  'https://base-rpc.publicnode.com',
 ];
```
mmioana, line 28: Used https://chainlist.org/chain/8453 as reference for the rpc urls update

admsk2 (approved):

PR #2212, opened 2025-08-25
feat: improve message-based flows

chybisov:
Related to [LF-15121](https://lifi.atlassian.net/browse/LF-15121) [LF-12715](https://lifi.atlassian.net/browse/LF-12715)

[LF-15121]: https://lifi.atlassian.net/browse/LF-15121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-12715]: https://lifi.atlassian.net/browse/LF-12715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


callMeTheQA:
hey @chybisov  seems that your PR breaks Playwright tests https://github.com/jumperexchange/jumper-exchange/actions/runs/17210203883/job/48819846408?pr=2212

vinzenzLIFI:
- @chybisov I had to downgrade both playwright and synpress as the latest version seems to have issues building caches. We will need to keep a static version for both until it's resolved. 
- e2e tests seem to be running again now...
- @callMeTheQA, let's keep an eye on new releases from playwright and synpress. We should check from time to time, if we can update to latest versions again.

chybisov:
@vinzenzLIFI Thank you! Works for me in this PR. However, are we sure the issue is on package's side in general? I guess we have some issues in our setup. We should also upgrade @metamask package and resolve all breaking changes it has.

vinzenzLIFI:
@chybisov , regarding the cache building there is most likely only the playwright and synpress dependency as the culprit. Synpress is installing their own custom metamask extension that was not compatible with the playwright version.  So it's not using the Metamask dependencies we are using for Jumper. 
There could still be issues on our setup and I agree that we should keep an eye on that.

chybisov:
@vinzenzLIFI if we don't use @metamask package installed explicitly on Jumper, we should probably remove it 🙂

callMeTheQA (approved):

chybisov:
Please wait till the stable widget release.

PR #2223, opened 2025-08-28
fix: add retry for balance refetching after zap is completed is balance is empty

mmioana:


mmioana:
Closing this ftm, as it seems to not be an actual bug

PR #2230, opened 2025-09-01
fix: zap wait for calls status improvements

mmioana:
Currently need to fix the `FAILED` status from the backend PR as otherwise the widget enters in an infinite loop.

Sister PR https://github.com/jumperexchange/jumper-backend/pull/437

## Testing steps

1. In order to easily test this, please comment `lines 99-100` in `src/providers/ZapInitProvider/WalletClient/quotes.ts`
Screenshot 2025-09-01 at 20 13 09

2. Now try depositing using `LINK` on `Base` for zap `http://localhost:3000/zap/morpho-katana-usdc-ioana`
3. Given the bounds are not set, most likely the 2nd transaction from the supertx should fail
4. If the tx fails, you should see a `partial` bridge successful message in the widget UI
5. You can also try running the unit tests to confirm the `waitForCallsStatus` behaves as expected for the failing first tx from the super tx.
6. To do so run `pnpm test:unit` - will add this to the github actions in another PR

### DIFF
--- a/src/providers/ZapInitProvider/WalletClient/methods.ts
+++ b/src/providers/ZapInitProvider/WalletClient/methods.ts
@@ -1,8 +1,9 @@
 import {
   BaseGetSupertransactionReceiptPayload,
   MeeClient,
+  MeeFilledUserOpDetails,
   MultichainSmartAccount,
-  parseTransactionStatus,
+  UserOpStatus,
   WaitForSupertransactionReceiptPayload,
 } from '@biconomy/abstractjs';
 import {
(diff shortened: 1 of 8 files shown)


src/providers/ZapInitProvider/WalletClient/methods.ts
```diff
@@ -41,15 +52,37 @@ const processTransactionReceipt = (
...
   extraParams: SendCallsExtraParams,
-  hasFailedCleanUpUserOps?: boolean,
+  hasFailedNonCleanUpUserOps?: boolean,
 ) => {
   if (!receipt) {
+    if (hasFailedNonCleanUpUserOps) {
+      return {
+        atomic: true,
+        id: getFormattedTransactionHash(hash),
+        status: 'failed',
```
chybisov: We should have `failure` status here.

src/providers/ZapInitProvider/WalletClient/methods.ts
```diff
@@ -41,15 +52,37 @@ const processTransactionReceipt = (
...
       id: getFormattedTransactionHash(hash),
-      status: hasFailedCleanUpUserOps ? 'failed' : 'success',
-      statusCode: hasFailedCleanUpUserOps ? 500 : 200,
-      receipts: [],
+      status: 'success',
+      statusCode: 200,
+      receipts: [
+        {
+          transactionHash: getFormattedTransactionHash(hash),
+          transactionLink: undefined,
```
chybisov: You can return the same `getMeeScanLink(hash)` here if you'd like it to be on the UI.

admsk2 (approved):

PR #2256, opened 2025-09-04
feat: add line chart component

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-15384

- [x] Need to update theme switching
- [x] Need to update skeleton values
- [x] Need to add a flag for tooltip rendering

### Cons: @nivo/line vs lightweight-charts

| Aspect | @nivo/line | lightweight-charts |
|--------|-----------|--------------------|
| Customization | To follow the design we had to build **custom layers** for grid, x/y axis, and marker. | Needs extra computation for visible range in order to avoid extra padding |
| Extra deps | Needs **d3-shape** to properly follow the design | - |
| Styling | - | **Doesn’t accept CSS vars** directly—needed to grab the computed value in JS. |
| Implementation | Involves **direct SVG manipulation** when doing custom stuff. | API is simpler overall, but a bit less versatile |
| Attribution | - | Need to **show or reference their attribution** somewhere if we hide it on the chart as mentioned [here](https://tradingview.github.io/lightweight-charts/docs#license-and-attribution). |
| Animation | - | Doesn't support fluid animation for the marker and tooltip |
| Bundle size | 82 KB gzipped | 44 KB gzipped |

Screenshot 2025-09-15 at 15 58 34
Screenshot 2025-09-15 at 15 59 59

After the wonderful idea from @laurentsenta we're going to go with `recharts` as they provide almost everything out of the box, beside the custom tooltip, but that was easy to implements ✨


package.json
```diff
@@ -61,6 +66,7 @@
     "i18next-resources-for-ts": "^1.7.4",
     "i18next-resources-to-backend": "^1.2.1",
     "immer": "^10.1.3",
+    "lightweight-charts": "^5.0.8",
```
laurentsenta: Are we merging both libraries? Should we take a decision on which to use? Do you know which libraries other similar products are using?

mmioana: Nope, we'll go with one of them, but would love to hear more opinions based on the analysis 😄 

Lightweight charts is pretty popular, but somehow I feel nivo is more easy to manipulate if we need more customisation

laurentsenta: Got it, if we're deciding between the two, I'll follow your gut feeling.
Do you know which products uses these library?

This is the kind of stats I'd look for if we're hesitating between libraries, checking a few random similar products:

- [recharts](https://www.npmjs.com/package/recharts) 9.8M weekly install
  - Used in [app.hyperwavefi.xyz](https://app.hyperwavefi.xyz/assets/hwhlp) and [app.vaults.fyi](https://app.vaults.fyi/opportunity/mainnet/0x4737D9b4592B40d51e110b94c9C043c6654067Ae)
- [highcharts](https://www.npmjs.com/package/highcharts) 1.5M weekly install
  - Used in [hyperevmscan.io](https://hyperevmscan.io/) and [coinmarketcap.com](https://coinmarketcap.com/currencies/pump-fun/) and [app.morpho.org](https://app.morpho.org/katana/vault/0xE4248e2105508FcBad3fe95691551d1AF14015f7/gauntlet-usdc)
- [lightweight-charts](https://www.npmjs.com/package/lightweight-charts) 0.1M weekly install
- [@nivo/line](https://www.npmjs.com/package/@nivo/line) 0.3M weekly install

mmioana: Thanks a lot @laurentsenta 🫶 I'll check also recharts and make a decision afterwards

laurentsenta (approved):
super precise work, thanks 🙏

PR #2259, opened 2025-09-05
fix: prevent task verification when form fields are empty

mmioana:
fix: prevent task verification when form fields are empty
Jira https://lifi.atlassian.net/browse/LF-15391

## Problem
The `Verify` badge on task cards could be clicked even when required form fields were empty. While the `Verify` button in the widget was properly disabled, the badge remained clickable, causing verification and tracking events to be sent with incomplete data.

## Solution
- Added per-task form state management to the Mission store
- Task verification now checks if the current task has form requirements and validates completion before proceeding
- Form state is tracked per task ID to handle multiple tasks correctly
- Verification is bypassed until all required form fields are completed

## Next Steps
- Add task form state initialization on mission selection to pre-populate form state for all tasks
- Add disabled state to the badge component itself (requires design approval)

## Testing Steps
1. Go to `/missions/win-a-solana-seeker`
2. Select the `Off-chain task`
3. Do not enter anything in the field
4. Click on the `Verify` badge on the task card
5. No posthog event or task verification request should be sent

### DIFF
--- a/src/app/ui/mission/MissionTask.tsx
+++ b/src/app/ui/mission/MissionTask.tsx
@@ -1,4 +1,4 @@
-import { FC, useEffect, useRef } from 'react';
+import { FC, useCallback, useEffect, useRef } from 'react';
 import { TaskVerificationWithApy } from 'src/types/loyaltyPass';
 import { TaskCard } from 'src/components/Cards/TaskCard/TaskCard';
 import { Badge } from 'src/components/Badge/Badge';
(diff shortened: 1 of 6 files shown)


callMeTheQA:
LGTM.

https://github.com/user-attachments/assets/51e4b694-b330-4f41-bc28-4ee80f437f5b

callMeTheQA (approved):

PR #2268, opened 2025-09-09
feat: Privacy Policy;

admsk2:


src/app/[lng]/(infos)/privacy-policy/page.tsx
```diff
@@ -0,0 +1,35 @@
...
+  alternates: {
+    canonical: `${getSiteUrl()}/privacy-policy`,
+  },
+};
+
+export const revalidate = 300;
+
+export default function Page() {
+  return (
+    <Suspense fallback={<div>Loading...</div>}>
```
Comment (oktapodia) [inline src/app/[lng]/(infos)/privacy-policy/page.tsx:31]: I do not think that the Suspense is needed as it should be statically generated

Comment (oktapodia) [inline src/app/[lng]/(infos)/privacy-policy/page.tsx]: I do not think that the Suspense is needed as it should be statically generated

oktapodia (dismissed):
Just a comment

src/components/PrivacyPolicy/PrivacyPolicyPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+import { PageContainer } from '@/components/Containers/PageContainer';
+import { DynamicPagesContainer } from '@/components/DynamicPagesContainer';
+import { Box, Typography, List, ListItem, ListItemText } from '@mui/material';
+import { format } from 'date-fns';
+
+export const PrivacyPolicyPage = () => {
+  const currentDate = format(new Date(), 'MMMM dd, yyyy');
+
+  const sections = [
```
mmioana, line 11: Should we move all this content inside the translation file and then call the `t(<KEY_HERE>)`?

admsk2: i don't think privacy policy will be translated, but if so then yeah, i will change the approach a bit

mmioana: No, no need, but then I guess we can remove these translations https://github.com/jumperexchange/jumper-exchange/pull/2268/files#diff-6b05968038c982f0197df42120e6ef4e936c5e30aa8c3227f4707d247b739b2fR267-R351 ?

mmioana (approved):

PR #2271, opened 2025-09-13
feat: Withdraw tab;

admsk2:


src/components/Zap/ZapWidgetStack.tsx
```diff
@@ -55,11 +104,15 @@ export const ZapWidgetStack: FC<ZapWidgetStackProps> = ({
...
+          <Box sx={{ marginBottom: 3 }}>
+            <HorizontalTabs
+              tabs={tabs}
+              value={activeTab}
+              onChange={handleTabChange}
             />
+          </Box>
+          <ClientOnly>
+            {renderWidget()}
           </ClientOnly>
```
mmioana: You can skip using the state as the `HorizontalTabs ` handles that internally. Those props are mostly needed if you want to add side-effects on top or the active tab is not necessarily the last selected one - see `src/components/Navbar/layout/MobileLayout.tsx`

So maybe something like - ofc you can use the `renderWidget` actually and just pass the current active tab
```
<HorizontalTabs
            tabs={tabs}
            renderContent={(currentTab) => (
              <ClientOnly>
                {currentTab === 'deposit' ? (
                  <ZapDepositBackendWidget
                    ctx={ctx}
                    customInformation={customInformation}
                  />
                ) : (
                  <ZapWithdrawWidget
                    ctx={ctx}
                    customInformation={customInformation}
                  />
                )}
              </ClientOnly>
            )}
            sx={{
              mb: 3,
              width: '100%',
              '& .MuiButtonBase-root': { flexGrow: 1, maxWidth: '100%' },
            }}
          />
```

src/components/Zap/ZapWidgetStack.tsx
```diff
@@ -34,6 +39,50 @@ export const ZapWidgetStack: FC<ZapWidgetStackProps> = ({
...
+    isLoadingDepositTokenData,
+  } = useEnhancedZapData(projectData);
+
+  const hasDeposited = !isLoadingDepositTokenData && !!depositTokenData;
+  const hasWithdrawAbi = !!zapData?.abi?.withdraw;
+
+  const tabs = useMemo(() => [
+    {
+      value: 'deposit',
+      label: 'Deposit',
```
mmioana: Can we use the translation here?

src/components/Zap/ZapWidgetStack.tsx
```diff
@@ -34,6 +39,50 @@ export const ZapWidgetStack: FC<ZapWidgetStackProps> = ({
...
+    },
+    {
+      value: 'withdraw',
+      label: 'Withdraw',
```
mmioana: And here 👀

src/i18n/translations/en/translation.json
mmioana, line 1: Please run `i18next-resources-for-ts` so we also update the `resources.d.ts`

mmioana (changes requested):
Left a few comments @admsk2 if you can please check 🙏

mmioana (approved):

PR #2274, opened 2025-09-15
feat: add on route selected and chain pinned - LF-14810 & LF-15119

laurentsenta:
Closes https://lifi.atlassian.net/browse/LF-14810
Closes https://lifi.atlassian.net/browse/LF-15119

- [x] Add back "on route selected" events and introduce "chain pinned" events
- [x] Introduce basics for a posthog tracker that cares "only" about moving events around, not internal state, can be wrapped with more code (hence the data driven approach)
- [x] Introduce a WidgetEventsManager and its config to make it data-driven (easier to reuse, wrap, and move around)
- [x] Update WidgetEvents and WidgetTrackingProvider to use that new system

**Test:**

- select a route, check posthog, the event should be visible in posthog
- pin a few chains in the widget, then unpin some, the events should be visible in posthog

Note on testing posthog: https://www.notion.so/lifi/Test-Posthog-Events-270f0ff14ac7804d898dfe606a75af12

Example:

### DIFF
--- a/src/components/Widgets/WidgetEvents.tsx
+++ b/src/components/Widgets/WidgetEvents.tsx
@@ -18,6 +18,7 @@ import type { RouteExtended } from '@lifi/sdk';
 import { type Route } from '@lifi/sdk';
 import { useAccount } from '@lifi/wallet-management';
 import type {
+  ChainPinned,
   ChainTokenSelected,
   ContactSupport,
   RouteExecutionUpdate,
@@ -357,6 +358,18 @@ export function WidgetEvents() {
       });
     };
 
+    const onChainPinned = async ({ chainId, pinned }: ChainPinned) => {
+      trackEvent({
+        category: TrackingCategory.WidgetEvent,
+        action: TrackingAction.OnChainPinned,
+        label: 'chain_pinned',
+        data: {
+          [TrackingEventParameter.ChainId]: chainId,
+          [TrackingEventParameter.Pinned]: pinned,
+        },
+      });
+    };
+
     widgetEvents.on(WidgetEvent.RouteExecutionStarted, onRouteExecutionStarted);
     widgetEvents.on(
       WidgetEvent.LowAddressActivityConfirmed,
@@ -386,6 +399,7 @@ export function WidgetEvents() {
     widgetEvents.on(WidgetEvent.PageEntered, onPageEntered);
     widgetEvents.on(WidgetEvent.SettingUpdated, onChangeSettings);
     widgetEvents.on(WidgetEvent.RouteSelected, onRouteSelected);
+    widgetEvents.on(WidgetEvent.ChainPinned, onChainPinned);
 
     return () => {
       widgetEvents.off(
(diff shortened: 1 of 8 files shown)


oktapodia (approved):

mmioana (approved):

mmioana (commented):
Actually @laurentsenta can we please add this event also to the `WidgetTrackingProvider`? Planning on using this in the future also for the main widget

laurentsenta:
@mmioana what's the difference between WidgetTrackingProvider and WidgetEvents? What are they used for? Why do they look so similar?

mmioana:
@laurentsenta I've extracted the logic for sending tracking events in `WidgetTrackingProvider` mostly for zap/missions. Didn't had the time to untangle yet the `WidgetEvents` which beside handling event tracking has some other logic inside. So as an intermin solution until the following days when I'll do some work on this, it would be nice to tackle this. Otherwise, I'll put in on my list

laurentsenta:
@mmioana got it, I'll dig a bit into it to figure out how to add the new events there.

Why do we need different events for zaps? Is this something Octave requested? `action_available_routes_zap` seems less pratical than have `action_available_routes` with a flag for zaps in the payload.

src/components/Widgets/WidgetEvents.tsx
```diff
@@ -245,7 +250,14 @@ export function WidgetEvents() {
       setDestinationChainToken(toChainData);
     };
 
-    const onLowAddressActivityConfirmed = async (props: {
+    const destinationChainTokenSelected = async (
+      destinationData: ChainTokenSelected,
+    ) => {
+      destinationChainTokenSelectedMultisig(destinationData);
+      destinationChainTokenSelectedRaw(destinationData);
+    };
```
laurentsenta, line 258: Example

laurentsenta:
@mmioana PR updated with a refactor that'll make easier to merge the code:

[Commit 1](https://github.com/jumperexchange/jumper-exchange/pull/2274/commits/85b19133c90a0ac38ceec9cb4b8ed55fca5ee87a) Introduce basics for a posthog tracker that cares "only" about moving events around, not internal state, can be wrapped with more code (hence the data driven approach).


[Commit 2](https://github.com/jumperexchange/jumper-exchange/pull/2274/commits/6ba35d62bdb0dbb317b646647123bbda4c0d88e8)  Introduce a WidgetEventsManager and its config to make it data-driven (easier to reuse, wrap, and move around) + clarify and unify naming.

[Commit 3](https://github.com/jumperexchange/jumper-exchange/pull/2274/commits/00ed423a60b44cf04e994ee315ee749a7735bff9) Reuse that code in the other widget setup.

The core idea is to lay down future work, we can easily:
- Move reusable code out of WidgetTrackingProvider and WidgetEvents and into the posthog tracker
- Implement local tweaks if we need state management by reusing the posthog functions and "just" wrapping them ([example](https://github.com/jumperexchange/jumper-exchange/pull/2274/commits/6ba35d62bdb0dbb317b646647123bbda4c0d88e8#r2354974653))

Also, I really really think we should think about sending regular events with a `isZap` data flag

mmioana (approved):

callMeTheQA:
looks good 
Thanks @laurentsenta 
Screenshot 2025-09-17 at 18 49 50

PR #2279, opened 2025-09-16
fix: prevent widget re-rendering on bridge conditions change

mmioana:
fix: prevent widget re-rendering on bridge conditions change
chore: simplify logic for showing or hiding to address field based on bridge conditions

Reference issue for testing: https://lifi.atlassian.net/browse/LF-13896


oktapodia (approved):

callMeTheQA:
hi @mmioana , here is the outcome of testing this PR on the local env :
1.  Selecting chain now working as it should 
2. Abstract chain is not visible under "all chains" on /zap but I'm able to pass e.g "?fromChain=2741&fromToken=0x0000000000000000000000000000000000000000" directly into URL and in that case ETH (Abstract) will appears into deposit widget.
Screenshot 2025-09-16 at 20 04 35

mmioana:
Nice catch @callMeTheQA 🙌 I've added a new Jira issue for the fix on the `/zap` page https://lifi.atlassian.net/browse/LF-15567

PR #2286, opened 2025-09-17
fix: zap withdraw widget styling + prevent form submission on empty value

mmioana:
Issues tackled:

- Prevent form submission if value is empty: https://www.notion.so/lifi/DEVELOP-Redeem-button-is-not-disabled-when-amount-field-is-0-271f0ff14ac7804ab03eff79c1e7b23e
- Max button for which I actually used the same styling/logic as for the widget https://www.notion.so/lifi/DEVELOP-Max-amount-is-splitted-into-two-rows-on-Withdraw-tab-271f0ff14ac7800ea923e44a804b0bd1
- Set the horizontal tabs to full column width

### DIFF
--- a/src/components/Button/Button.style.ts
+++ b/src/components/Button/Button.style.ts
@@ -1,7 +1,7 @@
 'use client';
 import type { ButtonProps as MuiButtonProps } from '@mui/material';
-import { Button as MuiButton, alpha } from '@mui/material'; //ButtonProps
-import { styled } from '@mui/material/styles';
+import { alpha, styled } from '@mui/material/styles';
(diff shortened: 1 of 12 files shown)


callMeTheQA:
@mmioana can you please remove shadow on hover over Redeem button ( since that shadow doesn't exist on the deposit button ) 

https://github.com/user-attachments/assets/352b38e5-e03f-4a85-9867-1601d5dfee98

2nd issue regarding max amount which was splitted into two rows - looks good now ✅

callMeTheQA (approved):

PR #2291, opened 2025-09-18
chore: refactor use or remove use widget config

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-12395

### Main widget

1. Go to `/`
2. Check widget still behaves as before
3. Try depositing
4. Try connecting with an `Abstract` wallet and check the required to address is shown

### Mission widget
1. Go to `/missions/test123`
2. Check the widget is pre-populated with values
3. Check widget still behaves as before
4. Try depositing

### Zap widget
1. Go to `/zap/morpho-katana-usdc-ioana`
2. Check the widget is pre-populated with values
3. Check widget still behaves as before
4. Try depositing


src/components/composite/DepositModal/DepositModal.tsx
```diff
@@ -85,6 +85,7 @@ export const DepositModal: FC<DepositModalProps> = ({
             )}
             <ClientOnly>
               <ZapDepositWidget
+                type="zap"
```
laurentsenta: Shouldn't that be set by default in the component?

src/components/Widgets/variants/base/ZapWidget/ZapDepositBackendWidget.tsx
```diff
@@ -120,22 +122,15 @@ export const ZapDepositBackendWidget: FC<ZapDepositBackendWidgetProps> = ({
...
       ...ctx,
-      includeZap: true,
       zapPoolName: poolName,
-      baseOverrides,
+      integrator: 'zap.morpho',
+      keyPrefix: 'zap.backend',
+      variant: 'wide' as const,
+      formData: {
+        minFromAmountUSD,
+      },
```
laurentsenta, line 132: how come the integrator is always `zap.morpho`? should that depends on the current zap page?

src/components/Widgets/variants/mission/MissionZapDepositWidget.tsx
```diff
@@ -20,15 +21,17 @@ export const MissionZapDepositWidget: FC<MissionZapDepositWidgetProps> = ({
...
-      // fromAmount,
-      toAddress,
-      taskType: currentActiveTaskType,
+      formData: {
+        toAddress,
+        // destinationChain,
+        // destinationToken,
+        // sourceChain,
+        // sourceToken,
+        // fromAmount,
```
laurentsenta, line 32: remove dead code

```suggestion
```

src/components/Widgets/variants/mission/MissionZapDepositWidget.tsx
```diff
@@ -40,5 +43,11 @@ export const MissionZapDepositWidget: FC<MissionZapDepositWidgetProps> = ({
     currentActiveTaskType,
   ]);
 
-  return <ZapDepositWidget ctx={ctx} customInformation={customInformation} />;
+  return (
+    <ZapDepositWidget
+      type="zap"
```
laurentsenta: why not set it by default, `ZapComponent type="zap"` seems redundant

src/components/Widgets/variants/base/ZapWidget/ZapDepositBackendWidget.tsx
mmioana: Leaving it for the moment to the previously hard coded value that was set during the zap backend migration, as we'll use a different key as we do for the main widget to properly handle the fees

src/components/composite/DepositModal/DepositModal.tsx
mmioana: Good call 🙌

src/components/Widgets/variants/mission/MissionZapDepositWidget.tsx
mmioana: Still leaving the dead code only because we should at some point after reaching a stable state with zaps migrate the `CustomInformation.projectData` to make use of the widget configuration in Strapi for the form values at least

callMeTheQA (approved):

src/components/Widgets/variants/mission/MissionZapDepositWidget.tsx
laurentsenta: That's something that should live in our project management and git history tbh, when someone else reads the code, they won't have the context you just shared, it makes the code longer and less clear.

laurentsenta (approved):

PR #2292, opened 2025-09-18
fix: use merkl rewards defined in strapi for filtering

mmioana:


src/app/lib/getMerklRewards.ts
```diff
@@ -0,0 +1,25 @@
...
+    },
+  });
+
+  if (!res.ok) {
+    throw new Error('Failed to fetch data');
+  }
+
+  const data: StrapiResponse<MerklRewardsData> = await res.json();
+
+  return { data };
```
laurentsenta, line 24: You're doing `{ data: merklRewardsResponse }` then `merklRewardsResponse.data` in the page, return the payload & simplif the code:

```suggestion
  return data.data;
```

laurentsenta (approved):

src/app/lib/getMerklRewards.ts
mmioana: I'll tackle this in a separate PR in order to align all ways data is returned from a Strapi endpoint 😓

PR #2295, opened 2025-09-19
chore: include lifi error code key to tx tracking

mmioana:
Given the route process error code is not very informative, if lifi has a key for the error code, we send also that as a parameter for the wallet tx tracking

Sister PR https://github.com/jumperexchange/jumper-backend/pull/478

### DIFF
--- a/src/utils/routes/routeProcess.ts
+++ b/src/utils/routes/routeProcess.ts
@@ -1,8 +1,9 @@
-import type {
-  LiFiStep,
-  LiFiStepExtended,
-  Process,
-  RouteExtended,
+import {
+  LiFiErrorCode,
+  type LiFiStep,
+  type LiFiStepExtended,
+  type Process,
+  type RouteExtended,
 } from '@lifi/sdk';
 import { TrackingEventParameter } from 'src/const/trackingKeys';
 import { getDetailInformation } from './routeUtils';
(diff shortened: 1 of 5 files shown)


src/utils/routes/routeProcess.ts
```diff
@@ -15,6 +16,18 @@ interface GetProcessInformationType {
...
+  if (!code) return null;
+
+  return (
+    Object.keys(LiFiErrorCode).find(
+      (key) =>
+        LiFiErrorCode[key as keyof typeof LiFiErrorCode].toString() ===
+        code.toString(),
+    ) || null
+  );
+};
```
laurentsenta, line 25: That's a bit confusing. Could you use [findKey](https://lodash.com/docs/4.17.15#findKey) or something like 

```ts
const found = Object.values(xs).find(([,value]) => value.toString() == code.toString())
return found[0] ?? null
```

also type the return value.

src/utils/routes/routeProcess.ts
```diff
@@ -37,10 +50,14 @@ export const getProcessInformation = (
...
           );
         }
+        const errorCodeKey = findErrorKeyFromErrorCode(process.error?.code);
         errors = {
           ...(process.error?.code && {
             [TrackingEventParameter.ErrorCode]: process.error?.code,
           }),
+          ...(errorCodeKey && {
+            [TrackingEventParameter.ErrorCodeKey]: errorCodeKey,
+          }),
```
laurentsenta: For what it's worth, in https://github.com/jumperexchange/jumper-exchange/pull/2274/files#diff-42e2cde0e9c382fed74f4cf44e1b1ee69bb0a764cfcf17e5e7225408c126e09b I refactored that pattern just to use explicit `if`s, the code was getting harder and harder to read, types where wrong, etc.

laurentsenta (approved):

src/utils/routes/routeProcess.ts
mmioana: You should be called the `lodash` oracle sir 🙌

PR #2297, opened 2025-09-22
feat: add test for navigating to the Privacy Policy page 

callMeTheQA:
Corresponding JIRA ticket - https://lifi.atlassian.net/browse/LF-15670

### DIFF
--- a/tests/testData/values.json
+++ b/tests/testData/values.json
@@ -5,6 +5,7 @@
   "discordURL": "https://discord.com/invite/jumperexchange",
   "telegramURL": "https://t.me/officialjumperexchange",
   "link3URL": "https://link3.to/jumperexchange",
+  "privacyPolicyURL": "http://localhost:3000/privacy-policy",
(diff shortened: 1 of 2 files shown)


tests/testData/values.json
```diff
@@ -5,6 +5,7 @@
   "discordURL": "https://discord.com/invite/jumperexchange",
   "telegramURL": "https://t.me/officialjumperexchange",
   "link3URL": "https://link3.to/jumperexchange",
+  "privacyPolicyURL": "http://localhost:3000/privacy-policy",
```
vinzenzLIFI: since in the playwright config we have set baseURL to be `process.env.BASE_URL || 'http://localhost:3000'` it is used throughout all playwright functions. You should not need to set it explicitly in the url anymore.
This also applies to all `goto` functions.

callMeTheQA: thanks @vinzenzLIFI  good catch, I fixed it  🔧

vinzenzLIFI (approved):

PR #2304, opened 2025-09-24
feat: earn filter pixel perfect apy select

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-15675


src/providers/ThemeProvider/MUIThemeProvider.tsx
mmioana, line 1: Will need to create a PR directly to `develop` for this one

callMeTheQA (approved):
LGTM

https://github.com/user-attachments/assets/55687b42-2a4c-4e13-8166-0e92e2aeae71

PR #2307, opened 2025-09-25
feat: update earn filtering to pixel perfect design

mmioana:
Figma link: https://www.figma.com/design/JUaT4CQ9YSCQ4IWsQQSRIK/Earn?node-id=370-60700&t=qD4CRav8BD3Sjr2s-0

Jira: https://lifi.atlassian.net/browse/LF-15676

## Testings steps
1. Please open storybook and check the `Earn/FilterBar` to match Figma

> [!NOTE]
> 1. It was agreed with design to have a "fade" animation for the bar's content when switching between `For you` and `All markets`
> 2. It was agreed with design to not show a filter if there is no value available, or in case of the range slider if min and max are the same
> 3. `For you` icon has been updated 

2. Please check the earn card component has `surface1` as background color
3. Refactored the opportunities section on the `/earn` page to match Figma, so that can be checked


src/app/ui/earn/EarnFilteringContext.tsx
```diff
@@ -69,6 +79,9 @@ export const EarnFilteringProvider = ({
   const address: Hex | undefined = useAccountAddress();
   const usedYourAddress = address !== undefined;
 
+  // TODO: introduce the loading state?
+  const [sortBy, setSortBy] = useState<SortByOptions>(SortByOptions.APY);
```
laurentsenta, line 83: For what it's worth, in https://github.com/jumperexchange/jumper-backend/pull/482
I add sorting there: https://github.com/jumperexchange/jumper-backend/blob/b874e5a11cfaa63ca97739eaed7f182c0a6f25a9/src/earn/earn.dto.ts#L168-L202

We'll reuse the same loading state

laurentsenta (approved):
impecabil 💯

src/app/ui/earn/EarnFilteringContext.tsx
mmioana: Perfect! Then we'll remove the type in a follow-up PR

callMeTheQA:
LGTM

Screenshot 2025-09-29 at 10 19 08

callMeTheQA (approved):

PR #2317, opened 2025-10-02
feat: earn filtering store filter state in url

mmioana:
Issue: https://lifi.atlassian.net/browse/LF-15663

> [!NOTE]
> 1. Initially wanted to optimise fetching for all opportunities and first batch of filtered ones on the server, but will not optimise too much, given we'll want to fetch also the for you opportunities that are wallet dependant. So skipping this for now.
> 2. Added the filtering also by APY min and max values

## Testing steps
1. Go to `/earn`
2. Play will the filters and check the search params update occordingly
3. Refresh the page
4. Check the search params are still there
5. Check the results are still filtered
6. Remove all filters
7. Check the search params are cleared up


callMeTheQA (approved):
LGTM

src/hooks/earn/useEarnTopOpportunities.ts
```diff
@@ -26,5 +26,6 @@ export const useEarnTopOpportunities = ({}: Props): Result => {
       return result.data.data;
     },
     refetchInterval: FIVE_MINUTES_MS,
+    enabled: !!address,
```
laurentsenta: Don't, you'll disable the top cards,
we're designing the app to show opportunities even if you're not logged in,

https://github.com/jumperexchange/jumper-backend/blob/656e620a38bd0d1af5587aa4ad136eba3ded2d20/src/earn/earn.dto.ts#L163

src/hooks/earn/useEarnFilterOpportunities.ts
```diff
@@ -27,5 +30,6 @@ export const useEarnFilterOpportunities = ({ filter }: Props): Result => {
       return result.data.data;
     },
     refetchInterval: ONE_HOUR_MS,
+    enabled,
```
laurentsenta: I suspect this should be removed as well.

src/app/ui/earn/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+import {
+  EarnFilteringParams,
+  EarnsPageSearchParams,
+  FilterKey,
+  SortByOptions,
+} from './types';
+import { EarnOpportunityFilter } from 'src/app/lib/getOpportunitiesFiltered';
+import { toFixedFractionDigits } from 'src/utils/formatNumbers';
+
+const parseString = (value: string): string | null => {
```
laurentsenta: Instead of rolling our own serialize / deserialize, could we use something like nuqs? It seems to be well maintained and is nextjs first (I guess we can easily use it in ssr and all).

https://www.npmjs.com/package/nuqs

src/app/ui/earn/types.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export interface EarnFilteringParams {
+  allChains: Chain[];
+  allProtocols: Protocol[];
+  allAssets: Token[];
+  allTags: string[];
+  allAPY: Record<number, number>; // histogram of apy
+}
+
+export interface EarnsPageSearchParams {
```
laurentsenta: Could you reuse the `EarnOpportunityFilter` sync'd with the backend? It'll catch type errors on API changes.

src/app/ui/earn/types.ts
```diff
@@ -0,0 +1,30 @@
+import { EarnOpportunityFilter } from 'src/app/lib/getOpportunitiesFiltered';
+import { Chain, Protocol, Token } from 'src/types/jumper-backend';
+
+// TODO: migrate to backend's typing
+export enum SortByOptions {
+  APY = 'apy',
+  TVL = 'tvl',
+}
```
laurentsenta: fwiw, there's a way to sync the enum with the backend:

```ts
// Comes from backend: "apy" | "tvl" | "slug" | "chain" | "protocol" | "asset"
type SortByType = NonNullable<NonNullable<EarnOpportunityFilter>['sortBy']>;

// The options we use in the frontend for now 
const SortByEnum = {
  APY: 'apy',
  TVL: 'tvl',
} as const satisfies Record<string, SortByType>; // ensures overlaps

export type SortByEnum = (typeof SortByEnum)[keyof typeof SortByEnum]; // Expose a type
```

This would catch stuff like: we changes sorting options in the backend, and now our frontend is outdated.
But I concede it's a bit convoluted.

src/app/[lng]/earn/page.tsx
```diff
@@ -11,10 +13,16 @@ export const metadata: Metadata = {
...
 };
 
-export default async function Page() {
+interface PageProps {
+  searchParams: Promise<EarnsPageSearchParams>;
+}
+
+export default async function Page(props: PageProps) {
+  const searchParams = await props.searchParams;
+  const initialFilters = parseFiltersFromUrl(searchParams);
```
Comment (laurentsenta) [inline src/app/[lng]/earn/page.tsx]: To me that's leaking internals of the filtering context, it should be the only one manipulating filtering related data. Could you keep it inside the filtering context?

src/app/ui/earn/EarnFilteringContext.tsx
```diff
@@ -57,82 +39,99 @@ export const EarnFilteringContext = createContext<EarnFilteringContextType>({
...
+
+  const forYou = useEarnFilterOpportunities(
+    {
+      filter: {
+        forYou: true,
+        address,
+      },
     },
-  });
+    !!address,
```
laurentsenta: Out of scope for this PR, and we do want to serve "for you opportunities" even if the user is not logged in.

src/app/ui/earn/EarnFilteringContext.tsx
```diff
@@ -57,82 +39,99 @@ export const EarnFilteringContext = createContext<EarnFilteringContextType>({
...
+      return EMPTY_FILTERING_PARAMS;
     }
 
-    setInitialData((current) => {
-      if (current.length === 0) {
-        return all.data;
+    return extractFilteringParams(allNoFilter.data);
+  }, [allNoFilter.data]);
+
+  const updateSearchParams = useCallback(
```
laurentsenta: too low level for the "filtering manager", use a library or extract to the utils where we defined the serialized, etc.

When you're in the context of the filtering provider you're trying to express "I want to update the search params with new values", but we're writing "I want to create a url params manager, then for each entries of my value, serialize them and call sets or delete on these". That's too low level.

src/app/ui/earn/EarnOpportunitiesAll/EarnOpportunitiesAllSkeleton.tsx
```diff
@@ -0,0 +1,24 @@
+'use client';
```
laurentsenta: Why use client?

src/app/ui/earn/EarnOpportunitiesAll/EarnOpportunitiesAllServer.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export const EarnOpportunitiesAllServer: FC<
+  EarnOpportunitiesAllServerProps
+> = async ({ initialFilters }) => {
+  return (
+    <Suspense fallback={<EarnOpportunitiesAllSkeleton />}>
+      <EarnOpportunitiesAllClient initialFilters={initialFilters} />
+    </Suspense>
+  );
+};
```
laurentsenta: From what I understand we're adding 3 files to render skeletons on the server side (we don't get data on the server yet, so we're just moving skeletons).

Could we use something like https://nextjs.org/docs/app/api-reference/file-conventions/loading instead? That seems more nextjs'y, we'd keep one file to render opportunities and have one server side skeleton in the loading.js.

src/app/ui/earn/EarnOpportunitiesAll/EarnOpportunitiesAllSkeleton.tsx
mmioana: Only because we were using it in a server component as a fallback and we can't use the `styled` without specifying it's a client component. However, given I've removed that part, this is no longer present 🙌

src/app/ui/earn/EarnOpportunitiesAll/EarnOpportunitiesAllServer.tsx
mmioana: Totally agree, though loading.ts isn't currently used in our app, so I'd prefer sticking with our existing pattern for consistency. We could explore this in a separate PR.
And I think modular skeleton components for specific page sections makes sense, as this makes it easier to update components independently or migrate parts of the page to SSR.

laurentsenta (approved):

PR #2321, opened 2025-10-03
feat: add earn analytics chart section to earn details page

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-15994

> [!NOTE]
> Will treat the links from the earn cards to the earn details page on a separate PR once the other ones get merged to avoid too many merging conflicts 🏳️ 

## Testing steps

1. Please go to `/earn/aave-v3-prime-usdc-on-mainnet` or `/earn/clearstar-openeden-usdc-on-base`
2. Check the analytics chart section matches [Figma](https://www.figma.com/design/JUaT4CQ9YSCQ4IWsQQSRIK/Earn?node-id=361-41241&m=dev)

### DIFF
--- a/src/app/ui/earn/EarnPageSkeleton.tsx
+++ b/src/app/ui/earn/EarnPageSkeleton.tsx
@@ -1,5 +1,3 @@
-'use client';
-
 import { EarnDetailsAnalyticsSkeleton } from 'src/components/EarnDetails/EarnDetailsAnalyticsSkeleton';
 import { EarnDetailsSection } from 'src/components/EarnDetails/EarnDetailsSection';
 
--- a/src/app/ui/earn/EarnPageSkeleton.tsx
+++ b/src/app/ui/earn/EarnPageSkeleton.tsx
@@ -1,3 +1,15 @@
+'use client';
+
+import { EarnDetailsAnalyticsSkeleton } from 'src/components/EarnDetails/EarnDetailsAnalyticsSkeleton';
+import { EarnDetailsSection } from 'src/components/EarnDetails/EarnDetailsSection';
+
 export const EarnPageSkeleton = () => {
-  return <div>EarnPageSkeleton</div>;
+  return (
+    <>
+      <EarnDetailsSection>
+        <EarnDetailsAnalyticsSkeleton />
+      </EarnDetailsSection>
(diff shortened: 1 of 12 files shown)


callMeTheQA:
verified on feature branch.
Screenshot 2025-10-03 at 16 44 40

callMeTheQA (approved):

src/app/ui/earn/EarnPageSkeleton.tsx
```diff
@@ -1,3 +1,15 @@
+'use client';
```
laurentsenta: why use client?

src/components/EarnDetails/hooks.ts
```diff
@@ -0,0 +1,81 @@
...
+      setQuery((query) => ({ ...query, range }));
+    },
+    [setQuery],
+  );
+
+  return useMemo(
+    () => ({
+      ...result,
+      value: query.value as AnalyticsValueFieldEnum,
+      range: query.range as AnalyticsRangeFieldEnum,
```
laurentsenta: why do you need the type cast?

src/app/ui/earn/EarnPageSkeleton.tsx
mmioana: Only because if any of this children uses `styled` components they can't be rendered on the server-side

src/components/EarnDetails/hooks.ts
mmioana: Didn't pay close attention. Eagle eyes 🦅 🥇

src/app/ui/earn/EarnPageSkeleton.tsx
laurentsenta: Shouldn't we make those component `use client` instead? I thought [nesting was sort of the point](https://nextjs.org/docs/app/api-reference/directives/use-client#nesting-client-components-within-server-components). Next is even supposed to let you [interleave components](https://nextjs.org/docs/app/getting-started/server-and-client-components#interleaving-server-and-client-components) when we get to SSR.

mmioana: Ah yes, could make actually more sense. I did it at the upper level knowing already the children should be client components

laurentsenta (approved):

PR #2328, opened 2025-10-09
feat: ui for zap sweep card

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-16065

### DIFF
--- a/src/components/ZapWidget/SweepTokensCard/SweepTokensCard.tsx
+++ b/src/components/ZapWidget/SweepTokensCard/SweepTokensCard.tsx
@@ -1,0 +1,158 @@
+import { FC, useMemo } from 'react';
+import { SectionCardContainer } from 'src/components/Cards/SectionCard/SectionCard.style';
+import { useSweepTokens } from 'src/hooks/zaps/useSweepTokens';
+import CheckIcon from '@mui/icons-material/Check';
+import ErrorIcon from '@mui/icons-material/Error';
+import { CustomInformation } from 'src/types/loyaltyPass';
+import { ProjectData } from 'src/types/questDetails';
+import { Button } from 'src/components/Button';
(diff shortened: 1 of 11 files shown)


src/components/ZapWidget/SweepTokensCard/SweepTokensCard.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const onClickHandler = async () => {
+    if (txHash) {
+      openInNewTab(`https://meescan.biconomy.io/details/${txHash}`);
+      return;
+    }
+
+    try {
+      await sweepTokens();
+    } catch (error) {
+      console.error('Sweep failed:', error);
```
admsk2, line 82: is there going to be any other indicator that sweep failed other than log in the console?

src/components/ZapWidget/SweepTokensCard/SweepTokensCard.tsx
```diff
@@ -0,0 +1,158 @@
...
+    });
+  }, [sweepableTokens]);
+
+  if (!hasTokensToSweep) {
+    return null;
+  }
+
+  const onClickHandler = async () => {
+    if (txHash) {
+      openInNewTab(`https://meescan.biconomy.io/details/${txHash}`);
```
admsk2: do we have constants somewhere to put that there?

src/components/ZapWidget/SweepTokensCard/SweepTokensCard.tsx
mmioana: From the design perspective we're just going to go back to the initial state 

https://www.figma.com/design/MmAjz3cDZU2BbPDFeisAdi/Missions--Webpage-?node-id=1662-123172&m=dev

But maybe it makes sense to catch the error in Sentry, nice catch

admsk2 (approved):

PR #2331, opened 2025-10-10
feat: missions ended generic indicator for missions

mmioana:
https://lifi.atlassian.net/browse/LF-15763

- [x] Need to validate logic for the ended missions

Sister PR: https://github.com/jumperexchange/strapi-cms/pull/78

 
1. Go to `/missions`
2. Check missions that have a single token reward (eg. `Hyperwave on HyperEVM`) shows both the token logo and the amount
3. Now check `Test New Version` mission shows up also the generic labels
4. Add some more labels in Strapi to this mission `Test New Version`
5. Check the overflow is shown as `+X`
6. Now for a mission that has not ended yet (endDate in the future), toggle the hasEnded flag
7. Check the `Ended` status tag is shown both on the mission and mission detail page


src/hooks/useOverflowItems.ts
mmioana, line 1: Wanted to use a package for this, but none of the available ones was that maintained or used, so went for a custom solution for now

oktapodia (approved):

tests/connectWallet.spec.ts
mmioana, line 1: @callMeTheQA was it intended to add all these test changes on this PR?

PR #2350, opened 2025-10-20
chore: add data test id to earn card elements

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-16209

## Testing steps
1. Go to `/earn`
2. Ensure each card contains a section having the `protocol-{protocolName}` test id for the upper section of the card
3. Ensure each item (tvl/apy/asset/lockup) have the test id set up
4. Change the view move for the list and ensure the test ids are still present
5. Go to the earn detail page
6. Check the overview card has the test ids set up


callMeTheQA:
LGTM on feature branch.
Screenshot 2025-10-23 at 10 32 05

src/components/Cards/EarnCard/variants/CompactEarnCard.tsx
```diff
@@ -39,6 +39,7 @@ export const CompactEarnCard: FC<Omit<EarnCardProps, 'variant'>> = ({
     return (
       <CompactEarnCardItem
         key={item.key}
+        dataTestId={item.dataTestId}
```
laurentsenta, line 42: It looks like we're leaking the test ids in the application model, can't we just use `card-${item.name}`?

src/hooks/earn/useFormatDisplayEarnOpportunityData.tsx
```diff
@@ -71,16 +80,20 @@ const buildTvlItem = (
     return null;
   }
 
+  const compactValue = toCompactValue(tvlUsdNumber);
+  const formatted = `$${compactValue}`;
+
   return {
     key: 'tvl',
+    dataTestId: `tvl-${compactValue}`,
```
laurentsenta, line 88: Same here and above cases, we're moving around dataTestIds, it looks like we're storing the value in the test ids, and we're leaking test ids in the app model. Can't you "just" render `data-test-id=${key}` in the component that render those? And not change the data model.

src/components/Cards/EarnCard/variants/CompactEarnCard.tsx
mmioana: I agree this is a leak of test ids in the application model, however, I think it will help us for the e2e tests if we want to make sure the filters work properly based on the data displayed in the cards (thus adding the `protocol`/`asset` etc). So I'd prefer to keep it unless you feel strongly against it

laurentsenta (approved):

PR #2353, opened 2025-10-21
feat: Jumper Scan routing via backend;

admsk2:


src/providers/FetchInterceptorProvider.tsx
```diff
@@ -14,9 +14,14 @@ export function FetchInterceptorProvider() {
       const zapQuestId = sessionStorage.getItem(
         ZAP_QUEST_ID_SESSION_STORAGE_KEY,
       );
+      // zaps flow
       if (request.url.startsWith(envConfig.NEXT_PUBLIC_LIFI_API_URL)) {
         request.headers.append('x-zap-quest-id', zapQuestId || '');
       }
+      // scan page flow
+      if (request.url.startsWith(envConfig.NEXT_PUBLIC_LIFI_API_URL) && request.url.includes('status')) {
```
mmioana: If we're calling `/status` from the zap page, we'll also add this header which might lead to unintended behaviour.

Let's add `const pathname = usePathname();` in the provider and then here add another check
```
if (request.url.startsWith(envConfig.NEXT_PUBLIC_LIFI_API_URL) && request.url.includes('status') && pathname?.includes(AppPaths.Scan)) 
```

mmioana (changes requested):

src/const/urls.ts
```diff
@@ -62,4 +63,5 @@ export enum AppPaths {
   Leaderboard = JUMPER_LEADERBOARD_PATH,
   Earn = JUMPER_EARN_PATH,
   PrivacyPolicy = JUMPER_PRIVACY_POLICY_PATH,
+  Status = JUMPER_STATUS_PATH,
```
mmioana: The `AppPaths` should be reserved for frontend paths only. Can we please add `BackendAppPaths` or sth similar?

mmioana (changes requested):

src/const/urls.ts
admsk2: oh ok, sure

mmioana (approved):

PR #2357, opened 2025-10-23
fix: viewport issues on main widget pages

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-16315

> [!NOTE]
> Ideally you try testing this with a real `iPhone 16 (Pro)`
> Otherwise xcode simulator might come handy as the browser responsive view is not enough to try reproducing the bug

## Testing steps
1.  Go to `/`
2. Try opening different views of the widget and ensure the bottom navigation is not shown on top of the widget screens

### DIFF
--- a/src/app/ui/app/App.tsx
+++ b/src/app/ui/app/App.tsx
@@ -82,8 +82,11 @@ const App = ({ children }: { children: React.ReactNode }) => {
         justifyContent="center"
         alignItems="start"
         sx={{
-          height: '100%',
-          overflow: { xs: 'scroll', sm: 'inherit' },
+          height: welcomeScreenClosed ? '100%' : 'auto',
+          overflow: {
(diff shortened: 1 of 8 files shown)


callMeTheQA:
@mmioana 
thanks I performed a testing on iphone 15 pro Max , where I connect my metamask wallet to jumper app , and then I opened transaction ( from transaction history page on widget) and I noticed scroll does not work as expected so basically you are not able to scroll either down or up. 
not sure if that can be fixed in scope of this PR or you would rather want to have a separate PR/ticket.


https://github.com/user-attachments/assets/cbf9a2a6-2308-4c6c-9edc-fbea4167d428

callMeTheQA (approved):

mmioana:
Closing this in favor of https://github.com/jumperexchange/jumper-exchange/pull/2370

PR #2370, opened 2025-10-29
fix: viewport issues on main widget pages

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-16315

Actually this seems to be the most stable approach to remove extra scrollbars.
Based this over the announcement banner changes as it will impact the layout on the main page

> [!NOTE]
> Ideally you try testing this with a real `iPhone 16 (Pro)`
> Otherwise xcode simulator might come handy as the browser responsive view is not enough to try reproducing the bug

## Testing steps
1.  Go to `/`
2. Try opening different views of the widget and ensure the bottom navigation is not shown on top of the widget screens

### DIFF
--- a/src/components/Widgets/Widget.style.ts
+++ b/src/components/Widgets/Widget.style.ts
@@ -98,6 +98,7 @@ export const WidgetWrapper = styled(Box, {
       },
       '& [id^="widget-app-expanded-container-"], & [id^="widget-scrollable-container-"]':
         {
+          height: '100% !important',
           maxHeight: '100% !important',
         },
     },
--- a/src/components/Widgets/Widget.style.ts
+++ b/src/components/Widgets/Widget.style.ts
@@ -27,9 +27,7 @@ export const WidgetWrapper = styled(Box, {
     prop !== 'contributionDisplayed',
 })<WidgetWrapperProps>(({ theme, autoHeight, contributionDisplayed }) => {
   // autoHeight is used to adapt widget-height automatically instead of default 686px
-  const widgetHeight: 'auto' | number = autoHeight
-    ? 'auto'
-    : DEFAULT_WIDGET_HEIGHT;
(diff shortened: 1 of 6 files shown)


src/components/Widgets/Widget.style.ts
```diff
@@ -93,6 +92,15 @@ export const WidgetWrapper = styled(Box, {
...
     }),
+    [theme.breakpoints.down('sm')]: {
+      '& [id^="widget-relative-container-"]': {
+        maxHeight: '100% !important',
+      },
+      '& [id^="widget-app-expanded-container-"], & [id^="widget-scrollable-container-"]':
+        {
+          maxHeight: '100% !important',
+        },
+    },
```
mmioana, line 104: Not a fan of this approach, but on mobile, given the widget has set the max height for these containers, it will cause to have beside the main page scrollbar, also ome scrollbars for the content underneath the widget title section, which makes it a bit more difficult to get to the desired content

src/app/lib/getAnnouncements.ts
```diff
@@ -0,0 +1,27 @@
...
+import type { StrapiResponse } from '@/types/strapi';
+import type { AnnouncementData } from '@/types/announcement';
+import { getStrapiApiAccessToken } from '@/utils/strapi/strapiHelper';
+
+export async function getAnnouncements() {
+  const urlParams = new AnnouncementStrapiApi().sort('desc');
+  const apiUrl = urlParams.getApiUrl();
+  const accessToken = getStrapiApiAccessToken();
+
+  const res = await fetch(decodeURIComponent(apiUrl), {
```
oktapodia, line 11: Would be good to use a useQuery instead to be consistent across the codebase

oktapodia (approved):

callMeTheQA:
LGTM

callMeTheQA (approved):

PR #2374, opened 2025-10-29
fix: earn latest UI fixes

mmioana:
Contributed https://lifi.atlassian.net/browse/LF-16324

1. Smoothen the scroll to top on filter select
2. Fixed the date labels as per [Figma](https://www.figma.com/design/JUaT4CQ9YSCQ4IWsQQSRIK/Earn--Webpage-?node-id=3625-165247&m=dev)
3. Fixed an issue with the APY filter on all filters clear
4. Added `%` also for APY tooltip

The description length should be fixed [here](https://github.com/jumperexchange/strapi-cms/pull/83)

### DIFF
--- a/src/utils/formatDateLocalized.ts
+++ b/src/utils/formatDateLocalized.ts
@@ -1,0 +1,30 @@
+import { format } from 'date-fns';
+
+/**
+ * Formats a date with localized pattern and ensures days are zero-padded.
+ * Uses date-fns format tokens for localization while guaranteeing consistent day padding.
+ *
+ * @param date - The date to format (Date object, timestamp, or ISO string)
+ * @param formatString - date-fns format string (e.g., 'PP', 'PP p')
+ * @returns Formatted date string with zero-padded days
+ *
+ * @example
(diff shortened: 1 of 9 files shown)


src/utils/formatDateLocalized.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @param formatString - date-fns format string (e.g., 'PP', 'PP p')
+ * @returns Formatted date string with zero-padded days
+ *
+ * @example
+ * formatDateLocalized(new Date('2025-10-03'), 'PP')
+ * // English: "Oct 03, 2025"
+ * // French: "03 oct. 2025"
+ * // German: "03.10.2025"
+ */
+export const formatDateLocalized = (
```
oktapodia, line 17: I think that you can use `dd` to have the padded 0 already, but you then have to build the whole PP yourself :s

oktapodia (approved):

src/utils/formatDateLocalized.ts
mmioana: I actually wanted to avoid that as much as possible 😢

PR #2390, opened 2025-11-05
feat: add data-testid  and test to analytics page

callMeTheQA:
Corresponding JIRA ticket - https://lifi.atlassian.net/browse/LF-16513


src/components/core/charts/LineChart/LineChart.tsx
```diff
@@ -120,6 +122,7 @@ export const LineChart = <V, T extends ChartDataPoint<V>>({
       width="100%"
       height="100%"
       enableCrosshair={enableCrosshair}
+      data-testid={dataTestId || 'analytics-chart'}
```
vinzenzLIFI: Hey @mmioana and @callMeTheQA , I am interested in your opinion here. I am not a fan of undeterministic ids. From what I understand, all linecharts will have the id `analytics-chart` if not provided otherwise. I don't have the full picture of context, but if the testcode relies on the default locator and at some point the linechart is used twice without further specification, then playwright will throw an error, because there are multiple.
Instead could we maybe just forward the props?
```typescript
export const LineChart = <V, T extends ChartDataPoint<V>>({
[other code]
// instead of  'data-testid': dataTestId,
  ...props
}
```
```typescript
  return (
    <StyledResponsiveContainer
      ref={chartContainerRef}
      width="100%"
      height="100%"
      enableCrosshair={enableCrosshair}
      {...props} //Forward here
    >
```

For usage we can simply set the id as we want in the component:
```typescript
      <EarnDetailsAnalyticsLineChartContainer>
        <LineChart
          isLoading={isLoading}
          data={chartData}
          dateFormat={chartDateFormat}
          dataSetId={value}
          dataSetValueAppend={value === AnalyticsValueFieldEnum.APY ? '%' : ''}
          theme={chartTheme}
          data-testid="analytics-chart"
        />
      </EarnDetailsAnalyticsLineChartContainer>
```

src/components/composite/DepositButton/DepositButton.tsx
```diff
@@ -29,12 +29,17 @@ export const DepositButton: FC<DepositButtonProps> = ({
     <DepositButtonLabelWrapper size={size}>{label}</DepositButtonLabelWrapper>
   );
 
+  const dataTestId =
```
vinzenzLIFI: same as the other comment. Maybe I am missing context here, do we really need a default id? The id is forwarded via props and can be set as desired like so:
```typescript
  return (
    <EarnDetailsActionsContainer>
      <DepositFlowButton
        earnOpportunity={earnOpportunity}
        displayMode={DepositButtonDisplayMode.LabelOnly}
        size="large"
        label={t('buttons.depositButtonLabel')}
        data-testid="quick-deposit-button" //set here
      />
      {managePositionSection}
    </EarnDetailsActionsContainer>
```

vinzenzLIFI (changes requested):

src/components/core/charts/LineChart/LineChart.tsx
mmioana: Hey @vinzenzLIFI I agree with dropping the fallback value for the test id! Not necessarily to spread remaining props  ` {...props} //Forward here` as some of them might be needed for other parts withing this component and not the `StyledResponsiveContainer` itself

src/components/composite/DepositButton/DepositButton.tsx
mmioana: I agree, let's not use fallback values!

callMeTheQA: thanks @vinzenzLIFI  @mmioana  I hope I fixed suggested change  on the correct way!

src/components/core/charts/LineChart/LineChart.tsx
callMeTheQA: thanks @vinzenzLIFI  @mmioana  I hope I fixed suggested change  on the correct way!

vinzenzLIFI (approved):
I approve, as it all seems to work, but `quick-deposit-button` and `analytics-chart` is not used in the tests yet. Is this going to be added later? @callMeTheQA

callMeTheQA:
> I approve, as it all seems to work, but `quick-deposit-button` and `analytics-chart` is not used in the tests yet. Is this going to be added later? @callMeTheQA

good catch, yes its going to be used in one of my next PRs where I'm planning to cover that part ! 

Thanks for review.

PR #2400, opened 2025-11-07
feat: differentiate between zap and earn

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-16551

Sister PR https://github.com/jumperexchange/jumper-backend/pull/584

## Summary by CodeRabbit

* **New Features**
  * Widget integrator configuration is now configurable via environment variables for flexible deployment.
  * Added optional market property to widget stack component.

* **Refactor**
  * Optimized component imports and updated widgets to use dynamic configuration instead of hardcoded values.


laurentsenta (approved):
lgtm,

if you want to simplify the PR and maintenance: what if we simply removed them from the env, since it's not really a configuration that relates to different environments, and just use 2 const in the app?

mmioana:
@laurentsenta the integrators we use throughout the app should be env-dependant, as based on the env we're on, different fees can be configured for the retrieved routes 🤔 or do you have a different idea in mind?

laurentsenta:
Ok ok, I'd assume we're using a hardcoded string that we check on the backend and both values have too be equals or you'll end up in the wrong service.

PR #2421, opened 2025-11-17
fix: earn cards blink on page refresh

mmioana:
Contributes to  https://lifi.atlassian.net/browse/LF-16652

## Testing steps
1. Go to `/earn`
2. Ensure at least one wallet is connected
3. Refresh the page
4. Observe there is no intermediate loading state between initial data and wallet-dependent data
5. Now disconnect all wallets
6. Refresh the page
7. Ensure there is a single loading state shown

## Summary by CodeRabbit

* **Refactor**
  * Preserve previous opportunity results during refreshes and memoize filtering context to reduce re-renders and stabilize derived loading state.
* **Bug Fixes**
  * Ensure stable list keys to prevent item flicker or duplication when identifiers are missing or repeated.
* **Style / UX**
  * Improve initial in-view/load behavior, memoize rendered grid items, and refine animation easing for smoother visual transitions.


laurentsenta:
@mmioana I don't think you could reproduce the issue without earn recommendation algorithms plugged in, since without it, the result is always the same, could you try on latest develop? (there's a video in https://lifi.atlassian.net/browse/LF-16652)

oktapodia (approved):

mmioana:
@laurentsenta I've pulled the latest recommendation endpoint changes, if you want to re-check.
The video in the ticket shows the loading state as an intermediate between the 2 data sets, but by using the `placeholderData`, there will be a single loading state at the inital data load. There will be a sort of "blink" as the card contents change, but no other intermediate loading state 😄

mmioana:
Will actually need to save the connected addresses in a cookie to solve this issue, but I'd say to do it in a separate PR. This PR should just improve the visual effect of the animations and memoize some values to prevent extra re-renders

laurentsenta (approved):

PR #2435, opened 2025-11-20
feat: portfolio add empty state for the filtered list

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-16219

## Summary by CodeRabbit

* **New Features**
  * Empty-state UI with illustration, title, description and a "Clear filters" action when portfolio has no items.
  * Filters become disabled when account is disconnected or when portfolio is empty (desktop/tablet empty layouts shown).
  * Single-action "clear all filters" available from filter bar and empty-state view.

* **Bug Fixes**
  * Prevents filter interactions when no portfolio data is available.

* **Documentation**
  * Added translations for the empty-state texts.


src/app/ui/portfolio/PortfolioDeFiProtocolsList.tsx
```diff
@@ -5,23 +5,36 @@ import {
...
   return (
     <>
       <PortfolioAssetsListContainer useFlexGap direction="column">
-        {defiPositions.map((defiPosition) => (
-          <PortfolioAssetContainer key={defiPosition.slug}>
-            <DeFiPositionCard
-              defiPosition={defiPosition}
-              isLoading={isLoading}
-            />
+        {defiPositions.length > 0 ? (
```
laurentsenta, line 24: Do you need that case even with the `if isEmpty return null` ?

src/app/ui/portfolio/PortfolioTokensList.tsx
```diff
@@ -6,21 +6,32 @@ import {
...
+  }
+
   return (
     <PortfolioAssetsListContainer useFlexGap direction="column">
-      {tokens.map((token) => (
-        <PortfolioAssetContainer
-          key={`${token.address}-${token.chain.chainId}`}
-        >
-          <TokenListCard size={TokenListCardTokenSize.MD} token={token} />
+      {tokens.length > 0 ? (
```
laurentsenta, line 22: same question, are we skipping this case earlier with `isEmpty` ?

laurentsenta (approved):

src/app/ui/portfolio/PortfolioDeFiProtocolsList.tsx
mmioana: Ah yes, the idea for this `isEmpty` is to be used for checking if all data (unfiltered data) is empty, so in that case we need to show placeholder disabled filters and no lists. Maybe will make sense to refactor and rename this flag 🤔

mmioana: Will handle this in a subsequent PR

mmioana: The `defiPositions` and `tokens` will be the filtered data btw

PR #2462, opened 2025-11-28
refactor: simplify extract portfolio filters

mmioana:
## Summary by CodeRabbit

* **Bug Fixes**
  * Clearing filters now consistently emits nulls for empty selections and ranges.
  * Improved filter synchronization across portfolio and earn views to reduce stale/unsynced state.

* **Refactor**
  * Reworked portfolio/token/DeFi filter handling and aggregation for clearer value/APY ranges and more reliable token/asset lists.

### DIFF
--- a/src/app/ui/portfolio/utils.ts
+++ b/src/app/ui/portfolio/utils.ts
@@ -216,6 +216,13 @@ export const extractDeFiPositionsFilteringParams = (
 
   const allAssets = uniqBy(allTokens, 'name');
 
+  const apyValues = map(
+    allPositions,
+    (position) => position.latest?.apy?.total || 0,
+  );
+  const minAPY = apyValues.length > 0 ? (min(apyValues) ?? 0) : 0;
+  const maxAPY = apyValues.length > 0 ? (max(apyValues) ?? 0) : 0;
+
   const values = map(
     allPositions,
     (position) => position.netUsd || position.assetUsd || 0,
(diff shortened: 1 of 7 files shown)


mmioana:
@coderabbit review

src/app/ui/portfolio/utils.ts
```diff
@@ -181,66 +176,50 @@ export const extractDeFiPositionsFilteringParams = (
...
+  );
+  const minAPY = apyValues.length > 0 ? (min(apyValues) ?? 0) : 0;
+  const maxAPY = apyValues.length > 0 ? (max(apyValues) ?? 0) : 0;
+
+  const values = map(
+    allPositions,
+    (position) => position.netUsd || position.assetUsd || 0,
+  );
+  const minValue = values.length > 0 ? (min(values) ?? 0) : 0;
+  const maxValue = values.length > 0 ? (max(values) ?? 0) : 0;
```
laurentsenta, line 231: For what it's worth, at some point I think we'll have to default the min value to something like 1$ to filter out dust. I'll wait for product to bring this up, because that brings the question: are we showing a "filter dust" option in UI, or just filtering without letting a choice to the user.

laurentsenta (approved):

src/app/ui/portfolio/utils.ts
mmioana: That makes total sense, let's wait for a product decision and then we can update the minimum value 🙌

PR #2469, opened 2025-12-02
feat: add partner theme to menu

mmioana:
- [x] Need to validate whether we keep the chain selection for showing up a theme

## Summary by CodeRabbit

* **New Features**
  * Partner themes added to the theme selector with avatar/icon display and mobile badge fallback.
  * Partner-theme selection tracking and clearer submenu presentation.

* **Improvements**
  * Partner-theme images now show only when both theme and page conditions are met.
  * Per-theme visibility with expiration-aware logic and page filtering.
  * Persistent per-theme state with migrations and recovery for reliable theme availability and faster menu updates.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "react-i18next": "^16.3.2",
     "recharts": "^3.4.1",
     "sharp": "^0.34.5",
+    "superjson": "^2.2.6",
     "swiper": "12.0.3",
     "universal-cookie": "^8.0.1",
     "uuid": "^13.0.0",
--- a/src/stores/theme/createThemeStore.tsx
+++ b/src/stores/theme/createThemeStore.tsx
@@ -190,22 +190,21 @@ export const createThemeStore = (props: ThemeProps) =>
           widgetTheme: state.widgetTheme,
           configThemeStates: state.configThemeStates,
         }),
-        onRehydrateStorage: () => {
-          return (state) => {
-            if (!state) {
-              return;
-            }
+        merge: (persistedState, currentState) => {
+          const persisted = (persistedState || {}) as PersistedThemeState;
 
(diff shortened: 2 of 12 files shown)


mmioana:
@coderabbit review

package.json
```diff
@@ -70,6 +70,7 @@
     "react-i18next": "^16.3.2",
     "recharts": "^3.4.1",
     "sharp": "^0.34.5",
+    "superjson": "^2.2.6",
```
oktapodia, line 73: Did not know about this one but this is a banger!

src/stores/theme/createThemeStore.tsx
```diff
@@ -1,67 +1,213 @@
...
+          return getOrCreateConfigThemeState(get().configThemeStates, uid);
+        },
       }),
       {
         name: 'jumper-theme-store',
-        version: 1,
-        migrate: (persistedState: any, version: number) => {
+        version: 2,
+        storage: {
+          getItem: (name) => {
```
oktapodia, line 124: I might be wrong but this should not be needed as it should be internally managed by zustand (getItem/setItem) within `      storage: createJSONStorage(() => sessionStorage), // (optional) by default, 'localStorage' is used
`

oktapodia (changes requested):
Couple of comments

src/stores/theme/createThemeStore.tsx
mmioana: Mmm the problem was that using `createJSONStorage` doesn't preserve the URL/Date objects and I wanted to avoid implementing our own serializer/deserializer

package.json
mmioana: @laurentsenta made me aware of it ✨

oktapodia (approved):

PR #2470, opened 2025-12-02
fix: earn page update y axis values on the chart

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-16829

Issues solved
- Updated the y value trange used in the chart
- Put % on Y-AXIS of APY chart
- Put $ on Y-AXIS of TVL chart
- Updated line chart skeleton color on dark mode

## Summary by CodeRabbit

* **Bug Fixes**
  * Value formatting now shows "0" for zero or invalid inputs instead of blank.
* **Enhancements**
  * Improved Y-axis range calculations with safer edge-case handling and more robust min/max offsets.
  * Chart skeleton now respects system light/dark mode when choosing fill colors.
* **Tests**
  * Added unit tests covering Y-axis range behavior across edge cases.


src/components/core/charts/LineChart/utils.ts
```diff
@@ -90,6 +91,64 @@ export const calculateVisibleYRange = <
...
+ * - Max value: dataMaxValue * 1.2 when positive (20% headroom), or 0 when all values are negative
+ * - When all values are identical, ensures a reasonable range (min -1 to 0 for negatives, 0 to 1 for small positives)
+ */
+export const calculateVisibleYRange = <
+  V extends number | string,
+  T extends ChartDataPoint<V>,
+>(
+  data: T[],
+) => {
+  const values = data.map((d) => Number(d.value)).filter((v) => !isNaN(v));
```
laurentsenta, line 106: Can you use `Number` if you're processing TVL, which are big int encoded as string, because they don't always fit on floats?

src/components/core/charts/LineChart/utils.ts
```diff
@@ -90,6 +91,64 @@ export const calculateVisibleYRange = <
...
+      isSymmetricRange: false,
+      isNegative: false,
+    };
+  }
+
+  const dataMinValue = Math.min(...values);
+  const dataMaxValue = Math.max(...values);
+
+  const isNegative = dataMinValue < 0;
+  const isSame = dataMinValue === dataMaxValue;
```
laurentsenta: Do you need an epsilon here?


```suggestion
  const isSame = Math.abs(dataMinValue - dataMaxValue) < Number.EPSILON // or even a larger value like 0.1
```

src/components/core/charts/LineChart/utils.ts
```diff
@@ -90,6 +91,64 @@ export const calculateVisibleYRange = <
...
+      if (Math.abs(dataMinValue) < 1) {
+        minValue = -1;
+      }
+    } else {
+      minValue = 0;
+      if (dataMaxValue < 1) {
+        maxValue = 1;
+      }
+    }
+  }
```
laurentsenta, line 134: You got 5 different branches here, it's pretty complex just for the case where there's a single value in the array (`isSame => `dataMinValue === dataMaxValue`).

Do you have an example of the chart rendered by this case?

mmioana: You're right, I've simplified this. Right now I no longer have an example where this is happening, but when initially implemented this helper function there were a few opportunities where the min and max where the same.

laurentsenta (approved):

PR #2471, opened 2025-12-02
fix: capitalize chain and protocol

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-16919

## Summary by CodeRabbit

* **Style**
  * Applied consistent capitalization formatting to display labels and titles throughout the interface for improved readability and visual consistency.

### DIFF
--- a/src/hooks/earn/useFormatDisplayEarnOpportunityData.tsx
+++ b/src/hooks/earn/useFormatDisplayEarnOpportunityData.tsx
@@ -170,7 +170,7 @@ const buildProtocolItem = (
     key: 'protocol',
     dataTestId: `protocol-${protocolValue}`,
     label: t('labels.protocol'),
-    value: protocolValue,
+    value: capitalizeString(protocolValue),
     tooltip: t('tooltips.protocol'),
     valuePrepend: (
       <EntityChainStack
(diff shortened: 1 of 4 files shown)


src/hooks/earn/useFormatDisplayEarnOpportunityData.tsx
```diff
@@ -170,7 +170,7 @@ const buildProtocolItem = (
     key: 'protocol',
     dataTestId: `protocol-${protocolValue}`,
     label: t('labels.protocol'),
-    value: protocolValue,
+    value: capitalizeString(protocolValue),
     tooltip: t('tooltips.protocol'),
     valuePrepend: (
       <EntityChainStack
```
laurentsenta, line 176: @mmioana I imagine `capitalizeString` will break as soon as you need uppercase anything that is not just the first letter:

Take `Upshift Finance` which is a protocol we'll release with on day 1: https://develop.jumper.exchange/earn?forYou=false&protocols=upshift+finance

also it's not rare to find names like `FUSE` or `Polygon zkEVM` on the widget.

laurentsenta: for what it's worth, a bit related: https://lifi.atlassian.net/browse/LF-16897?search_id=2240a590-9ea0-45d0-8a63-e91d82c0db51

I think protocols won't be straightforward, but is there something in the SDK we could use to go from chain id to chain name?

laurentsenta: Let's use capitalize strings on protocol names, that's what vaults does on their app:

https://app.vaults.fyi/


Could you check with chains? So we're consistent with the rest of the app

laurentsenta (approved):
(chain name done later, talked about it during sync)

PR #2481, opened 2025-12-04
Theme change on scan Page

Kayanski:
Closes https://lifi.atlassian.net/browse/JUM-161.

The theme storage hook was not refreshing as the theme was changed because of the Ref logic. 
Pushed the theme storage behind a useMemo hook to make it variable. 

Not sure about the ThemeConditionMet logic, I need to understand what is the expected UX (common theme, theme changes...)


Kayanski:
Closing, fixed by https://github.com/jumperexchange/jumper-exchange/commit/fe7f76c088a7876e07f64839e198ced80d08f962

PR #2483, opened 2025-12-04
fix: sort by desc value on portfolio

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-16959

## Release Notes

* **New Features**
  * Added configurable sorting direction (ascending/descending) for portfolio items and DeFi positions to allow custom ordering preferences.

* **Removed Features**
  * Removed APY (Annual Percentage Yield) filtering from the DeFi portfolio filter bar, simplifying available filter options.

### DIFF
--- a/src/app/ui/portfolio/utils.ts
+++ b/src/app/ui/portfolio/utils.ts
@@ -6,7 +6,7 @@ import {
   parseAsString,
   parseAsStringEnum,
 } from 'nuqs';
-import { map, uniq, uniqBy, flatMap, min, max, sortBy } from 'lodash';
+import { map, uniq, uniqBy, flatMap, min, max, sortBy, sumBy } from 'lodash';
 import type { CacheToken } from 'src/types/portfolio';
 import type {
   PortfolioTokensFilteringParams,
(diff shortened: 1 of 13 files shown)


src/app/ui/portfolio/utils.ts
```diff
@@ -6,23 +6,68 @@ import {
...
+  getAsset: (item: T) => string;
+}
+
+export const sortPortfolioItems = <T>(
+  items: T[],
+  sortByValue: SortByEnum,
+  order: OrderEnum,
+  accessors: SortAccessors<T>,
+): T[] => {
+  let sorted = [...items];
```
laurentsenta: fwiw sortBy creates an array, so you don't need the spreading

src/app/ui/portfolio/utils.ts
```diff
@@ -6,23 +6,68 @@ import {
...
+
+  if (sortByValue === SortByOptions.VALUE) {
+    sorted = sortBy(sorted, accessors.getValue);
+  } else if (sortByValue === SortByOptions.CHAIN) {
+    sorted = sortBy(sorted, accessors.getChain);
+  } else if (sortByValue === SortByOptions.ASSET) {
+    sorted = sortBy(sorted, accessors.getAsset);
+  }
```
laurentsenta: It's a bit convoluted, you need to create a structure which maps a name to a sort by option, then you write an if which maps a name to a sort by option, drop one level and make it typesafe.

Something like:

```ts
type SortAccessor<T> = Record<SortBy, (key: T) => string | number>

...

const sorted = sortBy(items, accessors[sortByValue]);

if (order == Desc) {
  return sorted.reverse();
}

return sorted

...

export const tokenSortAccessors: SortAccessors<CacheToken> = {
  SortByOptions.VALUE: (token) => token.cumulatedTotalUSD ?? token.totalPriceUSD ?? 0,
  SortByOptions.CHAIN: (token) => token.chainName ?? '',
  SortByOptions.ASSET: (token) => token.name ?? '',
};
```

Benefit: compiler will tell you when there are missing options, simpler code.

laurentsenta (approved):
requested simplif sorter

src/app/ui/portfolio/utils.ts
laurentsenta: fwiw, on backend we do something similar, with a switch / case + typecheck on never to get type safety: https://github.com/jumperexchange/jumper-backend/blob/38e98ae06e43ab26e23d00856d94426fe9901dda/src/earn/filtering.ts#L144-L148

Moving around a piece of data will be nicer here since we already have 2 instances of the same sorter.

PR #2525, opened 2025-12-15
fix: set dynamic flag for newsletter confirmation page

mmioana:
## Summary by CodeRabbit

* **Chores**
  * Updated the newsletter confirmation page to ensure dynamic rendering, guaranteeing fresh data retrieval on each visit rather than serving cached versions.


chmanie (approved):
Yes, I think this is an appropriate fix!

PR #2528, opened 2025-12-15
chore: logs for newsletter confirmation

mmioana:


mmioana:
Opened this only for debugging

PR #2540, opened 2025-12-17
fix: remove all buttons beside withdraw and deposit from defi position card

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-17101

## Summary by CodeRabbit

* **Bug Fixes**
  * Action buttons (claim, compound, repay, borrow) are no longer rendered in DeFi position cards.
* **Refactor**
  * Table layout updated to conditionally show an actions column or an empty placeholder based on mobile view and position type to preserve alignment.


src/components/composite/DeFiPositionCard/hooks.ts
Kayanski: position?.earn might change so you keep the position condition here as well ? You might want to add that to the if condition directly ?

mmioana: I wanted to keep the empty column on desktop no matter if there is the `position?.earn`, but actually this is a good point so I'll update a bit the logic to get the expected layout @Kayanski

src/components/composite/DeFiPositionCard/hooks.ts
```diff
@@ -180,11 +197,15 @@ export const useColumnDefinitions = (
       createEmptyColumn<DefiToken>('empty-column', {
         size: GRID_SIZES.apyColumn,
       }),
-      {
+    ];
+
+    // Note: The position?.earn condition might change
```
laurentsenta, line 202: what might change?

src/components/composite/DeFiPositionCard/utils.tsx
```diff
@@ -174,12 +170,7 @@ export const renderRewardActions = ({
     }}
     useFlexGap
   >
-    <StyledButtonAlphaDark fullWidth={isMobile}>
-      {t('portfolio.defiPositionCard.actions.claim')}
-    </StyledButtonAlphaDark>
-    <StyledButtonPrimary fullWidth={isMobile}>
-      {t('portfolio.defiPositionCard.actions.compound')}
-    </StyledButtonPrimary>
+    {/** TODO: Add claim and compound actions */}
```
laurentsenta: Don't worry about dead code, we'll find it back when we need it

```suggestion
```

src/components/composite/DeFiPositionCard/utils.tsx
```diff
@@ -199,11 +190,6 @@ export const renderBorrowedActions = ({
     }}
     useFlexGap
   >
-    <StyledButtonAlphaDark fullWidth={isMobile}>
-      {t('portfolio.defiPositionCard.actions.repay')}
-    </StyledButtonAlphaDark>
-    <StyledButtonPrimary fullWidth={isMobile}>
-      {t('portfolio.defiPositionCard.actions.borrow')}
-    </StyledButtonPrimary>
+    {/** TODO: Add repay and borrow actions */}
```
laurentsenta:
```suggestion
```

laurentsenta (approved):

src/components/composite/DeFiPositionCard/hooks.ts
mmioana: Will we need to check a position has earn linked to it for claiming/compounding etc.?

mmioana:
I'm going to close this PR as I found a bug with the rendering logic, so I need to update the position mapping as well as the rendering logic

PR #2542, opened 2025-12-18
feat: enable price impact ab testing

mmioana:
Jira: https://lifi.atlassian.net/browse/JUM-202
Sister PR: https://github.com/jumperexchange/jumper-backend/pull/650

## Summary by CodeRabbit

* **New Features**
  * Enabled A/B testing infrastructure to support feature experimentation and user variant assignment
  * Introduced price impact display variant test for widgets to optimize the user experience
  * Added TVL range filtering functionality with configurable minimum and maximum thresholds for enhanced data filtering options

### DIFF
--- a/src/app/lib/getPosthogFeatureFlag.ts
+++ b/src/app/lib/getPosthogFeatureFlag.ts
@@ -1,0 +1,25 @@
+import type { HttpResponse, PosthogFeatureFlag } from '@/types/jumper-backend';
+import { makeClient } from './client';
+
+export type GetPosthogFeatureFlagResult = HttpResponse<
+  PosthogFeatureFlag,
+  unknown
+>;
+
+export const getPosthogFeatureFlag = async (
+  feature: string,
+  distinctId: string,
+): Promise<GetPosthogFeatureFlagResult> => {
+  try {
+    const client = makeClient();
+    const posthogFeatureFlag =
+      await client.v1.postHogControllerGetFeatureFlagV1({
+        key: feature,
+        distinctId: distinctId,
+      });
+    return posthogFeatureFlag;
+  } catch (error) {
+    console.error('getPosthogFeatureFlag failed for feature', feature, error);
+    throw error;
+  }
+};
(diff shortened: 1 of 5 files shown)


src/hooks/useABTest.ts
```diff
@@ -16,34 +19,33 @@ export const useABTest = ({
...
-    gcTime: 15 * 60 * 1000, // 15 minutes
+    staleTime: TEN_MINUTES_MS,
+    gcTime: THIRTY_MINUTES_MS,
   });
 
-  const isEnabled =
-    feature in activeAbTests
-      ? activeAbTests[feature]
-      : (data?.isEnabled ?? false);
+  const isEnabled = feature in activeAbTests ? activeAbTests[feature] : !!data;
```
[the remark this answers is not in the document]
mmioana: At the same time, the value can be a boolean, so checking for the presence of any truthy value ensures that the feature flag is considered enabled. In this case, we also want to check the specific feature flag value to control whether a particular test variant should be displayed. Since the value can hold any string, we shouldn’t limit the check to specific string values only. This approach provides the necessary flexibility to handle both boolean flags and string-based test variants consistently

src/app/lib/getPosthogFeatureFlag.ts
```diff
@@ -0,0 +1,25 @@
...
+  distinctId: string,
+): Promise<GetPosthogFeatureFlagResult> => {
+  try {
+    const client = makeClient();
+    const posthogFeatureFlag =
+      await client.v1.postHogControllerGetFeatureFlagV1({
+        key: feature,
+        distinctId: distinctId,
+      });
+    return posthogFeatureFlag;
```
laurentsenta, line 20: just making sure we checked this, we need to do .data.data sometimes (https://github.com/jumperexchange/jumper-backend/blob/f3b47587484e54aca59f77c102e90f26f2da70e8/src/common/interceptors/transform.interceptor.ts#L127)

mmioana: Yes, noticed we do need to do data.data for this endpoint, so updated the BE response type

callMeTheQA:
LGTM 
Here are the comparasion ( on 1 screenshot price impact appears, while on other does not) 
Screenshot 2025-12-18 at 11 58 58

Screenshot 2025-12-18 at 11 58 28

callMeTheQA (approved):

PR #2560, opened 2025-12-30
feat: enable earn with multi-sig and non-evm

mmioana:
## Summary by CodeRabbit

* **New Features**
  * Enhanced Safe and WalletConnect multisig wallet support
  * Chain switching now functional for multisig wallet users
  * Price impact visibility improvements via testing

* **Improvements**
  * Simplified widget interface with streamlined flow
  * More reliable wallet connectivity detection


src/components/Widgets/variants/base/ZapWidget/ZapWithdrawWidget.tsx
```diff
@@ -126,24 +119,6 @@ export const ZapWithdrawWidget: FC<ZapWithdrawWidgetProps> = ({
 
   const widgetConfig = useWidgetConfig('zap', enhancedCtx);
 
-  if (showZapPlaceholderWidget || !isEvmWallet) {
```
mmioana, line 129: I know I've removed the check for the `isEvmWallet` but we should test we can deposit using a non-EVM wallet 👀

src/components/EarnDetails/EarnDetailsActions.tsx
```diff
@@ -30,9 +31,10 @@ export const EarnDetailsActions = ({
   earnOpportunity,
 }: EarnDetailsActionsProps) => {
   const { t } = useTranslation();
+  const { account } = useAccount();
   const accountAddress = useAccountAddress();
 
-  const isConnected = !!accountAddress;
+  const isConnected = !!account?.address;
```
mmioana, line 37: Do we need both?

chmanie: I'm not sure if we can simplify by a lot. `account.address` is set, even if it's not an EVM address, `accountAddress` is only set when it's an EVM address.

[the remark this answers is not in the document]
chmanie: We don't want to show "Connect wallet" while a wallet (even a non-evm one) is already connected, hence that change. 

We have an error message in place when the user tries to deposit funds.

chmanie (approved):
I approve of the work we've done here

PR #2574, opened 2026-01-06
fix: add missing token symbol to token list card

mmioana:
Jira: https://lifi.atlassian.net/browse/JUM-241 + https://lifi.atlassian.net/browse/JUM-217

### DIFF
--- a/src/components/composite/TokenListCard/TokenStackItem.tsx
+++ b/src/components/composite/TokenListCard/TokenStackItem.tsx
@@ -61,7 +61,7 @@ export const TokenStackItem: FC<TokenStackItemProps> = ({
         title={t('format.currency', { value: token.totalPriceUSD })}
         titleVariant={config.titleVariant}
         hintVariant={config.descriptionVariant}
-        hint={t('format.decimal', { value: token.balance })}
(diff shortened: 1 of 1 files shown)


src/components/composite/TokenListCard/TokenStackItem.tsx
```diff
@@ -61,7 +61,7 @@ export const TokenStackItem: FC<TokenStackItemProps> = ({
         title={t('format.currency', { value: token.totalPriceUSD })}
         titleVariant={config.titleVariant}
         hintVariant={config.descriptionVariant}
-        hint={t('format.decimal', { value: token.balance })}
+        hint={`${t('format.decimal', { value: token.balance })} ${token.symbol}`}
```
chmanie, line 64: I did not know i18next could format things like that! I wonder if we should do that everywhere in the app 🤔. I know of at least a bunch of places where we are using different techniques.

chmanie (approved):
A simple fix. Great!

src/components/composite/TokenListCard/TokenStackItem.tsx
mmioana: Oh totally agree! Maybe we can add a ticket for this?

PR #2588, opened 2026-01-08
fix: filter out empty positions

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-17207

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed portfolio display to filter out empty DeFi protocol groups. Protocol groups without displayable positions are now hidden, showing only active positions.

### DIFF
--- a/src/app/ui/portfolio/PortfolioDeFiProtocolsList.tsx
+++ b/src/app/ui/portfolio/PortfolioDeFiProtocolsList.tsx
@@ -10,6 +10,7 @@ import { DeFiPositionCardSkeleton } from '@/components/composite/DeFiPositionCar
 import { AnimatePresence } from 'motion/react';
 import { useContactSupportEvent } from '@/components/Widgets/events/hooks/useContactSupportEvent';
 import { PortfolioAnimatedAssetContainer } from './PortfolioAnimatedAssetContainer';
+import { hasPositionDataToDisplay } from '@/components/composite/DeFiPositionCard/utils';
 
 export const PortfolioDeFiProtocolsList = () => {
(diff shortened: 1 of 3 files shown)


src/app/ui/portfolio/PortfolioDeFiProtocolsList.tsx
```diff
@@ -32,13 +33,21 @@ export const PortfolioDeFiProtocolsList = () => {
...
 
     if (protocolGroups.length > 0) {
-      return protocolGroups.map((positions, index) => (
-        <PortfolioAnimatedAssetContainer
-          key={`${positions[0].protocol.name}-${positions[0].chain.chainId}-${index}`}
-        >
-          <DeFiPositionCard defiPositions={positions} isLoading={isLoading} />
-        </PortfolioAnimatedAssetContainer>
-      ));
+      return protocolGroups.map((positions, index) => {
```
laurentsenta: That's a lot of data transformation in a react component, shouldn't we implement those in the data layers?

mmioana: I was thinking about it, but it feels like this is more tied to the actual component and what it can display, so I'd vote to keep the filtering here

fredy-QA (approved):

src/app/ui/portfolio/PortfolioDeFiProtocolsList.tsx
mmioana: Moved it actually to the `useFormatDisplayDeFiPositions`

PR #2592, opened 2026-01-08
fix: use position fallback value only if balanceOf retrieves undefined

mmioana:
Jira: https://lifi.atlassian.net/browse/JUM-255

## Summary by CodeRabbit

* **Refactor**
  * Improved deposit amount handling in the earn details view to prevent stale values and ensure the displayed USD amount is derived from the current input.
  * Adjusted logic that determines whether a deposit is present so the UI reflects deposit state more reliably.

### DIFF
--- a/src/components/EarnDetails/EarnDetailsActions.tsx
+++ b/src/components/EarnDetails/EarnDetailsActions.tsx
@@ -49,10 +49,16 @@ export const EarnDetailsActions = ({
       return;
     }
 
-    return positionsData.data[0]?.netUsd;
-  }, [positionsData, isLoadingPositions]);
+    // If the deposit amount is defined, we don't need to use the positions data which might be outdated
+    // The depositAmountUSD will be derived from the deposit amount
+    if (depositAmount !== undefined) {
+      return;
+    }
 
-  const hasDeposited = !!depositAmountUSD || !!depositAmount;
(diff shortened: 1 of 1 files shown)


Kayanski (approved):
Feels a little hacky though

mmioana:
@Kayanski if you want to check it again, as I've changed the approach after getting some more feedback 🙏

src/components/EarnDetails/EarnDetailsActions.tsx
```diff
@@ -49,10 +49,16 @@ export const EarnDetailsActions = ({
...
 
+    // If the deposit amount is defined, we don't need to use the positions data which might be outdated
+    // The depositAmountUSD will be derived from the deposit amount
+    if (depositAmount !== undefined) {
+      return;
+    }
+
     return positionsData.data[0]?.netUsd;
-  }, [positionsData, isLoadingPositions]);
+  }, [depositAmount, positionsData, isLoadingPositions]);
```
laurentsenta, line 59: could you `return undefined` to be explicit about the data we're moving around? That looks like a void function.

laurentsenta (approved):

src/components/EarnDetails/EarnDetailsActions.tsx
mmioana: Done! 🎉

PR #2612, opened 2026-01-14
[WIP] feat: update portfolio component to use portfolio provider

mmioana:


mmioana:
Will close this ftm as it makes sense to start clean with the new approach

PR #2614, opened 2026-01-15
fix: same chain by default on deposit withdrawal

mmioana:
Jira: https://lifi.atlassian.net/browse/LF-17236

## Testing steps

1. Go to `/earn`
2. Select any opportunity
3. Check on deposit the from chain is the opportunity chain
4. Check the from token is the asset token (visible in the overview card)
5. Now for an opportunity into which you have deposited, open the withdraw modal
6. Check the to chain is the same chain as the opportunity chain
7. Go to `/portfolio`
8. Open the deposit/withdraw modals from there and perform steps `2-6`

## Summary by CodeRabbit

* **Refactor**
  * Deposit flow now pre-fills the source token and source chain in deposit settings so forms reflect the selected deposit context.
  * Withdrawal flow now reliably tracks and applies the destination chain, ensuring widget configuration and form data update correctly when the target chain changes.


callMeTheQA (approved):
verified on feature branch.

PR #2658, opened 2026-02-03
fix: env-config.js content-type

Kayanski:

https://lifi.atlassian.net/browse/JUM-256

Because of how mini-apps work, we need to be able to update the base:app_id (to prove ownership) and the accountAttestation metadata to prove the metadata are valid. We can't rely on github deployments solely to update those values. Using strapi data allows anyone with a base.dev access to update the metadata of the mini-app. Which makes it valid for develop, staging and production.


Kayanski:
Wrong branch pushed

PR #2676, opened 2026-02-06
feat: enable slack notificaton

callMeTheQA:
https://lifi.atlassian.net/browse/JUM-526 

## Summary by CodeRabbit

* **New Features**
  * Language menu now displays 15 items (one additional option visible to users).

* **Chores**
  * Added automated Slack notifications for pull request test failures.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -164,3 +164,105 @@ jobs:
             ${{ needs.merge-reports.outputs.summary }}
 
             ${{ needs.create-qase-run.outputs.public_link != '' && needs.create-qase-run.outputs.public_link != null && format('📋 [View Detailed Qase Report]({0})', needs.create-qase-run.outputs.public_link) || '' }}
+
+  notify-slack-on-failure:
+    needs: [create-qase-run, test, merge-reports]
+    if: ${{ failure() && github.event_name == 'pull_request' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Build Slack payload
+        id: build-payload
+        run: |
+          PR_AUTHOR="${{ github.event.pull_request.user.login }}"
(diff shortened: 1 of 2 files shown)


Kayanski:
Very nice ! Have you tested it already ? Is it actually sending notifications to Slack ?
You could introduce a test error (invert a test for instance) to check

Kayanski (approved):
Very nice ! Don't forget to update the branch before merging :)

vinzenzLIFI:
Very cool! 
If the necessary changes for the security risks are minimal, it's worth to do them as a defensive measurement. 

Improvement idea (that does not need to be done now): I am wondering, if we can add a label or such, to deactivate the slack notifications. Example: someone wants to open a draft PR or on the actual PR something is not working and it needs to be  fixed in multiple iterations. Then we might not want to have a slack notification on every commit that is done, where the tests will continue to fail. Maybe it would be enough to turn it off during DRAFT.

callMeTheQA:
> Very nice ! Have you tested it already ? Is it actually sending notifications to Slack ? You could introduce a test error (invert a test for instance) to check

hey yes its sending a slack notification but only for failing CI/CD not for "green" one

callMeTheQA:
> Very cool! If the necessary changes for the security risks are minimal, it's worth to do them as a defensive measurement.
> 
> Improvement idea (that does not need to be done now): I am wondering, if we can add a label or such, to deactivate the slack notifications. Example: someone wants to open a draft PR or on the actual PR something is not working and it needs to be fixed in multiple iterations. Then we might not want to have a slack notification on every commit that is done, where the tests will continue to fail. Maybe it would be enough to turn it off during DRAFT.

good suggestion , let me open a dedicated ticket for that and will implement it soon , thank you 🙏

PR #2696, opened 2026-02-13
fix: balance range filter progressively update

mmioana:

Contributes to https://lifi.atlassian.net/browse/LF-17199

## Testing steps

1. Connect a wallet with small balances (below $1 would be ideal), notice the the balanceMinValue
2. Connect another wallet type with a very large balance, go to /profile (or just remove all filtering params and refresh the page)
3. Notice how the balanceMinValue progressively updates
4. Now disconnect the wallet with large balance, the balanceMinValue should be updated to match the other wallet balances


Kayanski:
Very nice, 
I'm always wondering why we are setting this filter by default. 
Why can't we just filter the tokens under 1$ and that's it by default ? 
is it for wallets with very high asset values for which the dust is bigger ?

Kayanski (approved):

mmioana:
@Kayanski we use this logic so that users that have larger amounts do not see the dust by default, while users who only have dust tokens still see tokens in their list. Also we clamp the filter value that can be set through the url params to the available amount range in order to avoid tokens not being shown in the list only because an out-of-bounds filter was used

PR #2699, opened 2026-02-18
feat: add supporting jumper widget

mmioana:

Contributes to https://lifi.atlassian.net/browse/JUM-520 and https://lifi.atlassian.net/browse/JUM-521

## Testing steps
1. Open storybook
2. Search for `JumperWidget`
3. Play with the different dummy variants
4. Switch between mobile and desktop and ensure the widget behaves (visually) as expected


src/components/composite/JumperWidget/JumperWidget.stories.tsx
```diff
@@ -0,0 +1,446 @@
...
+  );
+};
+
+export const Default: Story = {
+  render: () => {
+    const [sheetOpen, setSheetOpen] = useState(false);
+    const [sheetMode, setSheetMode] = useState<'confirmation' | 'success'>(
+      'confirmation',
+    );
+    const nextStepRef = useRef<((id: string) => void) | null>(null);
```
Kayanski: This element could be included in a useJumperWidgetHook, what do you think ? It seems like it's going to be used a lot.
Even the sheetOpen, sheetMode variables ?

src/components/composite/JumperWidget/JumperWidget.stories.tsx
```diff
@@ -0,0 +1,446 @@
...
+
+    const fromChain = defineField({
```
Kayanski: Could we add helpers, like defineChainField, defineTokenField, defineAmountField ?
This clarified the flow. I see a fieldKey collision danger, but we could enforce setting those for every field

src/components/composite/JumperWidget/components/AmountThreshold.tsx
```diff
@@ -0,0 +1,53 @@
...
+import Box from '@mui/material/Box';
+import { FieldWrapper, Label } from '../JumperWidget.style';
+
+export const amountThresholdSchema = z.object({
+  amount: z.number().min(0),
+});
+
+export type AmountThresholdValue = z.infer<typeof amountThresholdSchema>;
+
+export interface AmountThresholdFieldProps extends BaseFieldProps {
```
Kayanski: Sounds more general than an AmountThreshold. More like a general Threshold (or even just a percentage select, or even a number select)

Kayanski: Or even a select !

Kayanski (approved):
Wonderful, love those Code patterns. Don't forget to address my comments :)

src/components/composite/JumperWidget/JumperWidget.stories.tsx
mmioana: The `goToView` is included in the `useWidgetNavigation`, but given this is a story it would've meant to maybe create another component to make use of it, and for the sake of simplicity decided to go with a ref 😬 

Regarding the sheet, it depends. I decided to go with a very generic sheet config passed to the widget, so any instance of the widget can decide what and when to show this sheet

src/components/composite/JumperWidget/components/AmountThreshold.tsx
mmioana: Oh yes, we could make it a bit more generic 🥇

src/components/composite/JumperWidget/components/Shared.tsx
```diff
@@ -0,0 +1,89 @@
...
+                  <CheckIcon
+                    sx={{ marginLeft: 'auto', height: 24, width: 24 }}
+                  />
+                )}
+              </MenuItemWrapper>
+            );
+          })
+        ) : (
+          <Label>
+            No {header?.toLowerCase() || 'items'} available for selection
```
oktapodia: Would be good to use translation

src/components/composite/JumperWidget/components/Token.tsx
```diff
@@ -0,0 +1,239 @@
...
+}
+
+export const createTokenMultiSelectSchema = (
+  options: TokenMultiSelectSchemaOptions = {},
+) => {
+  const { min = 1, max, allowedAddresses } = options;
+
+  let selectedTokensSchema = z
+    .array(z.string())
+    .min(min, `Please select at least ${min} token${min === 1 ? '' : 's'}`);
```
oktapodia: Would be good to use translation

oktapodia (commented):
Would be good to use translation for most of the texts

PR #2712, opened 2026-03-03
feat: add support for both svm and evm wallets on earn

mmioana:

Closes https://linear.app/lifi-linear/issue/JUM-234/integrate-connect-solana-wallet-prompt-and-logic

## Summary by CodeRabbit

* **New Features**
  * Deposit/Withdraw flows now show connection-aware placeholders when wallet not connected; full widget available when connected.
  * Earn cards now visually indicate wallet connection state (reduced opacity when disconnected).

* **Improvements**
  * Unified chain/connection data sharing across widgets and hooks for more consistent behavior and broader chain support.
  * Removed wallet-type gating so widgets render uniformly when supported.

* **Documentation**
  * Added translations for deposit, withdraw, and unsupported-chain placeholder messaging.

* **Tests**
  * Test setup updated to provide query client context and improved account mocks.


src/components/Widgets/variants/base/ZapWidget/ZapDepositBackendWidget.tsx
```diff
@@ -159,15 +154,7 @@ export const ZapDepositBackendWidget: FC<ZapDepositBackendWidgetProps> = ({
...
-  if (!isEvmWallet) {
-    return (
-      <ZapPlaceholderWidget
-        titleKey="widget.zap.placeholder.non-evm.title"
-        descriptionKey="widget.zap.placeholder.non-evm.description"
-        style={widgetConfig.theme?.container}
-      />
-    );
-  }
+  console.log(widgetConfig);
```
laurentsenta: remove

mmioana:
@coderabbitai review

src/components/Widgets/variants/widgetConfig/useMissionWidgetConfig.ts
```diff
@@ -20,6 +20,7 @@ export function useMissionWidgetConfig(
   return useMemo(() => {
     const chainType = account?.chainType;
     const isEvmWallet = chainType === ChainType.EVM;
+    const isSVMWallet = chainType === ChainType.SVM;
```
laurentsenta: Those are mutually exclusive,
this code is turning one value (chainType) into 2 that are mutually exclusive. It's like doubling the complexity required to understand the rest of the code.

Do you actually need this? Or do you need something like:

```ts
const isSupportedType = (type: ChainType): type is ChainType.EVM | ChainType.SVM => {
   // if else... LLM will write this.
}

...

const isSupported = isSupportedType(chainType)

...

isSupported && !isSafe ...
```

src/components/composite/WithdrawModal/WithdrawModal.tsx
```diff
@@ -29,28 +31,54 @@ export const WithdrawModal: FC<WithdrawModalProps> = ({
     (state) => state.refetchCallback,
   );
 
+  const earnDataByChainId = useEarnByChainId(
+    earnOpportunity?.lpToken?.chain?.chainId,
+  );
```
laurentsenta: I think you have an abstraction problem here: we're moving very low-level data structures, and extracting more low-level data structures that we need to transform again and again. You should not need to write `map` and `check?.again?.and?.again`,

Move up your abstraction ladder should make your code clearer and easier to maintain:

```suggestion
  const {allowChain, chainType, isConnected} = useTypedAccount(earnOpportunity)
```

src/hooks/earn/useEarnByChainType.ts
```diff
@@ -0,0 +1,38 @@
...
+      },
+      [ChainType.SVM]: {
+        chains: chains.filter((chain) => chain.chainType === ChainType.SVM),
+        account: connectedAccounts.find(
+          (account) => account.chainType === ChainType.SVM,
+        ),
+      },
+    }),
+    staleTime: THIRTY_MINUTES_MS,
+  });
```
laurentsenta: there's no IO, why is that a useQuery and not a useMemo?

src/hooks/earn/useEarnByChainType.ts
```diff
@@ -0,0 +1,38 @@
+import { ChainType } from '@lifi/sdk';
+import { useChains } from '../useChains';
+import { useMemo } from 'react';
+import { useAccount } from '@lifi/wallet-management';
+import { useQuery } from '@tanstack/react-query';
+import { THIRTY_MINUTES_MS } from '@/const/time';
+
+export const useEarnByChainType = () => {
```
laurentsenta: Don't export, you're increasing your API surface, keep private and move inside the useEarnByChainId.

src/components/composite/WithdrawModal/WithdrawModal.tsx
laurentsenta: That would make it more clear that this is actually a sibling to `useAccount` and not related to earn data.

src/hooks/earn/useEarnByChainId.ts
```diff
@@ -0,0 +1,25 @@
...
+  const chainData = chain
+    ? data?.[chain.chainType as keyof typeof data]
+    : undefined;
+
+  return {
+    account: chainData?.account,
+    chains: chainData?.chains ?? [],
+    chainType: chain?.chainType,
+  };
+};
```
laurentsenta: I find it a bit awkward that we need 2 calls to useChains, a `getChainId` side effect, 3 useMemo and a useQuery to grab an account by chain type.

Feel free to contradict me, but I'd say that makes it hard to review and get into the codebase because there are so many low-level data transformations to be aware of.

A bit like the above comment about moving up the abstraction layer, could we push towards higher-level abstractions?

```ts
export const useTypedAccount(opportunity) {
   const {chain, loading} = useChainDetails(opportunity.asset.chain.chainId)
   const chains = useChainsForChainType(chain?.chainType)
   const account = useAccountForChainType(chain?.chainType)
   
   return {chains, account, loading}
}
```

laurentsenta (changes requested):
left a bunch of comments

src/components/composite/WithdrawModal/WithdrawModal.tsx
mmioana: Given we'd pass the full earn opportunity, it would still be tied to earn data, so using `useTypedAccount` makes the hook name a bit too generic imo. I'd suggest to go with `useChainTypeData` and still pass the chainId we're interested in, thus making the hook easily reusable wherever we might need the retrieved data

src/hooks/earn/useEarnByChainType.ts
mmioana: If you only use `useMemo`, you'd recompute these filtering ops each time you call the hook. Using `useQuery` will make the filtering run once and then it would be only a matter of retrieving the data. I removed it though for the moment, but will add it back in future (once I tackle the page improvements for /earn)

src/components/composite/WithdrawModal/WithdrawModal.tsx
laurentsenta: But passing the earn data is the point: you move one abstraction layer up, from low level integers to actual business data.

Here's a concrete example, say in two weeks we add earn opportunities without an lptoken (like most dialect vaults), you'll have to update all your callers.

If you think about your business model first you'll go from _"I need that integer inside that field inside that field and filter on that number"_ which changes and break all the time to _"I need account details for a earn opportunity"_.

On naming, it's a similar discussion, it sounds like we're starting from the low level implementation. As far as I understand, we `useChainTypeData` first to grab the related accounts & config. It's like calling a function `useOrangeData` because it's getting orange napkins. Just call it `useNapkins`.

laurentsenta (approved):

PR #2714, opened 2026-03-03
feat: add tests for deposit and withdraw button on DeFI tab

callMeTheQA:
https://lifi.atlassian.net/browse/LF-17569 

### ** Why did I implement it this way?**
Page object (PortfolioPage):
 - Added locators for the first deposit/withdraw buttons (getByTestId(...).first()) because there are many position cards with the same test ids; we only need to interact with the first visible one in this flow.
 - Used a dedicated position card locator (Gearbox) to expand a DeFi position before asserting and clicking deposit/withdraw, so the test runs against a concrete card state.
 - Modal checks use text-based locators for “Quick deposit” and “Withdraw” because those titles are rendered inside the LiFi widget and don’t have data-testids we can set.
 - Close button uses getByTestId('CloseIcon') to close the modal after verifying the deposit modal.
**Spec flow:**
 - The test goes through: open portfolio → close welcome → switch to DeFi Protocols tab → expand a position card → assert deposit/withdraw buttons are visible → click deposit → assert deposit modal → close modal → click withdraw → assert withdraw modal. This covers the main happy path for DeFi deposit/withdraw from the portfolio page.

### **# Checklist before requesting a review**

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for deposit and withdraw functionality on the DeFi positions tab, including validation of modal interactions and button visibility.

### DIFF
--- a/tests/pages/PortfolioPage.ts
+++ b/tests/pages/PortfolioPage.ts
@@ -22,6 +22,12 @@ export class PortfolioPage {
   private readonly portfolioHeaderOverviewElement: Locator;
   private readonly tokensOverviewElement: Locator;
   private readonly defiPositionsOverviewElement: Locator;
+  public readonly depositButton: Locator;
+  public readonly withdrawButton: Locator;
+  private readonly gearBoxPositionCard: Locator;
+  private readonly depositModalTitle: Locator;
+  private readonly withdrawModalTitle: Locator;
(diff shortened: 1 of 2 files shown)


mmioana (approved):
We'll probably need to add some test ids for the modal header so we don't rely on the xpath. However will approve for now 👍

PR #2734, opened 2026-03-09
fix: add path guard for positions fetching

mmioana:

Closes https://linear.app/lifi-linear/issue/JUM-618/disable-positions-query-when-not-on-portfolio-page

## Testing steps
1. Go to `/`
2. Open the network tab and filter for `/positions`
3. Connect your wallet
4. Check no call to `/positions` endpoint is made
5. Open the wallet management drawer
6. Check again no call to `/positions` endpoint is made
7. Now go to `/portfolio`
8. Check a call to `/positions` is indeed made
9. Now disconnect your wallet
10. Refresh the page
11. Check no call to `/positions` endpoint is made
12. Connect your wallet
13. Call to `/positions` endpoint is made

In order to decrease the number of requests made to `/positions` endpoint, we're going to fetch this data only upon navigating to `/portfolio`. If we'll need this data elsewhere in the future we might change this condition

## Summary by CodeRabbit

* **Refactor**
  * Portfolio position queries are now limited to portfolio routes, so address-based position lookups only run when viewing portfolio-related pages.

### DIFF
--- a/src/providers/PortfolioProvider/hooks/usePositionsData.ts
+++ b/src/providers/PortfolioProvider/hooks/usePositionsData.ts
@@ -10,7 +10,7 @@ import type { EVMAccount } from '@lifi/wallet-management';
 import { useAccount } from '@lifi/wallet-management';
 import { usePortfolioCacheStore } from '@/stores/portfolio/PortfolioCacheStore';
 import { usePathnameWithoutLocale } from '@/hooks/routing/usePathnameWithoutLocale';
-import { isAllowedPositionPath } from '../utils';
+import { isCurrentPageUsingPositionData } from '../utils';
 
 export interface UsePositionsDataProps {
   filter?: Omit<PortfolioPositionsQuery, 'evm'>;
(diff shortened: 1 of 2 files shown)


src/providers/PortfolioProvider/hooks/usePositionsData.ts
```diff
@@ -50,6 +52,8 @@ interface QueryData {
 export const usePositionsData = ({
   filter,
 }: UsePositionsDataProps): UsePositionsDataResult => {
+  const pathname = usePathnameWithoutLocale();
+  const isPathEnabled = isAllowedPositionPath(pathname);
```
laurentsenta: isAllowedPositionPath sounds  auth related,
rename & wrap the method in something like:

```ts
const isEnabled = isCurrentPageUsingPositionData()
```

?

laurentsenta (approved):

laurentsenta (approved):

src/providers/PortfolioProvider/hooks/usePositionsData.ts
mmioana: Fair, I actually like your proposal better 🥇

PR #2780, opened 2026-03-31
fix: blog article highlight correct toc entry on url with hash nav

mmioana:
Closes https://linear.app/lifi-linear/issue/JUM-739/toc-does-not-highlight-correct-entry-on-direct-url-hash-navigation

## Summary by CodeRabbit

* **New Features**
  * Responsive header-height detection to ensure anchors offset correctly across screen sizes
  * Heading elements now adjust their anchor offset responsively for better in-page linking

* **Bug Fixes**
  * Improved anchor-link scrolling and hash-based navigation to correctly identify and scroll to target sections on load
  * More stable page navigation by re-aligning scroll position as content/layout stabilizes (reducing missed targets)

### DIFF
--- a/src/hooks/useActiveTocSectionId.ts
+++ b/src/hooks/useActiveTocSectionId.ts
@@ -1,7 +1,6 @@
 import { useLayoutEffect, useRef, useState } from 'react';
+import { useHeaderHeight } from './useHeaderHeight';
 
-const DEFAULT_SCROLL_OFFSET_PX = 120;
-const EXPECTED_TOP_MAX_PX = 80;
 const LAYOUT_STABILIZE_MS = 500;
 
 const computeActiveId = (
@@ -20,10 +19,12 @@ const computeActiveId = (
 
 export const useActiveTocSectionId = (
   sectionIds: readonly string[],
-  scrollOffsetPx: number = DEFAULT_SCROLL_OFFSET_PX,
+  scrollOffsetPx?: number,
 ): string | undefined => {
   const [activeId, setActiveId] = useState<string | undefined>(undefined);
   const rafId = useRef(0);
+  const headerHeight = useHeaderHeight();
+  const offset = scrollOffsetPx ?? headerHeight;
 
   useLayoutEffect(() => {
     if (sectionIds.length === 0) {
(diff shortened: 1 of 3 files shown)


yordanove (commented):
Tested locally. The TOC highlight on hash navigation works correctly now — correct entry is highlighted on direct URL load.

However, there's a scroll-blocking issue introduced:

**Repro 1 — Direct URL:**
1. Navigate to `/learn/jumper#the-problem-jumper-solves`
2. Try scrolling up immediately
3. Page stutters and pulls you back down

**Repro 2 — TOC clicks:**
1. Navigate to `/learn/jumper` (no hash)
2. Click any TOC entry on the left sidebar
3. Try scrolling up
4. Same stutter — can't scroll up, need to scroll down first

This happens consistently on every TOC click, not just on initial page load.

Could be the `tryAlign()` + `ResizeObserver` in `useActiveTocSectionId` re-scrolling to the target section while the user is trying to scroll away?

https://github.com/user-attachments/assets/c743eea2-a92d-4939-9170-cba8e4978ed3

src/hooks/useActiveTocSectionId.ts
```diff
@@ -18,37 +19,84 @@ const computeActiveId = (
...
+            parseFloat(getComputedStyle(section).scrollMarginTop) || offset;
+          const { top } = section.getBoundingClientRect();
+          if (top < 0 || top > scrollMarginTop) {
+            section.scrollIntoView({ behavior: 'smooth', block: 'start' });
+          }
+        };
+
+        tryAlign();
+
+        // Re-align whenever the document layout shifts (images/fonts loading).
```
Kayanski: Not sure about this section. 
When I scroll the page after reading the anchor, I get redirected to the original anchor interrupting my read, the tryALign doesn't seem to be removed correctly

mmioana: Yup, it was an issue there, so decided to go back to the old plain interval and make use of refs

Kayanski (approved):

yordanove (approved):
Scroll-blocking regression from last review is **fixed** ✅ — tested on `/learn/jumper`, can scroll freely after hash navigation.

Two minor remaining items (not blockers):

1. **Offset mismatch** (CodeRabbit's point) — hash alignment uses `headerHeightPx + 10` but `computeActiveId` uses `scrollOffsetPx` (default 120px). Can cause TOC to highlight the wrong entry occasionally.

2. **Same-hash re-click no-op** — if you click a TOC entry, scroll away, then click the same entry again, the page doesn't scroll back (hash is already set, browser treats it as no-op). Pre-existing behavior, but more noticeable now. 

Neither blocks merge — the core fix works correctly.

PR #2810, opened 2026-04-14
feat: add portfolio holdings view and filtering

mmioana:

Closes https://linear.app/lifi-linear/issue/JUM-802/portfolio-page-refactor-tokens-and-defi-protocols-on-single-tabbed

- [x] Need to check if we want to show the disabled tabs for Performance/Transactions

## Summary by CodeRabbit

* **New Features**
  * Tabbed portfolio view: Holdings, Performance, Transactions (Performance/Transactions marked “Soon”).
  * Consolidated Holdings view combining tokens and grouped positions with expandable sections, localized titles, and loading skeletons.
  * Filter bar now focused on Holdings with in-UI categories and apply/clear/reset controls.

* **Refactor**
  * Unified holdings filtering with URL-sync for filter state.
  * Cleaner token and position summary rows and improved token presentation.
  * Select inputs support leading/trailing adornments.


src/components/composite/BalanceCard/components/TokenSummaryRow.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+export const TokenSummaryRow: FC<TokenSummaryRowProps> = ({
+  balances,
+  config,
+}) => {
+  const isMobile = useMediaQuery((theme) => theme.breakpoints.down('sm'));
+
+  const hasMultipleChains = balances.length > 1;
+  const primaryBalance = balances[0]!;
+  const chainEntities = getUniqueChains(balances);
```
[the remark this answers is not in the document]
mmioana: This should already happen in the provider/higher components

src/app/ui/portfolio/PortfolioPositionsList.tsx
```diff
@@ -1,68 +0,0 @@
...
-  if (isEmpty && !isLoading) {
-    return null;
-  }
-
-  return (
-    <>
-      <PortfolioAssetsListContainer useFlexGap direction="column">
-        <AnimatePresence mode="popLayout">{renderContent()}</AnimatePresence>
-      </PortfolioAssetsListContainer>
-      <DepositFlowModal />
```
Kayanski, line 64: No more flow modals ?

src/components/PortfolioFilterBar/hooks.tsx
```diff
@@ -42,10 +36,12 @@ export const usePortfolioBalancesFilterBar = () => {
     clearFilters,
     sortBy,
     setSortBy,
-  } = useBalancesFiltering();
+  } = useHoldingsFiltering();
   const { getChainById } = useChains();
   const { accounts } = useAccount();
 
+  const isLoading = balancesIsLoading || balancesIsEmpty;
```
Kayanski, line 46: are you sure ?

src/providers/PortfolioProvider/filtering/types.ts
```diff
@@ -19,38 +18,19 @@ export const OrderOptions = {
...
   assets?: string[];
   minValue?: number;
   maxValue?: number;
 }
 
-export interface BalancesFilterUI extends BalancesFilter {}
-
-export interface PositionsFilteringParams {
-  allChains: number[];
-  allProtocols: Protocol[];
```
Kayanski, line 41: No filtering by protocols .?

src/app/ui/portfolio/PortfolioHoldings/PortfolioTokenHoldings.tsx
```diff
@@ -0,0 +1,47 @@
...
+  title,
+}) => {
+  const {
+    balancesData: data,
+    balancesIsLoading: isLoading,
+    balancesIsEmpty: isEmpty,
+  } = useHoldingsFiltering();
+  const { totalPortfolioUsd } = usePortfolioSummary();
+
+  const balanceGroups = Object.entries(data);
```
Kayanski, line 27: Should this be memoized ?

src/components/PortfolioFilterBar/hooks.tsx
```diff
@@ -298,364 +204,39 @@ export const useBalancesFilterCategories = () => {
...
-          options: typeOptions,
-          searchable: true,
-          searchPlaceholder: t('portfolio.filter.search', {
-            filterBy: t('portfolio.filter.type').toLowerCase(),
-          }),
-          testId: 'portfolio-filter-type-select',
-        })
-      : null,
-    assetOptions.length > 1
+    walletOptions.length > 1
```
Kayanski, line 245: wallet is after now ?

Kayanski (approved):

src/components/PortfolioFilterBar/hooks.tsx
mmioana: Same as this line just that now I extracted it in a separate const https://github.com/jumperexchange/jumper-exchange/pull/2810/changes#diff-530f86df46157cc8e9098c50e4806fd5830318a53d2f09cd289700a979a41ea9L175

src/components/PortfolioFilterBar/hooks.tsx
mmioana: Yes, in Figma the order of the filters has changed https://www.figma.com/design/vk3DeaqHF1o7GNHRaV5Xni/Portfolio--Webpage-?node-id=1154-107039&m=dev

src/app/ui/portfolio/PortfolioHoldings/PortfolioTokenHoldings.tsx
mmioana: React compiler should handle this

src/providers/PortfolioProvider/filtering/types.ts
mmioana: Nope, I think the reasoning is that tokens can't be filtered by protocol and given we merge all assets under a single view it doesn't make sense to have it https://www.figma.com/design/vk3DeaqHF1o7GNHRaV5Xni/Portfolio--Webpage-?node-id=1154-107039&m=dev

src/app/ui/portfolio/PortfolioPositionsList.tsx
mmioana: Nice catch! But no, the detail page should handle this once we'll have it in place

src/providers/PortfolioProvider/filtering/HoldingsFilteringContext.tsx
```diff
@@ -0,0 +1,351 @@
...
+      ...sanitized,
+      minValue: hasInitialFilterRangeRef.current
+        ? sanitized.minValue
+        : effectiveValueRange.min,
+    };
+
+    if (!isEqual(withDefaults, filter)) {
+      setFilter(removeNullValuesFromFilter(withDefaults));
+      setSearchParamsState({
+        holdingsWallets: withDefaults.wallets,
```
yordanove: Finding 1 lands here: on a cold load with `?holdingsChains=999` (HyperEVM — a chain I hold a position on, and which this filter's own menu offers), the sanitize pass runs before the chain set is fully loaded and writes the param back as empty `holdingsChains=`. Reproduced 4/4, including reloading a URL this view generated itself via Apply. Prod's `balancesChains` survives reloads.

src/providers/PortfolioProvider/filtering/HoldingsFilteringContext.tsx
```diff
@@ -0,0 +1,351 @@
...
+        order,
+      ),
+    [balancesState.balancesByAddress, filter, sortBy, order],
+  );
+
+  // ─── Filter & sort positions ───────────────────────────────────────────────
+
+  // chains/assets pre-filtered here; value range handled in filterSortPositionsData
+  const processedPositions = useProcessedPositions({
+    filter: {
```
yordanove: Finding 4: the positions filter only receives `chains`/`assets`/`sortBy`/`order` — no `wallets` — while balances get the full filter. Empirically: with MetaMask + Phantom connected and the Wallet filter set to the Phantom (Solana) wallet only, Tokens narrows correctly but DeFi/Perps keep showing the EVM wallet's positions.

src/components/PortfolioFilterBar/hooks.tsx
```diff
@@ -42,10 +36,12 @@ export const usePortfolioBalancesFilterBar = () => {
     clearFilters,
     sortBy,
     setSortBy,
-  } = useBalancesFiltering();
+  } = useHoldingsFiltering();
   const { getChainById } = useChains();
   const { accounts } = useAccount();
```
yordanove, line 45: This conflation makes "empty" indistinguishable from "loading": a wallet with zero token balances keeps `isLoading` true forever, so the filter bar stays disabled permanently (verified with an empty wallet — three $0.00 sections, dead Filters button, no empty state).

yordanove (commented):
**PR-level QA pass on the Vercel preview** — desktop + mobile width, real wallets: a diversified EVM wallet, an empty wallet, a dust-heavy wallet, and an EVM+Solana (MetaMask + Phantom) pair. Also covered: disconnected state, old-URL compat, Convert Dust adjacency, and the app-wide SelectBase change (earn-page selects spot-checked). Where relevant I compared behavior against prod's current portfolio to separate new issues from carried-over ones.

**Works well:** sections are internally consistent on every wallet I tried (row sums match section headers exactly); multichain token rows expand to clean per-chain breakdowns; chain/asset filtering on positions is server-side and works (params land on `/v1/portfolio/positions`); the asset filter even decomposes positions by underlying asset (Ipor $3.67 → $2.68 under a USDC-only filter — nice touch); Convert Dust flow intact; "Soon" tabs properly disabled on desktop and unselectable in the mobile Select; no junk/blank position rows on any wallet.

**Findings:**

1. **Chain filters don't survive a reload — regression vs prod.** Apply a chain filter → URL gets `holdingsChains=999` → reload (or share/bookmark) that URL → the param is stripped back to empty `holdingsChains=` during load (4/4 repros on a fast connection, including URLs the app itself generated via Apply). `holdingsAssets`/`holdingsMinValue`/`holdingsMaxValue` survive — only chains are sanitized away, which points at the URL-sanitize effect validating chains against a not-yet-loaded chain set. On prod the equivalent `balancesChains=1` survives a reload fine. Since filter-state-in-URL is a headline feature of this view, this is the one I'd fix first.

2. **The default $1 min-value filter is now enforced but can't be removed on its own.** Value → Clear → both "Filter and sort" and "Clear all" go disabled; closing the modal discards the change. Only escapes: hand-edit the URL, or bundle the clear with another filter change (apply only enables when the panel is non-empty). The modal mechanics are carried over from the current portfolio (prod behaves the same), but it matters now: prod doesn't actually hide sub-$1 rows from the list, while this view does — so being stuck with the default filter has real consequences here (see 3).

3. **Sections never reconcile with the header total; percentages don't sum to 100%.** Default state: Tokens $108.19 + DeFi $35.79 + Perps $7.88 = $151.86 vs header $153.97; 70.27 + 23.25 + 5.12 = 98.64%. Section numerators use filtered data, the denominator is the unfiltered total. Counter-test with min≈0: sums match the header and percentages hit exactly 100.00%. During load it's briefly worse — Tokens showed 98.49% (balances-only denominator) before settling at 70.27%.

4. **The Wallet filter only scopes the Tokens section.** With MetaMask (EVM) + Phantom (Solana) connected and the filter set to Phantom only: Tokens correctly narrows to the Solana holdings ($5.10), but DeFi ($37.76) and Perps ($7.88) keep showing the EVM wallet's positions — a Solana wallet can't hold those.

5. **Dead affordances.** Every position card and single-chain token row renders with a pointer cursor but does nothing on click — only multichain token rows respond (expand). With deposit/withdraw removed and the detail page not built yet, these shouldn't present as clickable.

**Minor notes:**
- Apply serializes the full filter state into the URL — including an unrequested `holdingsMaxValue=<current top balance>` and empty `holdingsWallets=`/`holdingsAssets=` params, and `minValue=0` gets rewritten to an epsilon (`0.00003492862`). These serialization quirks are pre-existing (prod does the same with the `balances*` params), but worth noting that a bookmarked URL will exclude the top holding once it grows past the recorded max — and in this view the max actually filters.
- Filtered loads fire the positions request twice — once with filter params, once bare (both 200). Didn't check whether prod does the same.
- An empty wallet gets no empty state: three "$0.00 / 0.00%" sections and a permanently disabled Filters button.
- "Perps $0.00 / 0.00%" renders for perps-less views (incl. when chain-filtering excludes all perps).
- Perps cards (Hyperliquid/Lighter) show no chain sub-label while DeFi cards do — flagging in case it's unintentional.
shot-01-default-filter-sections-dont-sum-to-header
shot-02-minvalue-0-percentages-reach-100-and-note-epsilon-url
shot-03-value-clear-leaves-apply-and-clearall-disabled
shot-04a-hyperevm-chain-filter-applied-via-ui
shot-04b-after-reload-chain-filter-silently-lost
shot-05-wallet-filter-phantom-only-but-defi-perps-show-evm

mmioana:
@yordanove here is some feedback based on the review - the rest of the points I will treat 🫶 

> 3. Sections never reconcile with the header total; percentages don't sum to 100%. 
This is due to filters being applied, once they are cleared, the percentages should sum up to 100%

Minor notes:

> An empty wallet gets no empty state: three "$0.00 / 0.00%" sections and a permanently disabled Filters button.

This is the same as in prod currently

> "Perps $0.00 / 0.00%" renders for perps-less views (incl. when chain-filtering excludes all perps).

For the moment don't think we should hide sections from view if there is no data or filters have been applied

> Perps cards (Hyperliquid/Lighter) show no chain sub-label while DeFi cards do — flagging in case it's unintentional.

Perps do not have a chain attached to them

yordanove:
**Retest on `4d9a545` ("fix: changes after testing") — all findings resolved ✅**

Re-ran the full pass on the preview (desktop; real wallets: diversified EVM, an empty wallet, and a MetaMask + Phantom EVM/Solana pair). Every actionable item from my earlier review is addressed, and I didn't spot any regressions from the fix.

1. **Chain filter survives reload** ✅ — applied `holdingsChains=999` (HyperEVM), reloaded the URL the view generates, and the param + filter now persist (narrows to HyperEVM). The sanitize pass waits for the chain set before validating, so valid chains are no longer stripped on cold load.
2. **Default $1 min removable on its own** ✅ — Value → Clear now enables "Filter and sort"; applying it drops the default min and brings back sub-$1 rows. No longer stuck needing a URL edit or a bundled change.
3. **Sections reconcile / percentages** ✅ — agreed it's the filter artifact: with the default $1 in place the numerator is filtered against the unfiltered total (~98%); once the min is cleared, sections sum to the header and percentages hit 100.00%. Reachable now thanks to #2.
4. **Wallet filter scopes positions** ✅ — filtering to the Phantom (Solana) wallet narrows Tokens to the Solana holdings *and* takes DeFi/Perps to $0.00; they no longer keep showing the EVM wallet's positions.
5. **Dead affordances** ✅ — position cards now `cursor:default`. (Token rows, incl. single-chain, are block-explorer links — pointer is correct there.)
- **Empty wallet** ✅ — now settles to a stable empty state (no skeletons/stuck spinners); the Filters bar being disabled is the genuine "nothing to filter" state, not the old loading-conflation. (No dedicated empty-state copy, but that's the existing/by-design behavior.)

Rebuffs all hold up: perps having no chain sub-label and $0.00 sections rendering are by-design as you noted, and the URL serialization now omits the default-min / portfolio-max / empty-array params.

Didn't re-verify the minor "positions request fires twice on filtered loads" note — flagging only so it's not assumed checked.

QA-wise this looks good to merge. 🚀

yordanove (approved):

PR #2828, opened 2026-04-24
chore: update sentry config

mmioana:

## Release Notes

* **New Features**
  * Wallet activity is now automatically tracked for analytics purposes.
  * Console messages can now be sent to error monitoring.

* **Bug Fixes**
  * Error reporting is now consistently enabled across all environments.

* **Documentation**
  * New guide available for configuring error monitoring, performance tracing, replay features, and related settings.


src/app/[lng]/error.tsx
```diff
@@ -17,10 +15,7 @@ export default function Error({
   reset: () => void;
 }) {
   useEffect(() => {
-    if (isProduction) {
```
Comment (chmanie) [inline src/app/[lng]/error.tsx:20]: Does this mean we also capture on `develop` and `staging` as well?

src/config/env-config.ts
```diff
@@ -68,6 +68,7 @@ const clientBuildTimeEnv: RuntimeConfig = {
     process.env.NEXT_PUBLIC_VERCEL_BRANCH_URL ?? '',
   NEXT_PUBLIC_NOTIFICATIONS_URL:
     process.env.NEXT_PUBLIC_NOTIFICATIONS_URL ?? '',
+  NEXT_PUBLIC_SENTRY_DSN: process.env.NEXT_PUBLIC_SENTRY_DSN ?? '',
```
chmanie, line 71: Is it ok to expose this?

src/sentry/sharedOptions.ts
chmanie, line 1: Is this really shared? I want to prepare a package that we can share between the backend and the notification system. I am expecting to also integrate Sentry there as well at some point. Maybe we can move that file there as well?

src/sentry/sharedOptions.ts
```diff
@@ -0,0 +1,179 @@
...
+  runtime: SentryRuntime,
+  options?: { isProd: boolean },
+): SentryBaseOptions {
+  const prod = options?.isProd ?? isProduction;
+  const dsn = process.env.NEXT_PUBLIC_SENTRY_DSN;
+  const off = isStorybookOrTest();
+  const isClient = runtime === 'client';
+  const tracesSampleRate = getTracesSampleRate(isClient, prod);
+
+  const enableLogs = prod
```
chmanie, line 133: Ah it moved here, I see!

sentry.edge.config.ts
```diff
@@ -1,30 +1,22 @@
...
-  dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
-  environment: process.env.NEXT_PUBLIC_ENVIRONMENT || 'development',
-
-  // Adjust this value in production, or use tracesSampler for greater control
-  tracesSampleRate: isProduction ? 0.4 : 1.0,
+const base = getSentryBaseOptions('edge');
 
-  // Setting this option to true will print useful information to the console while you're setting up Sentry.
-  debug: !isProduction,
+init({
```
chmanie: 🤔 Hmm, can't this just be `init(base)`? Or is `init` modifying the `base` object?

chmanie (commented):
Seems good, let's try this one! Few tiny comments only!

src/config/env-config.ts
oktapodia: Yes, that's origin blocked and the common way of setting it up

src/sentry/sharedOptions.ts
```diff
@@ -0,0 +1,179 @@
...
+          PROD_REPLAY_SESSION,
+        )
+      : parseSampleRate(
+          process.env.NEXT_PUBLIC_SENTRY_REPLAY_SESSION,
+          DEV_REPLAY_SESSION,
+        ),
+  };
+}
+
+export function shouldSendConsoleToSentryInProduction(): boolean {
```
Kayanski, line 177: I'm not sure about this, won't it just flood sentry and burn credits REAL fast if we do that ?
In the backend, we make sure to send only errors that we don't understand, so sending all the console logs will be a lot ! 
We could add the console logs in the breadcrumbs though

Kayanski (approved):
Some comments, otherwise good

src/sentry/sharedOptions.ts
mmioana: Only if we ever set `NEXT_PUBLIC_SENTRY_CONSOLE` to true will we use the console logs to Sentry. The idea was to be able to customise things in case we really need it

src/sentry/sharedOptions.ts
mmioana: So this should definitely be shared for the nextjs environments. I think between the FE and BE init only the integrations should be different

PR #2856, opened 2026-05-12
fix: try to reduce cold-load latency for missions/earn

mmioana:
Closes JUM-891

## Summary by CodeRabbit

* **Bug Fixes**
  * More resilient data loading with safe fallbacks
  * Removed several route-level loading placeholders

* **Performance**
  * Increased mission page revalidation interval
  * Removed an experimental stale-time caching setting

* **New Features**
  * Async missions banner that loads campaign banners

* **Behavior**
  * Legacy quest routes now redirect to missions and are set to avoid indexing

* **Refactor**
  * Removed debug logging and tightened hydration gating for wallet providers

* **UX**
  * Reduced route prefetch targets for faster startup

[Review Change Stack](https://app.coderabbit.ai/change-stack/jumperexchange/jumper-exchange/pull/2856)


yordanove:
Spent some time on the 27 CI failures — nearly all hit the same `.welcome-screen-container.not.toBeVisible()` timeout after clicking `get-started-button`.

Pulled traces from this PR's failing run vs our last passing CI on develop:

- In the failing trace, console has: `Cannot update a component (SolanaProviderValues) while rendering a different component (SuiProvider)`. Stack points at `SuiProvider.tsx:21` — the `createDAppKit(...)` call inside render.
- Same error is absent from develop's passing CI traces.

The hydration guard added to `SuiProvider` (return children early if `!isHydrated`) plus the matching one on `SVMProvider` looks like the trigger — both providers re-mount in the same render cycle once `useHydrated` flips, and `createDAppKit` doing work inside render is the known [setState-during-render anti-pattern](https://react.dev/link/setstate-in-render) React's catching here.

https://react.dev/reference/eslint-plugin-react-hooks/lints/set-state-in-render

https://react.dev/reference/rules/components-and-hooks-must-be-pure

Can't prove from the artifacts alone whether the React warning directly blocks the welcome state update, or whether it's a pre-hydration click race that's a parallel symptom of the same timing change. Both fit the evidence. Either way the fix direction is the same: pull `createDAppKit` out of render (useEffect or memoize outside) — should keep the perf win without disrupting hydration.

Lmk when there's a build to retest.

yordanove (approved):

PR #2863, opened 2026-05-18
feat: collapse dust token & USD amounts to <0.0001 / <$0.01 (JUM-840)

laurentsenta:
## Summary

Dust positions (non-zero token amounts with many leading decimal zeros) previously rendered as long strings like `0.000000000000000001 eETH`, which is noisy and unreadable. This PR collapses them consistently across both axes:

- **Token amounts**: any non-zero value `< 0.0001` → `<0.0001 SYMBOL`
- **USD amounts** (Earn + Portfolio DeFi positions only): any non-zero value `< $0.01` → `<$0.01`

The subscript approach (`compactNumberFormatter`) was evaluated and rejected — the Unicode subscript notation (e.g. `0.0₁₇1 eETH`) reads as a rendering bug rather than intentional formatting.

## Changes

**Commit 1 — token amount collapse**
- **`src/utils/formatNumbers.ts`** — adds `DUST_AMOUNT_THRESHOLD`, `DUST_AMOUNT_LABEL`, `formatTokenAmountWithDust` as single source of truth.
- **`src/utils/Token.ts`** — `SimpleToken.formatTokenWithSymbol` delegates to the helper.
- **`src/components/composite/DeFiPositionCard/utils.tsx`** — `renderValueCell` hint uses the helper.
- **`src/utils/Token.spec.ts`** / **`src/utils/formatNumbers.spec.ts`** — unit tests.
- **`docs/decisions/dust-rendering.md`** — decision note.
- **`ARCHITECTURE.md`** — links to `docs/decisions.md`.

**Commit 2 — USD amount collapse**
- **`src/utils/formatNumbers.ts`** — adds `DUST_USD_THRESHOLD`, `DUST_USD_LABEL`, `formatUSDWithDust` mirroring the token helper.
- **`src/utils/Token.ts`** — `ExtendedToken.formatAmountUSD` routes through `formatUSDWithDust`.
- **`src/components/EarnDetails/EarnDetailsActionsPosition.tsx`** — `formatUSD(amountUSD)` fallback swapped to `formatUSDWithDust`.
- **`src/components/composite/DeFiPositionCard/utils.tsx`** — `renderValueCell` title uses `formatUSDWithDust`.
- **`src/utils/Token.spec.ts`** / **`src/utils/formatNumbers.spec.ts`** — unit tests for USD path.
- **`docs/decisions/dust-rendering.md`** — updated to cover both axes.

## Linked Linear ticket

[JUM-840 — Display dust position values as '<0.0001' instead of long decimals](https://linear.app/lifi-linear/issue/JUM-840/display-dust-position-values-as-00001-instead-of-long-decimals)

## Sister PRs

N/A — purely presentational change in `jumper-exchange`.

## Summary by CodeRabbit

* **New Features**
  * Dust-aware formatting: token amounts below 0.0001 render as "<0.0001 [symbol]" and USD below $0.01 render as "<$0.01"; applied across position cards and earnings/price displays with localized labels

* **Tests**
  * Added tests covering dust-formatting behavior and boundary cases for token and USD values

* **Documentation**
  * Added decision and docs describing dust-rendering rules, thresholds, examples, and where it's applied; architecture doc updated to point to decisions file

[Review Change Stack](https://app.coderabbit.ai/change-stack/jumperexchange/jumper-exchange/pull/2863?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


laurentsenta:
tested 
CleanShot 2026-05-18 at 13 09 19@2x

with subscript (see @Kayanski comment)
CleanShot 2026-05-18 at 14 32 40@2x

with Octave's feedback (back to <xxxx) and adding to $$$
CleanShot 2026-05-18 at 15 05 02@2x

docs/decisions/dust-rendering.md
```diff
@@ -0,0 +1,40 @@
+# Dust rendering
```
Kayanski: Could you link to this file inside the code ? 
Or maybe we run the risk of the links becoming stale if we don't enforce them in CI ?

Kayanski:
0.0001 Bitcoin is 6 EUR currently, so that wouldn't qualify as dust, yet it would show up as <0.0001, which is not the best behavior
Here are some strategies that could work : 
- Display like the widget : Capture d’écran 2026-05-18 à 14 01 56

Another strategy, if you prefer the <0.x1 strategy: 
- Compute the price of the asset. If it's <0.01$ (1cent) AND <0.0001, show <0.0001μ
- If it's greater than 1 cent, still show the normal amount (or use widget formatting for instance)

Kayanski (changes requested):
We could use this, exported from Lifi widget : https://github.com/lifinance/widget/blob/91424eaa9a7ebe8e316591961d29711221bbc8b8/packages/widget/src/utils/compactNumberFormatter.ts#L26

laurentsenta:
> We could use this, exported from Lifi widget : https://github.com/lifinance/widget/blob/91424eaa9a7ebe8e316591961d29711221bbc8b8/packages/widget/src/utils/compactNumberFormatter.ts#L26

Thanks for the pointer. We tried it locally and the subscript notation (e.g. `0.0₁₇1 eETH`) reads as a rendering bug rather than intentional formatting — same feedback came back from Octave. We're keeping the explicit `<0.0001 SYMBOL` collapse and extending the same pattern to USD (`<$0.01`) so dust positions read consistently across both axes.

On the `0.0001 BTC ≈ €6` misclassification: agreed the threshold is denomination-blind on the token side, but the USD column always sits next to it and now collapses below `$0.01`, so the user still has a non-misleading anchor to interpret the row. We'd rather revisit a price-aware threshold in a follow-up than ship subscript notation here.

docs/decisions/dust-rendering.md
laurentsenta: Agreed that links inside a long-lived decision doc can rot. We're keeping the decision file as a standalone, high-level note rather than a code map — to flag freshness, we'll prefix the filename and title with a `YYYYMMDD` date stamp (e.g. `20260518-dust-rendering.md`) as a visual staleness signal. CI enforcement felt heavy for a half-page note; we'll revisit if these docs grow.

Kayanski (approved):

docs/decisions/dust-rendering.md
laurentsenta: (experimented with agent taking over the whole task)

the end game is to have agents keep architectures files up to date (wouldn't be surprising if the whole industry moved away from md files tbh because the whole thing gets stale if not tended to on every PR), so please treat this as just a way to track recent decisions that an agent can read if it matches what they are looking for.

src/components/composite/DeFiPositionCard/utils.tsx
mmioana, line 1: Mmmm the `DeFiPositionCard` is no longer used - was missed to be cleaned up after the new Portfolio Provider was tested and approved. We should apply changes only to the `PositionCard` and sub-components/hooks.

Given we want to display this dust amount, we should probably enhance the `useTokenFormatters` and `usePortfolioFormatters` - which are used by the TokenAmount's components

laurentsenta:
Follow-up on the open review threads (latest commits `a26915fe`, `1bf88631`):

**Targeting the live code (@mmioana):** dropped the dead `DeFiPositionCard` and moved dust formatting into `useTokenFormatters` / `usePortfolioFormatters`, which the live `TokenAmount` components consume. Earn "Your position" stays on the existing `Token` formatters.

**`Token.ts` "applied too broadly" (CodeRabbit):** keeping the dust collapse in `Token.formatAmount`/`formatAmountUSD`. In practice the only consumer of these methods is the in-scope Earn position card (`EarnDetailsActionsPosition.tsx`), so the scope is effectively contained — no unrelated surface changes formatting. The Portfolio surface goes through the hooks instead.

**Import alias (CodeRabbit):** `EarnDetailsActionsPosition.tsx` now imports `@/utils/formatNumbers` and `@/utils/Token`.

**Decision doc (@Kayanski):** removed `docs/decisions/*` from this PR; the docs/staleness convention will be handled separately so it doesn't block this change.

**Denomination-blind threshold / `0.0001 BTC ≈ €6` (@Kayanski, Octave):** keeping the explicit `<0.0001 SYMBOL` collapse (subscript notation read as a rendering bug). The adjacent USD column now collapses `<$0.01`, giving a non-misleading anchor; a price-aware token threshold is deferred to a follow-up.

mmioana (approved):
Really nice work! Thank you for cleaning up the unused component and adding the tests ❤️

PR #2867, opened 2026-05-19
fix(UnicornScene): stop rendering script-load error inside Golden Rou…

Octave-byte:
…te modal

When the UnicornStudio CDN script fails to load (ad-blocker, CSP, flaky network), the developer error string was being rendered inside the Golden Route modal in place of the animation. The animation is decorative — log the failure to console instead so the modal copy and Collect CTA remain unobstructed.

- [x] I have performed a self-review of my code
- [x] This pull request is as small as possible and only tackles one problem
- [x] If this changed the API, I have updated the documentation


Octave-byte:
--> Not intended fix

PR #2890, opened 2026-05-28
fix(tests): walletSwapExecute handle pre-approved-allowance - JUM-1024

yordanove:
## Which Linear ticket belongs to this PR?

Closes [JUM-1024](https://linear.app/lifi-linear/issue/JUM-1024)
Related: [JUM-952](https://linear.app/lifi-linear/issue/JUM-952)

## Summary

`walletSwapExecute.spec.ts` (qase 205) called `signPopup` twice — once for USDC approval, once for the swap signature. On wallets with sufficient existing USDC allowance for the LiFi diamond contract on Arbitrum (every run after the first), the widget skips the approval transaction; only one popup appears. Second `signPopup` timed out → false failure even when the on-chain swap succeeded.

Adds `MetaMaskPage.signPopupIfPresent(context, timeoutMs = 15_000): Promise<boolean>` — returns `true` if a popup appeared and was signed, `false` if none appeared within the timeout. No exception-for-control-flow. Spec updated to call `signPopup` (always-present first popup) then `signPopupIfPresent` (conditional second).

Spec stays `test.fixme()`'d on develop (JUM-952 still gates removal); this PR corrects the body so it works correctly when JUM-952 lifts the fixme. Without it, every cron run after the first would falsely fail.

## Files

- `tests/e2e/wallet/pages/MetaMaskPage.ts` — adds `signPopupIfPresent` helper
- `tests/e2e/walletSwapExecute.spec.ts` — uses the new helper for the conditional second popup

## Summary by CodeRabbit

* **Tests**
  * Improved handling of wallet signing popups to reduce flaky timing and support optional/conditional confirmations.
  * Enhanced popup detection with a fast-path for already-open dialogs and a retry fallback, speeding and stabilizing flows.
  * Updated wallet-swap test to a two-phase signing sequence: always handle initial approval, then conditionally confirm a follow-up signing popup.
  * Added a helper to attempt an optional sign confirmation when a popup may or may not appear.


tests/e2e/wallet/pages/MetaMaskPage.ts
```diff
@@ -305,6 +305,44 @@ export default class MetaMaskPage extends WalletPage {
...
+        throw error;
+      }
+    }
+
+    await popupPage.waitForLoadState('domcontentloaded');
+    await popupPage.bringToFront().catch(() => {});
+    const popupWallet = this.createExtensionPageInstance(popupPage);
+    await popupWallet.confirmTransaction();
+    return true;
+  }
```
mmioana, line 320: I feel this code is more or less similar to what the code in `WalletPage`'s `confirmInPopup` and `getPopupPage`. Wondering if we could add a flag there to perform this conditional check?

yordanove: Good call, there was real overlap there. Consolidated it: `confirmInPopup` now takes an options bag (`optional` / `timeoutMs` / `urlMatchers`) and returns a boolean, and `signPopupIfPresent` is just a thin delegate (`confirmInPopup(ctx, { optional: true, timeoutMs })`). `getPopupPage` gained the `timeoutMs` + an already-open scan so the optional path stays snappy.

Going through `confirmInPopup` also surfaced that the old hand-rolled `waitForLoadState`/`bringToFront` lines were redundant — `getPopupPage → preparePopupPage` already does both (plus the CI viewport sizing the old optional path was skipping), so I dropped them.

I left `connectInPopup`/`rejectPopup` alone for now — they share the same tail, but I didn't want to widen the diff into the connect path (runs before every connected test) without a real-wallet run behind it. Happy to do that as a small follow-up if you'd like.

Smoke-checked `connect` locally (green); the sign/swap specs stay on the real-wallet workflow as before.

yordanove:
@coderabbitai full review

tests/e2e/walletSwapExecute.spec.ts
```diff
@@ -33,12 +33,12 @@ test.describe('Wallet swap — execute on Arbitrum', () => {
...
-      await test.step('Sign USDC approval in MetaMask', async () => {
+      await test.step('Sign first wallet popup (approval or swap)', async () => {
         await wallet.signPopup(walletContext);
       });
 
-      await test.step('Sign swap transaction in MetaMask', async () => {
-        await wallet.signPopup(walletContext);
+      await test.step('Sign swap signature if approval preceded it', async () => {
+        await wallet.signPopupIfPresent(walletContext);
       });
```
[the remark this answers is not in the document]
yordanove: @coderabbitai good point on the approval-mining latency, but there's a tradeoff in the other direction I'd like your take on.

In the pre-approved case — which is the common one here, since the QA wallet already has allowance for this pair — no second popup ever appears, so `signPopupIfPresent` dead-waits the *entire* `timeoutMs` before returning false. So bumping the default 15s → 60s adds ~45s of dead wait to every pre-approved run, for a benefit that only applies to the not-pre-approved branch.

And that not-pre-approved branch is currently `test.fixme`'d (qase 205) — the wallet is pre-approved — so the >15s tail-latency case can't actually manifest until we un-fixme it and run on the real-wallet harness.

Given that, would you still bump the default now, or hold at 15s and tune from observed approval-mining latency once the path is actually exercised? A middle option would be ~30s as a hedge. Which would you weigh higher — the dead-wait cost on the verified path, or margin for the unverified one?

mmioana (approved):

PR #2898, opened 2026-06-01
feat: filter expired notifications client-side using expiresAt (JUM-1034)

chmanie:
## What

Frontend half of [JUM-1034](https://linear.app/lifi-linear/issue/JUM-1034/filter-expired-notifications-client-side-using-expiresat). The notifications backend now stops filtering expired rows server-side and returns `expiresAt` on every notification, leaving presentation to the client.

This moves the expiry decision to the frontend, centralised in a single `isExpired(n)` helper so the policy is trivial to change later (e.g. show expired greyed-out or behind a toggle). **No visible behaviour change in this PR** — expired notifications stay hidden, exactly as today.

## ⚠️ Depends on notifications PR

This requires [jumperexchange/jumper-notifications#43](https://github.com/jumperexchange/jumper-notifications/pull/43) to be **merged and deployed first**. That PR:
- stops filtering expired rows from the list endpoint (returns them with `expiresAt`), and
- changes the summary response from `{ count, latestCreatedAt }` to `{ count, expiredCount }`.

Until it ships, the list won't contain expired rows (so the new filter is a no-op) and the summary type here won't match the live response.

## Summary by CodeRabbit

* **New Features**
  * Expired notifications are now filtered client-side; unread and badge counts exclude expired items.
  * Notification summary now includes a separate expired-count value.

* **Tests**
  * Added tests covering expiration detection and boundary behavior.
  * Updated notification tests to reflect the new summary shape and badge-visibility logic.

### DIFF
--- a/src/utils/notifications/isExpired.ts
+++ b/src/utils/notifications/isExpired.ts
@@ -1,0 +1,28 @@
+import type { Notification } from '@/types/notifications';
+
+/**
+ * Whether a notification has passed its expiry.
+ *
+ * The backend no longer filters expired rows server-side — it returns every
+ * notification with an `expiresAt` and lets the client decide how to present
+ * them (see JUM-1033/JUM-1034). This helper is the single place that encodes
+ * that policy: today we hide expired notifications to match prior behaviour.
+ * To show them later (greyed-out, behind a toggle, …) change this helper or
+ * its call sites — nothing else needs to move.
+ *
+ * Notifications without an `expiresAt` (or with an unparseable one) never
+ * expire and are always shown.
+ */
+export const isExpired = (
+  notification: Pick<Notification, 'expiresAt'>,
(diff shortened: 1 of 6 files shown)


mmioana (approved):

src/utils/notifications/isExpired.ts
```diff
@@ -0,0 +1,30 @@
...
+): boolean => {
+  if (!notification.expiresAt) {
+    return false;
+  }
+  const expiresAt = parseISO(notification.expiresAt);
+  if (!isValid(expiresAt)) {
+    return false;
+  }
+  // Expired once expiresAt <= now (the exact boundary counts as expired).
+  return !isAfter(expiresAt, now);
```
mmioana: Lovely 💯

mmioana (approved):

PR #2904, opened 2026-06-03
feat: translate notifications via i18next - JUM-1032

chmanie:
## Summary

Notifications arrive from the backend as English `title`/`body`/`ctaLabel` strings plus a `sourceRuleId` and a `metadata` bag. This renders them through an i18next catalogue keyed by `sourceRuleId`, interpolating directly from `metadata`, so notification copy is localizable. The stored English strings remain the fallback for any rule without a catalogue entry, so notifications are always displayable.

Frontend half of [JUM-835](https://linear.app/lifi-linear/issue/JUM-835); the backend per-rule metadata contract shipped in jumperexchange/jumper-notifications#47.

## Changes

- Add `notifications.<ruleId>.{title,body,cta}` catalogue entries for the six rules in the backend metadata contract (`idleAssets`, `bridgeToEarn`, `apyDrop`, `newOpportunity`, `newToolLaunch`, `userLevelUp`).
- Add `chainNameExt`/`chainNamesExt` i18next formatters (backed by `findChain`) so chain ids render as brand names; resolve the tool verb via `$t()` nesting. Interpolation is fully declarative — no per-rule JS.
- Reuse the existing `percentExt`/`currencyExt` formatters for raw-fractional APY and raw USD values.
- Resolve copy in `useNotificationContent`, render it in `NotificationItem`.
- Unit test covering formatting, chain/verb resolution, and the unknown-rule fallback.

## Linked Linear ticket

Fixes JUM-1032 (frontend half of JUM-835).

## Sister PRs

- jumperexchange/jumper-notifications#47 — backend per-rule metadata contract (JUM-835)

## Summary by CodeRabbit

* **New Features**
  * Notifications now resolve content via translations with localized CTA labels and safe fallbacks; CTA buttons show translated text.
  * Chain-name formatting added for human-friendly chain names in notification copy.
  * New notification templates: APY drops, new tool launches, user level progression, idle assets, and bridge opportunities.

* **Tests**
  * Added tests covering notification content resolution, translation formatting, and fallback scenarios.

PR #2906, opened 2026-06-04
feat: translate perkLevelUp notification via i18next (JUM-1058)

chmanie:
Adds the frontend i18next translation for the new `perkLevelUp` notification rule (jumperexchange/jumper-notifications#49, JUM-998), following the pattern from #2904. Stacked on top of #2904 (JUM-1032).

## Changes

- Add `notifications.perkLevelUp` catalogue entry with `_one`/`_other` plural variants driven by `count`, interpolating `newLevel` and the perk names.
- Add a generic `listExt` i18next formatter (`{{items, listExt(prop: name)}}`) that joins an array, optionally plucking a property; `chainNamesFormatter` now delegates its join to it.
- Make `resolveNotificationContent` pass `metadata` into `i18n.exists()` so plural-only entries resolve, falling back to the stored English copy when `count` is absent.
- Unit tests for the single-perk, multi-perk, and missing-`count` fallback cases.

## Backend pairing

Translation activates once the `perkLevelUp` metadata carries `count` (alongside the existing `perks`). Until then it safely falls back to the backend-rendered English string.

## Summary by CodeRabbit

* **New Features**
  * Perk level-up notifications with pluralized title/body and a localized CTA.
  * Consistent, locale-aware list formatting for multi-item displays; new formatter exposed to translations.

* **Tests**
  * Added tests covering single/multiple perk unlocks and fallback behavior.

* **Bug Fixes**
  * Improved translation existence checks and CTA fallbacks to ensure correct notification content.

* **Docs/Types**
  * Typing updates to support the new list-format interpolation keys.

### DIFF
--- a/src/hooks/notifications/useNotificationContent.ts
+++ b/src/hooks/notifications/useNotificationContent.ts
@@ -23,30 +23,41 @@ export function resolveNotificationContent(
   notification: Notification,
   ctx: { t: TFunction; i18n: I18nInstance },
 ): NotificationContent {
+  const fallback: NotificationContent = {
+    title: notification.title,
+    body: notification.body,
+    ctaLabel: notification.ctaLabel,
+  };
+
   const ruleId = notification.sourceRuleId;
-  if (!ruleId) {
-    return {
-      title: notification.title,
-      body: notification.body,
-      ctaLabel: notification.ctaLabel,
-    };
+  const params = notification.metadata ?? {};
+  const titleKey = `notifications.${ruleId}.title`;
+  const bodyKey = `notifications.${ruleId}.body`;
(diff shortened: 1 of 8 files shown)


src/hooks/notifications/useNotificationContent.ts
mmioana, line 1: So we don't want to translate whatever possible and use fallbacks only where there is no existing key?

mmioana (approved):

src/hooks/notifications/useNotificationContent.ts
chmanie: I'm not sure how we are handling that in other parts of the app, but I feel like it's better if everything is in english rather than mixed languages in the notification. WDYT?

PR #2910, opened 2026-06-05
feat: track notification engagement events in posthog - JUM-238

chmanie:
Wires the notification engagement funnel (received → seen → clicked → dismissed) into PostHog via the existing Jumper tracking pipeline. PostHog is analytics-only and is not the read/unread source of truth.

- New `cat_notifications` events and `param_notification_*` properties in `trackingKeys.ts`
- `useNotificationTracking` hook centralises the funnel with per-session dedup for impression events
- `received` fired on fetch, `seen` via IntersectionObserver, `clicked`/`dismissed` from `NotificationItem`

JUM-238

## Summary by CodeRabbit

* **New Features**
  * Added notification engagement tracking (received, seen, clicked, dismissed). Events fire when notifications are fetched, when an item scrolls into view, on CTA click, and on dismiss. Events are deduplicated per user session; CTA clicks mark unread notifications as read.

* **Tests**
  * Added tests validating event emission, deduplication behavior, and per-user scoping for impressions.


src/hooks/userTracking/useNotificationTracking.ts
```diff
@@ -0,0 +1,108 @@
...
+    [emit],
+  );
+
+  const trackDismissed = useCallback(
+    (notification: Notification) =>
+      emit(
+        TrackingAction.NotificationDismissed,
+        'notification-dismissed',
+        notification,
+      ),
```
mmioana: Can we dismiss more than once?

src/hooks/userTracking/useNotificationTracking.ts
mmioana, line 1: Maybe we could move this hook and its spec under `src/hooks/userTracking` given we're keeping all tracking hooks there

src/components/Notifications/NotificationItem.tsx
```diff
@@ -51,12 +53,33 @@ export const NotificationItem: FC<NotificationItemProps> = ({
...
+          trackSeen(notification);
+          observer.disconnect();
+        }
+      },
+      { threshold: 0.5 },
+    );
+    observer.observe(element);
+    return () => observer.disconnect();
+  }, [notification, trackSeen]);
+
```
mmioana, line 74: Maybe we could use `useInView` from `motion/react`

src/hooks/userTracking/useNotificationTracking.ts
chmanie: Nope, you're right, we shouldn't be able to. I'll adjust

src/hooks/userTracking/useNotificationTracking.ts
chmanie: Ah yes, thank you!

mmioana (approved):
This is some really nice work! 🥇

PR #2920, opened 2026-06-09
fix: performance, latency and stale data for missions and earn

mmioana:

Contributes to https://linear.app/lifi-linear/issue/JUM-884/perf-tab-navigation-latency-portfolio-earn-and-missions-detail-pages

## Summary
Earn and mission pages showed stale “updated X ago” data and slow cold loads (p95 often 5–12s). Root cause was competing cache layers: unstable_cache + ISR baked in build-time data, while the client refetched fresh API data on mount — hydration then restored the older dehydrated React Query state.

## Fix

- Single SSR hydration path — shared fetch helpers on server and client; refetchOnMount: false on prefetched queries to avoid overwrite/flicker
- Server-only page cache — cachedEarnFetch / cachedMissionsFetch with per-key unstable_cache (300s), aligned with page revalidate
- Faster SSR — Suspense streaming, parallel prefetches (earn list + tops; missions list + banner), createServerQueryClient with matching staleTime
- Perf gates — earn/mission routes enforce p95 LCP ≤ 3s (PERF_EARN_P95_BUDGET_MS, PERF_MISSIONS_P95_BUDGET_MS)

## Test plan

- [ ]  Open earn detail — “updated” badge stays consistent after load (no revert to older timestamp)
- [ ]  Cold-load /en/earn, /en/missions, and a detail slug — content ready within ~2–3s
- [ ]  PERF_SAMPLES=20 pnpm test:perf:cold-lcp passes earn + mission p95 budgets after deploy

## Summary by CodeRabbit

* **Performance Improvements**
  * Pages now use Suspense-based loading with lightweight skeletons for snappier perceived load.

* **Faster Updates**
  * Mission content refresh interval shortened to ~5 minutes for fresher content.

* **Stability & UX**
  * Centralized and improved data fetching/caching reduces redundant loading and improves consistency across pages.


src/app/lib/makeQueryClient.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export const queryClientDefaultOptions: DefaultOptions = {
+  queries: {
+    enabled: true,
+    staleTime: FIVE_MINUTES_MS,
+    refetchInterval: false,
+    refetchIntervalInBackground: false,
+    refetchOnWindowFocus: true,
+    refetchOnReconnect: true,
+    refetchOnMount: false,
```
Kayanski, line 20: so we never refetch refetchOnMount, do we have some queries that need that ?

src/app/[lng]/earn/[slug]/page.tsx
```diff
@@ -3,45 +3,17 @@ import {
...
-import { getOpportunityBySlugCached } from 'src/app/lib/getOpportunityBySlug';
-import { getOpportunityRelatedMarketCached } from 'src/app/lib/getOpportunityRelatedMarket';
+import { Suspense } from 'react';
 
 type Params = Promise<{ slug: string }>;
 
 export const dynamicParams = true;
 export const revalidate = 300;
 
-export async function generateStaticParams(): Promise<{ slug: string }[]> {
```
Comment (Kayanski) [inline src/app/[lng]/earn/[slug]/page.tsx:27]: No more building earn pages on compile ?

Kayanski (approved):

src/app/lib/makeQueryClient.ts
mmioana: This is the default, so in case we need to refetchOnMount we'll need to set that flag

src/app/[lng]/earn/[slug]/page.tsx
Comment (mmioana) [reply src/app/[lng]/earn/[slug]/page.tsx:27]: Nope, cause it slows down the build and there is no real benefit

PR #2921, opened 2026-06-10
feat: add portfolio pnl values and chart to header section

mmioana:

Contributes to https://linear.app/lifi-linear/issue/JUM-966/portfolio-wallet-balance-chart-via-zerion

## Summary by CodeRabbit

* **New Features**
  * Added portfolio balance history and PnL tracking with selectable periods.
  * Introduced a new portfolio header PnL overview with loading/empty handling and an optional line chart.
  * Added a PnL chart disclaimer tooltip when the chart is unavailable, plus updated skeleton visuals.

* **Bug Fixes**
  * Improved consistency of chart/analytics data formats across portfolio and earnings.
  * Refined loading, refreshing, and display logic for PnL and chart states.

* **Documentation**
  * Added localized text for the PnL chart disclaimer.


Kayanski (approved):

Kayanski:
Some comments though : 
1. Important limitation: Zerion does not return historical protocol-position data. The chart reflects wallet-token balances over time only — DeFi positions are included at current value but not retro-priced. Call it out in the UI tooltip / disclaimer.
--> You could have a small (!) icon to call that out
2. Decouple graph and PNL. You can display one without the other. There are cases where pnl data is not available while chart is (and the other way around), can you decouple both when printing those on the screen
3. On the PNL data, can you show nothing as a default value when PNL is not available (query error state) ? From the task : hide the header delta, keep the chart visible, Wallets with >1M transactions (because of zerion error): chart still renders, PnL header gracefully hidden.
4. I'm not sure, but I would show a negative amount on the PNL absolute value with a minus sign and not only in RED
5. Color sync: Use the same delta >= 0 ? green : coral rule for the PnL header and the chart line. Both must agree on screen.

Kayanski (approved):
Very nice. Please take this into account (task change) https://github.com/jumperexchange/jumper-exchange/pull/2979

PR #2938, opened 2026-06-15
feat: add transaction view to portfolio page

mmioana:

Closes JUM-792

Sister PR: https://github.com/jumperexchange/jumper-backend/pull/949

## Summary by CodeRabbit

* **New Features**
  * Added a feature-flagged Transactions tab in portfolio with URL-driven tab selection.
  * Introduced transaction filtering, sorting, and cursor pagination, including refresh controls with rate-limit tooltips.
  * Added transactions-specific filter UI and an empty-page state for “no results on this page”.
  * Virtualized large multi-select option lists and improved filter modal navigation/back support.
* **Bug Fixes**
  * Updated transaction table summary layout and removed USD total display to match new per-side amount fields.
  * Improved date-range badge labeling and select/filter overflow behavior.
* **Chores**
  * Added virtualized list dependency.
* **Tests**
  * Added cursor pagination component interaction tests.


Kayanski:
We need to be able to force refresh the transactions, can you implement that, like a button that says refresh, that sends the forceRefresh field to the query and that updates the transactions (in case the user has a transaction they want to see)
However, it's limited to 5 requests every 15 minutes (this is not exposed to the frontend though)
In a similar rate limiting way, we limit to 15 actual zerion requests (uncached) per 15 minutes.
It might be a good idea to have that in the response headers and print that in the frontend , what do you think ?

Kayanski:
Hey @mmioana — on the asset filter UX: the current flat asset list is quite hard to navigate, especially with many tokens across chains. We discussed this briefly but wanted to capture it as a concrete suggestion: could we replace it with a **chain → token** two-step filter (pick a chain first, then see only the tokens that appeared in that chain's transactions), with the token list **prioritised by assets present in the connected wallet**? That would make the most common case ("show me my ETH transactions") a two-tap flow instead of scrolling through a flat list. Happy to spec it out more if useful.

Kayanski:
`disabled={isLoading}` is always `false` here — the `if (isLoading) return null` guard on line 15 means this component never renders while loading, so `disabled` is dead code.

https://github.com/jumperexchange/jumper-exchange/blob/c77af147f2c9146d2292da40a3527844bd78df6e/src/app/ui/portfolio/PortfolioTransactions/PortfolioTransactionPagination.tsx#L15-L29

**Proposed fix** — pick one depending on the desired UX:
- Option A (keep pagination visible but disabled during load): remove the `if (isLoading) return null` guard, let `disabled={isLoading}` do the work.
- Option B (keep the disappear behaviour): remove the `disabled` prop since it never fires.

Kayanski:
When `goToNextPage` fires before the cursor-storage effect has written the cursor into the map, `cursors.get(page) ?? null` evaluates to `null`. React Query treats `null` and `undefined` as distinct keys, so it fires a fetch with no cursor and returns page-0 data.

https://github.com/jumperexchange/jumper-exchange/blob/c77af147f2c9146d2292da40a3527844bd78df6e/src/providers/TransactionProvider/TransactionProvider.tsx#L42-L53

**Proposed fix:**
- Line 42: `cursors.get(page) ?? undefined` — drop the `null` fallback so RQ does not treat it as a distinct key.
- Pass `enabled: enabled && (page === 0 || cursors.has(page))` to `useTransactionsData` so the query waits for the cursor instead of fetching without one.

Kayanski:
`datesBadge` is defined in at least two places in this PR and a third exists in the codebase:

https://github.com/jumperexchange/jumper-exchange/blob/c77af147f2c9146d2292da40a3527844bd78df6e/src/components/PortfolioFilterBar/utils.ts#L8-L29

https://github.com/jumperexchange/jumper-exchange/blob/c77af147f2c9146d2292da40a3527844bd78df6e/src/app/ui/learn/LearnPageArticlesSection/utils.ts#L17-L37

`getDatesBadge` in `src/components/composite/FilterSortModal/FilterSortModal.stories.tsx:231` does the same thing.

Worth extracting to a shared util (e.g. `src/utils/filters/datesBadge.ts`) to avoid silent divergence when the logic changes.

Kayanski (approved):

yordanove:
Tested this on the preview (EVM + Solana + a Sui wallet, backend on develop) — overall it's in really good shape, happy for it to go in behind the flag 

Ran through:
- flag gating both ways (on → tab live; off → "soon" badge + `?tab=transactions` bounces back to Holdings)
- pagination (cursor round-trips, no dupes), filters + sort, and they **survive a hard reload** (nice URL-param work)
- refresh button (fires `forceRefresh=true`, quota tooltip shows the right remaining count)
- multi-wallet selector, Solana feed renders, mobile reflow
- spot-checked amounts/fees/timestamps vs basescan/etherscan — all match, and the spam filtering is spot on (0 genuine txs dropped over an 18-day window)

Two things I'd flag before we flip the flag on — not merge blockers since it ships dark:

**1. No error state.** If the endpoint 429s or errors, the tab shows the "No results — try clearing your filters" empty state (screenshot below). A rate-limited user can't tell they're rate-limited, and clearing filters won't help them. The provider already exposes `error`, it's just not rendered in `PortfolioTransactions`. Same thing happens when a Sui wallet is selected (backend 400s it — Sui isn't supported by the endpoint). FYI there's no error/empty state in the Figma either (checked) — might be worth a quick ping to design for what rate-limited/failed should look like, so it doesn't block you.

jum968-429-state


**2. Solana rows have no explorer link.** The design note says transactions should link to their native explorer, and EVM rows do (basescan/etherscan/arbiscan all correct) — but Solana rows come out with an empty Tx hash column (screenshot below). The backend returns a valid signature, so it's FE-side: `EntityExplorerLink` resolves the URL from `chain.metamask.blockExplorerUrls`, which Solana (tx `chainId` comes back as `101`) doesn't populate → returns null. Needs a solscan path.

jum968-solana


Everything else I noticed (column order, protocol column, in/out amounts, numbered pagination, `<0.0001` threshold) matches Octave's notes on JUM-792, so I left those alone. One scope check: the JUM-968 AC mentions Jumper-originated txs being visually distinguished + linking into the Jumper detail flow — didn't see that here, is that intentionally a later iteration?

Backend bits (429 Retry-After etc.) I'll raise with Nico separately. Nice work, this is a big one!

mmioana:
@yordanove I'll treat the first point, but for the 2nd we'll need to do changes on the BE side because that is not a valid LiFi chain id for Solana

yordanove (approved):

PR #2947, opened 2026-06-16
feat: [Jumper Pass] add running-month activity progress to Earn XP

chmanie:
Adds running (current-month) activity progress to the Earn XP section: ongoing per-credential progress with next-tier targets, a tier progress ring, and a shared achievement-tile shell used by the mission/activity cards.

- OngoingActivityCard + TierProgressRing
- useOngoingActivity hook (+ spec); removes useOngoingNumericQuests
- Shared AchievementTile* shell in Section.style (used by Earn XP + Your Achievements)
- SectionCarousel gains a `desktopSlidesPerView` prop for the activity layout
- Earn XP / ongoing-activity i18n + ongoing-credential tooltip

Note: the last-month "pass XP stat" referenced in the commit message landed in PR1 (#2935); this PR is just the Earn XP activity work.

Part of the Jumper Pass revamp stack (JUM-1038). This is PR 5/6, stacked on #2946 — review/merge after PR4.

Linear: https://linear.app/lifi-linear/issue/JUM-1131

PR #2948, opened 2026-06-16
feat: [Jumper Pass] revamp pass nav button

chmanie:
Revamps the Jumper Pass button in the navbar — renames LevelButton → PassButton, shows level progress via a shared ProgressRing, and surfaces the unlocked-perks count.

- Rename LevelButton → PassButton; update Navbar buttons + hooks
- Extract a reusable core ProgressRing (also adopted by TierProgressRing) + getLevelProgress util
- usePerks hook; perks-unlocked navbar label + i18n

Part of the Jumper Pass revamp stack (JUM-1038). This is PR 6/6, stacked on #2947 — review/merge after PR5.

Linear: https://linear.app/lifi-linear/issue/JUM-1132


src/components/core/ProgressRing/ProgressRing.tsx
mmioana, line 1: Could we somehow merge this and `src/components/ProfilePage/sections/EarnXpSection/TierProgressRing.tsx` together?

There is also the `src/components/composite/AssetProgress` - not sure it's helpful in this case?

chmanie: Good call on the redundancy. Resolved it by inlining `TierProgressRing` into its single call site (`OngoingActivityCard`) and deleting the file (92e3f01a) — it was just a tooltip + success-color wrapper used once.

I kept it out of the core `ProgressRing` deliberately: `ProgressRing` is the generic primitive (bare ring on a faint track, themed via `currentColor`, sizable) and is used directly by `PassButton`. Folding tooltip/tier concerns into it would couple the core to a feature, so the wrapper's logic now lives at the one place that needs it.

On `AssetProgress` — I don't think it fits here. It's a filled progress widget with centered content (percent text / entity icons), dollar formatting and thresholds (built on the `Percent` core); reusing it for a 20px navbar ring would drag in all that portfolio/value plumbing. There *is* a deeper overlap worth noting: `Percent` also draws a ring-on-track, but via the MUI track slot with a hardcoded radius tied to one size — which is exactly why the new `ProgressRing` uses two overlapping `CircularProgress` so it scales to any size. Converging those two is a bigger refactor I'd rather not fold into this PR.

mmioana (approved):

PR #2950, opened 2026-06-17
ci(e2e): serve a prod build in CI to fix dev-server boot timeouts

yordanove:
> Branch name is historical — it began as the settings-qase-8 investigation. The actual change here is the CI infra win below; it does **not** fix qase 8.

## What
Under CI (and local `E2E_PROD_BUILD=1`), serve a **production build** (`pnpm build && pnpm start`) instead of the Turbopack **dev** server, and bump shard `timeout-minutes` 25 → 40 for the added cold-build time. Also gates `output: 'standalone'` off for the e2e build so `next start` serves cleanly. Mirrors `playwright.perf.config.ts`.

## Why
The Turbopack dev server intermittently **boot-hangs on CI runners** → `config.webServer` 300s timeout → an entire shard fails with **0 tests run** (seen on multiple recent runs, 1–2 shards each). A pre-built prod server avoids the on-demand-compile boot hang. A dispatched run on this branch had **0 boot-timeouts** (vs dev-server runs with 1–2), and shard 1 — which gets timeout-cancelled at 25 min on the heavy wallet/earn specs — passed with the 40-min bump.

## Trade-off
Adds a per-shard cold `next build` (~5–8 min), offset by faster/more-stable test runs (no compile stalls mid-test). Worth your call: per-shard build vs build-once-and-share-artifact.

## Note
Does **not** fix settings qase 8 Desktop (a separate CI-environment issue under JUM-1116). The value here is CI **stability** — removing the boot-timeout + shard-cancel class of red.

Relates to JUM-1116.

## Summary by CodeRabbit

* **Tests**
  * Increased end-to-end test job timeouts to accommodate production build testing.
  * Updated test environment to validate against production builds instead of development server.

* **Chores**
  * Adjusted build configuration for test environments.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -108,6 +108,7 @@ jobs:
           QASE_TESTOPS_RUN_ID: ${{ needs.create-qase-run.outputs.run_id || '' }}
           QASE_TESTOPS_API_TOKEN: ${{ secrets.QASE_TOKEN }}
           SHARD: ${{ matrix.shard }}/${{ strategy.job-total }}
+          E2E_PROD_BUILD: "1" # serve a plain prod build (not the dev server) — JUM-1116
           TEST_WALLET_SEED_PHRASE: ${{ secrets.TEST_WALLET_SEED_PHRASE }}
           TEST_WALLET_PASSWORD: ${{ secrets.TEST_WALLET_PASSWORD }}
         if: ${{ !cancelled()}} # Always attempt to run the tests even if cache setup failed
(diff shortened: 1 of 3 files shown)


.github/workflows/playwright.yml
```diff
@@ -78,19 +78,18 @@ jobs:
     needs: [create-qase-run]
     name: Test (${{ matrix.shard }} / ${{ strategy.job-total}})
     continue-on-error: false # Instead we use fail-fast to forward the failure state to subsequent jobs
-    # 25 min headroom — `swapActions.spec.ts` and `mainMenu.spec.ts` use
-    # `test.slow()` (3× per-test timeout) and one Hyperliquid case has
-    # `timeoutMs: 90_000`. 20 min was tight against worst-case shard timing.
-    timeout-minutes: 25
+    # 40 min: per-shard cold prod build (~5-8 min) on top of the existing headroom
```
Kayanski, line 81: That's Huge now !

Kayanski (approved):

.github/workflows/playwright.yml
yordanove: Yeah, hopefully it never hits that - I will monitor if we hit these timings and enter loops that would have timed out earlier if not for this huge increase.

PR #2952, opened 2026-06-17
fix(perf): guard pathname cookie and cache token list fetches server-side

mmioana:
- Guard pathname cookie in middleware — set only on first page load instead of every request, removing the unconditional Set-Cookie header that forced cf-cache-status: BYPASS on all HTML responses
- Extract getTokensQuery / getChainTokensQuery from useTokens into src/app/lib/tokens/tokenQueries.ts
- Add cachedTokensFetch.ts (unstable_cache, revalidate 300s) to bypass the 2 MB Next.js fetch-cache cap on the 7 MB token list; sitemap bridge generator and widget-amounts image route now use cached helpers

Contributes to JUM-884

## Problem

Two root causes of elevated LCP on earn/missions pages in production:

1. **Cloudflare cache bypass** — `src/proxy.ts` unconditionally called
   `response.cookies.set('pathname', …)` on every request. Any response
   with a `Set-Cookie` header gets `cf-cache-status: BYPASS`, meaning
   Cloudflare never served HTML from edge cache.

2. **7 MB token fetch exceeding Next.js's 2 MB data-cache cap** — the
   LI.FI `/pipeline/v1/tokens?chainTypes=EVM` response is ~7 MB, which
   silently fails Next.js's `fetch()` cache (`items over 2MB can not be
   cached`). Every server-side invocation (sitemaps, widget-amounts OG
   image route) re-fetched and re-parsed the full list.

### `src/proxy.ts`
Guard the `pathname` cookie so it is only written on the **first** page
load (when the cookie is absent). Subsequent requests no longer emit
`Set-Cookie`, allowing Cloudflare to cache HTML normally.

### `src/app/lib/tokens/tokenQueries.ts` (new)
Raw fetch functions (`getTokensQuery`, `getChainTokensQuery`) extracted
from `useTokens.ts` into a dedicated lib file, following the existing
`src/app/lib/` pattern. Decouples server-side data fetching from the
React hook.

### `src/app/lib/tokens/cachedTokensFetch.ts` (new)
Wraps the token queries in `unstable_cache` (5-minute TTL), matching the
`cachedEarnFetch` / `cachedMissionsFetch` pattern. `unstable_cache`
stores results at the data-cache level with no size limit, bypassing the
2 MB `fetch()` cap.

### `src/hooks/useTokens.ts`
Trimmed to only the `useTokens` React hook + re-exports of the query
functions from `tokenQueries.ts`. No behaviour change for client-side
callers.

### `src/app/api/widget-amounts/route.tsx`
Switched from raw `getChainTokensQuery` to `fetchChainTokensForPage`
(cached) since this is a server route.

### `src/utils/sitemaps/bridge.ts` + `bridge.spec.ts`
Switched from raw `getTokensQuery` to `fetchTokensForPage` (cached).
Test mock updated to target the new import path.

## Summary by CodeRabbit

* **Performance Improvements**
  * Token fetching is now cached with a 5-minute revalidation interval, reducing repeated API calls across token-dependent pages and sitemap generation.
  * Earn and missions page data caching was migrated to Next.js cache controls for more consistent refresh behavior.
  * Improved Next.js caching for components to better leverage cached renders.
* **Bug Fixes**
  * Middleware now sets the `pathname` cookie only on the first page load to avoid redundant updates.
* **Tests**
  * Updated sitemap bridge tests to use the new cached token-fetch pathway.


src/app/lib/tokens/cachedTokensFetch.ts
```diff
@@ -0,0 +1,20 @@
+import { unstable_cache } from 'next/cache';
```
oktapodia, line 1: seems that unstable_cache is deprecated in favor of 'use cache' https://nextjs.org/docs/app/api-reference/functions/unstable_cache

mmioana: Nice nice! Replaced the unstable cache 🎉

Kayanski (approved):

PR #2973, opened 2026-06-23
feat: add 7d/30d APY window toggle across earn - JUM-839

laurentsenta:
## Summary
Lets users view APY over a 7d or 30d window across the Earn surfaces. The choice persists in the URL (`?apyWindow`) and is shared between the markets filter bar and the opportunity detail card. When 30d data is unavailable, the card shows an explicit "Unknown" rather than a misleading 7d value.

## Changes
- `EarnApyWindowToggle` (TabSelect sliding pill) in the filter bar + a toggle button on the overview/detail APY tile (`OverviewEarnCardItem` gains `onClick`).
- `useApyWindow` hook persists the window in the URL; threaded into earn-card hrefs via `buildEarnHref`.
- New `getDisplayApy(latest, apyWindow)` centralizes window selection; wired into the display hook, sorting (`makeSortAccessors`), and filtering/histogram (`filterOpportunities`, `extractFilteringParams`) through `EarnFilteringContext`.
- Missing 30d -> "Unknown" (no fallback); regenerated `apy30d` types; new i18n keys; Storybook stories for filter bar, toggle, and overview card cases.

## Linked Linear ticket
JUM-839

## Sister PRs
- jumperexchange/jumper-backend (apy30d field): https://github.com/jumperexchange/jumper-backend/pull/977

## Summary by CodeRabbit

* **New Features**
  * Added an APY window toggle (7d/30d) across Earn pages, including filtering, sorting, and card displays.
  * Earn cards and opportunity links now preserve the selected APY window (via URL parameters).
  * Overview sections support interactive APY-window switching.
* **Bug Fixes**
  * APY display, sorting, and filtering now remain consistent with the selected window.
  * When 30d APY data is unavailable, the UI shows an explicit “unknown” state.
* **Tests**
  * Added Storybook and snapshot coverage for APY-window variants.
* **Documentation**
  * Updated i18n strings for the APY window selector (including an “unknown” fallback).


src/components/EarnFilterBar/EarnFilterBar.tsx
```diff
@@ -70,6 +80,9 @@ export const EarnFilterBar: React.FC<EarnFilterBarProps> = ({
             }}
           >
             <EarnListMode variant={variant} setVariant={setVariant} />
+            {hasApyWindow && (
+              <EarnApyWindowToggle value={apyWindow} onChange={setApyWindow} />
+            )}
```
mmioana: I feel adding this element here might make it pretty crowded, so we might consider having it in the same line with the view selector. Wdyt? 🤔 

Image

Also we need to make sure to show the selector (probably same dropdown pattern) is visible on mobile

Image

src/app/ui/earn/filterOpportunities.ts
```diff
@@ -90,7 +94,7 @@ export function filterOpportunities(
     }
 
     // Rewards APY filter (exclusive min to filter out 0 rewards)
-    const rewardsApy = item.latest.apy.jumperReward;
+    const rewardsApy = apy?.jumperReward;
```
Kayanski: (ai) **Bug: rewards filter incorrectly excludes vaults in 30d mode**

`rewardsApy` is now sourced from `getDisplayApy(item.latest, apyWindow)?.jumperReward`. When `apyWindow` is `30d` and the vault has no `apy30d`, `getDisplayApy` returns `undefined`, so `rewardsApy` is `undefined` here:

```ts
// current — breaks when apyWindow=30d and apy30d is absent
const rewardsApy = apy?.jumperReward;
```

The filter below then treats this as "no rewards" and returns `false`, silently dropping reward-bearing vaults from the list whenever a user selects **30d + With Rewards**.

`jumperReward` is a protocol-level incentive that has no relationship to the APY window. The fix is to read it directly from the 7d field regardless of window:

```ts
// fix — rewards are window-independent
const rewardsApy = item.latest?.apy?.jumperReward;
```

Same root cause applies to `extractFilteringParams` further down (~line 172) where `withRewards` is computed via `getDisplayApy`: if all reward-bearing vaults lack `apy30d`, the **With Rewards** filter option is hidden entirely in 30d mode.

src/components/EarnFilterBar/EarnFilterBar.tsx
laurentsenta: trying out options:
CleanShot 2026-07-01 at 18 25 09@2x
CleanShot 2026-07-01 at 18 25 29@2x

src/types/jumper-backend.ts
```diff
@@ -909,10 +917,11 @@ export interface EarnOpportunityWithLatestAnalytics {
   description: string;
   tags: string[];
   rewards: Token[];
+  messages: VaultMessage[];
   lpToken: Token;
   slug: string;
   featured: boolean;
-  lockupDays?: number | null;
+  lockupDays?: number;
```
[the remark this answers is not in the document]
laurentsenta: Acknowledged — the generated types currently narrow `lockupDays` to `number` while the backend Swagger defines it as `number | null`. This is a pre-existing gap in the generated types (not introduced by this PR). I'll apply the manual fix in the same style used for `apy30d`.

src/components/EarnFilterBar/EarnFilterBar.tsx
laurentsenta: Fixed — the toggle is now in the header row on the same line as the view selector (right-aligned), and rendered on all viewports including mobile.

src/app/ui/earn/filterOpportunities.ts
laurentsenta: Fixed — `jumperReward` is now always sourced from `item.latest?.apy?.jumperReward` (the window-independent 7d field) in both the per-item filter and the `withRewards` availability check in `extractFilteringParams`.

src/components/EarnFilterBar/EarnFilterBar.tsx
laurentsenta: CleanShot 2026-07-01 at 18 41 03@2x
CleanShot 2026-07-01 at 18 41 19@2x

laurentsenta: (thanks for the feedback @mmioana I'll have @Octave-byte review those)

Kayanski (approved):

src/components/EarnFilterBar/EarnFilterBar.tsx
laurentsenta: (reviewed with Octave, we're good to go)

yordanove:
Tested against the JUM-839 QA pass (full findings on the ticket: https://linear.app/lifi-linear/issue/JUM-839/add-30d-apy-toggle-across-earn-charts-overview-markets#comment-293286fe). One AC fail lands on this PR and I'd fix it before merge:

**Unknowns don't sort last once negative 30d APYs exist.** `makeSortAccessors` coerces missing `apy30d` to `0` (`getDisplayApy(item.latest, apyWindow)?.total ?? 0`) under the hardcoded DESC sort, so "Unknown" vaults land between positives and negatives. Live on the preview with 30d selected: the five Unknown vaults sit on page 11, and −0.19% / −42.8% / −100% follow them on page 12 — the AC says unknowns sort last.

jum839-unknown-sort-position


Also flagged on the ticket as questions rather than gates: the tooltip is one generic string for both windows (ticket asks for window-specific definitions), custom-reward vaults lose the "APR" label (now "APY (7d)"), hero cards ignore the toggle, and the detail-chart 30d AC isn't in this PR (needs a BE unblock too — scope call on the ticket).

Everything else held up well live: label toggle + URL param + deep links, all list/compact cards switching without refetch, filter bounds following the window (−2.89…38.24 ↔ −100…23.48), "Unknown" rendering, SSR/mobile clean.

mmioana (approved):

PR #2990, opened 2026-06-30
feat: add chart in limit order ui

Kayanski:
Closes JUM-1178

> [!NOTE]
> We need an event on the widget side to know when the limit has changes in order to display that on the Market Price chart (limitPrice prop)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,9 @@ yarn-error.log*
 /.pnp
 .pnp.js
 
+# Never commit auth (GitHub Packages PAT, etc.); copy from .npmrc.example locally.
+.npmrc
+
 # testing
 /coverage
 .cache-synpress/**
--- a/src/hooks/useUrlParams.tsx
+++ b/src/hooks/useUrlParams.tsx
@@ -1,5 +1,35 @@
 import { useEffect, useState } from 'react';
 
+declare global {
+  interface Window {
+    __jumperHistoryPatched?: boolean;
+  }
+}
+
+const URL_CHANGE_EVENT = 'jumperurlchange';
+
+/**
+ * `pushState`/`replaceState` never fire `popstate` (that only fires on
+ * browser back/forward), yet the widget updates `fromChain`/`fromToken`/etc.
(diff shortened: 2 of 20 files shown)


Kayanski:
Heads up: the backend was returning UDF `t` timestamps in seconds instead of milliseconds — now fixed in jumperexchange/jumper-backend#988. Since `CandlestickChart` in shared-ui divides `Bar.time` by 1000 internally, the datafeed here just needs to pass the ms values through as-is (which it already does). No change needed in this PR.

Also opened jumperexchange/jumper-shared-ui#6 to fix a related fixture bug where Storybook was generating negative bar timestamps.

src/hooks/useUrlParams.tsx
```diff
@@ -1,5 +1,35 @@
...
+
+const URL_CHANGE_EVENT = 'jumperurlchange';
+
+/**
+ * `pushState`/`replaceState` never fire `popstate` (that only fires on
+ * browser back/forward), yet the widget updates `fromChain`/`fromToken`/etc.
+ * in the query string via the History API as the user picks tokens. Patch
+ * both methods once so same-page URL changes are observable.
+ */
+function patchHistoryForUrlChangeEvent() {
```
chmanie, line 17: This looks wild but I was told it's fine :D

.gitignore
```diff
@@ -28,6 +28,9 @@ yarn-error.log*
 /.pnp
 .pnp.js
 
+# Never commit auth (GitHub Packages PAT, etc.); copy from .npmrc.example locally.
```
chmanie, line 31: Do we _have_ an `.npmrc.example`?

chmanie (approved):

src/hooks/useUrlParams.tsx
mmioana: I think I can find another workaround, but would stick with this in the meantime 🥺

.gitignore
mmioana: Adding it!

PR #2991, opened 2026-06-30
fix: align perk counters

chmanie:
### Problem
The Jumper Pass page reported different "unlocked perks" numbers in two places, and the number itself could be too low.

Two independent causes:
1. The top stats chip counted all unlocked perks, while the Unlocked Perks section counted only *still-claimable* ones (unlocked minus already-claimed), so the two disagreed (e.g. header 10, section 8).
2. The profile page and navbar fetched perks with `pageSize: 10`, sorted `Featured desc, UnlockLevel asc`. The unlocked count was computed over only those 10 perks, so it silently capped and undercounted (e.g. a wallet with >10 unlocked perks showing 6). This is unrelated to the loyalty level, which is correct.

### Fix
Reconcile every counter to "all unlocked perks", and stop capping the fetch.

- Unlocked Perks section now counts all unlocked perks (`unlockedPerks.length`), matching the stats chip and the navbar which already counted all unlocked.
- The section carousel now shows every unlocked perk: still-claimable ones first, then already-claimed ones rendered dimmed with a "Claimed" badge (the same card the Perks Hub already uses). Ordering uses a stable sort so incoming order is preserved within each group.
- Added `getAllPerks` (fetch the total count, then request them all) and routed the profile pages, the navbar's `usePerks`, and the Perks Hub through it, replacing the `pageSize: 10` fetch. Removed the now-unused `PAGE_SIZE` constant.

### Result
- Header 10 vs section 8: both now report all unlocked perks and agree.
- Header 6 vs actual >10: the fetch no longer caps at 10, so the count reflects the full catalog. The wallet's level was already correct.

### Testing
- `pnpm typecheck` and eslint pass.
- Exact catalog counts could not be verified from the dev sandbox (no network access to Strapi); worth a quick manual check on `develop` after deploy.

## Summary by CodeRabbit

* **New Features**
  * Perks now load the full available list on the perks and profile pages (and the perks hook), improving consistency across the app.
* **Bug Fixes**
  * Unlocked perks now render all unlocked items with a stable order that shows unclaimed perks before already claimed ones.
  * Empty states and displayed counts now reflect the total number of unlocked perks.
* **Refactor**
  * Standardized perks retrieval around a shared “all perks” fetch flow.

### DIFF
--- a/src/components/ProfilePage/sections/UnlockedPerksSection/UnlockedPerksSection.tsx
+++ b/src/components/ProfilePage/sections/UnlockedPerksSection/UnlockedPerksSection.tsx
@@ -25,6 +25,7 @@ import {
   openHubButtonSx,
   unlockedPerksCardSx,
 } from './UnlockedPerksSection.styles';
+import { sortBy } from 'lodash';
 
 interface UnlockedPerksSectionProps {
   perks: PerksDataAttributes[];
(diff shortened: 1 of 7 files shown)


src/components/ProfilePage/sections/UnlockedPerksSection/UnlockedPerksSection.tsx
```diff
@@ -46,14 +46,19 @@ export const UnlockedPerksSection = ({ perks }: UnlockedPerksSectionProps) => {
...
+
+  // Show every unlocked perk so the count here matches the Jumper Pass stat.
+  // Still-claimable perks come first; already-claimed ones follow, rendered
+  // dimmed with a "Claimed" badge (same card as the Perks Hub). Array.sort is
+  // stable, so the incoming order is preserved within each group.
+  const orderedPerks = [...unlockedPerks].sort(
+    (a, b) =>
+      Number(isClaimedPerk(a, claimedIds)) -
+      Number(isClaimedPerk(b, claimedIds)),
   );
```
mmioana: Would it make sense to use `sortBy` from lodash?

mmioana (approved):
🌟

PR #3001, opened 2026-07-06
feat: add order row actions (cancel, modify, repeat)

mmioana:
## Summary
- Adds Cancel / Modify / Repeat actions to the limit order row menu (`OrderRowMenu`), gated by order status (ongoing vs. completed/expired).
- Cancellation is fully wired: fetches cancel calldata, then either sends an on-chain tx (via `useTransactionFlow`, same chain-switch/execute/confirm state machine used elsewhere) or signs + relays an off-chain cancellation, with a confirmation modal and a `StatusBottomSheet` for pending/success/error feedback (height-animated, matching existing widget conventions).
- Modify and Repeat currently render placeholder "coming soon" modals until the real order-editing widget exists; they're wired through their own stores/components so swapping in the real widget later is a self-contained change.
- Adds a `Filled` badge to the expiry column, a shared `isOrderExpired` check reused by both the table and the row menu, and a `maxWidth` + ellipsis guard on the amount columns.
- Adds an `Error` variant to the shared `Button` component (`src/components/core/buttons`) for the destructive cancel action.

Contributes to JUM-1179
Closes https://linear.app/lifi-linear/issue/JUM-1174/cancel-order-modal-flow

## Test plan
- [ ] `pnpm typecheck` / `pnpm lint` / unit tests pass (verified locally via `tsc --noEmit`, `eslint`, `prettier --check`, `vitest run`)
- [ ] Manually open the orders table, exercise the row menu for an active order (Modify, Cancel with confirm/keep, retry-on-error) and a cancelled/expired order (Repeat)
- [ ] Confirm the cancel status sheet animates height correctly and doesn't close the modal when retrying after an error

### DIFF
--- a/src/components/composite/CancelOrderFlow/hooks/useCancelOrder.ts
+++ b/src/components/composite/CancelOrderFlow/hooks/useCancelOrder.ts
@@ -94,6 +94,8 @@ export const useCancelOrder = () => {
             ],
           })
           .catch(() => {
+            setIsAwaitingOnchainTx(false);
+            setOnchainChainId(undefined);
             throw new CancelOrderError(
               'Failed to send the cancellation transaction',
               CancelOrderErrorType.TransactionFailed,
--- a/src/components/composite/CancelOrderFlow/hooks/useCancelOrder.ts
+++ b/src/components/composite/CancelOrderFlow/hooks/useCancelOrder.ts
@@ -1,0 +1,184 @@
+'use client';
+
+import { useCallback, useState } from 'react';
+import { useMutation } from '@tanstack/react-query';
+import { useSignTypedData } from 'wagmi';
+import type { Hex } from 'viem';
+import { makeLimitOrderClient } from '@/app/lib/limitOrderClient';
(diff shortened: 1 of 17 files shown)


src/components/composite/CancelOrderFlow/hooks/useCancelOrder.ts
```diff
@@ -0,0 +1,184 @@
...
+          throw new CancelOrderError(
+            'Failed to fetch cancellation calldata',
+            CancelOrderErrorType.CalldataFailed,
+          );
+        });
+      const { transactionRequest, typedData, estimate } = calldataResponse.data;
+
+      if (transactionRequest) {
+        setOnchainChainId(transactionRequest.chainId);
+        setIsAwaitingOnchainTx(true);
```
chmanie, line 81: Do we need to reset these two if the execution fails?

chmanie (approved):

src/components/composite/CancelOrderFlow/hooks/useCancelOrder.ts
mmioana: Nice catch!

PR #3006, opened 2026-07-07
fix(e2e): paste wallet seed/password into MetaMask instead of typing

yordanove:
Closes JUM-1235

## Why

Playwright records `fill()`/`type()` values verbatim in its trace, and our e2e traces get uploaded to Qase. Typing the test wallet's seed phrase and MetaMask password therefore wrote those secrets into trace artifacts. Wallet secrets shouldn't live in test artifacts.

## What

Enter the seed and password by **pasting from the OS clipboard** (`pbcopy` on macOS, `xclip` on Linux) via Ctrl/Cmd+V, instead of typing. A pasted value travels OS clipboard → browser and never passes through a traced Playwright call, so the trace only records the keypress. MetaMask distributes a pasted multi-word phrase across the SRP inputs.

- `tests/e2e/wallet/utils/clipboard.ts` — new clipboard helpers. The value is passed on stdin (never argv, so it can't surface in a process list). On Linux, `xclip` daemonizes to serve the selection, so its stdout/stderr are detached and a hard timeout is set — otherwise the synchronous call blocks the run headless.
- `tests/e2e/wallet/pages/MetaMaskPage.ts` — a `pasteSecret()` helper; `fillSeedPhrase` and `setupPassword` paste instead of `fillSecret`.
- `.github/workflows/playwright.yml` — install `xclip` (needed to own the CLIPBOARD selection under xvfb).

## Validation

- A per-test trace contains **0** occurrences of the seed and **0** of the password; only non-secret form fields remain as `fill()` values.
- Verified in CI (headless, xvfb + xclip): `walletAddCustomNetwork` (qase 204) — the full seed-import-via-paste path — passes.

## Summary by CodeRabbit

* **Tests**
  * Improved wallet setup handling in end-to-end tests by using clipboard-based input for seed phrases and passwords, making test runs more reliable.
  * Reduced the chance of sensitive pasted values appearing in browser traces during automated testing.

* **Chores**
  * Updated the Playwright test environment to include required clipboard support for Linux-based runs.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -100,8 +100,10 @@ jobs:
       - name: Install Playwright Browsers
         run: |
           pnpm exec playwright install --with-deps chromium
-      - name: Install xvfb
-        run: sudo apt-get update && sudo apt-get install -y xvfb
+      - name: Install xvfb and xclip
+        # xclip owns the CLIPBOARD selection under xvfb; the wallet seed is pasted
+        # (not typed) into MetaMask so it never lands in a Playwright trace.
+        run: sudo apt-get update && sudo apt-get install -y xvfb xclip
       - name: Run Playwright tests
         env:
           QASE_TESTOPS_RUN_ID: ${{ needs.create-qase-run.outputs.run_id || '' }}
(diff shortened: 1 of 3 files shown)


yordanove:
@coderabbitai Intentional — `pbcopy` (macOS, local dev) and `xclip` (Linux, CI) are platform-specific equivalents, not duplicates, covering the only two OSes this suite runs on. `pbcopy` writes to the macOS pasteboard service and exits immediately, so it has none of `xclip`'s daemon-holds-the-pipe hang — which is exactly why only the `xclip` branch needs the `stdio`/`timeout` guards. There's no Windows target, so an unsupported-platform branch would guard a case that can't occur. Leaving as-is.

mmioana (approved):

PR #2552, opened 2025-12-22
feat: enhance GHA  to publish   reports to GH pages

callMeTheQA:
## Summary by CodeRabbit

* **New Features**
  * HTML Playwright test reports are automatically published to GitHub Pages.
  * Workflow summaries now include a "View Report" link to the deployed HTML report.
  * Pull request comments include the report link when available; a clear fallback message is shown until deployment completes.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #3007, opened 2026-07-07
ci: publish per-PR Playwright reports to GitHub Pages - JUM-1236

vinzenzLIFI:
## Summary

Give each PR (and each run/attempt) its own Playwright HTML report on GitHub Pages, with re-runs stacking instead of overwriting, plus a cron that keeps the branch bounded. Closes JUM-1236.

- **`publish-report` job** (`.github/workflows/playwright.yml`): deploys each run's merged HTML report to its own `gh-pages` path — `pr-<n>/<YYYYMMDD>-<run>-<attempt>/` (or `run-…` for `workflow_dispatch`) — via `peaceiris/actions-gh-pages` with `keep_files: true`, so reports accumulate rather than replace the whole site (which `actions/deploy-pages` would do).
- **Security**: strips the trace viewer UI + trace `.zip` attachments before publishing (keeps only `.png`/`.webm`). jumper-exchange Pages is public, and per JUM-1235 secrets can land in traces, so this is a security control. Publishing is gated to same-repo PRs (fork PRs get a read-only token).
- **PR comment**: links the run's own report URL (guarded so a skipped/failed publish never renders an empty link). The Qase link was dropped from the comment; Qase reporting itself (`create-qase-run`/`complete-qase-run`, `playwright-qase-reporter`, Slack failure link) is unchanged. Qase runs stay fully intact for now and are only slated for removal in a future change once we no longer need Qase.
- **`playwright-cleanup-reports.yml`** (new): daily cron that prunes report dirs older than 7 days by their in-path date (strict `YYYYMMDD` regex + year bound) and rebuilds `gh-pages` as a single orphan commit so git history / repo size stays bounded. Shares a `gh-pages-publish` concurrency group with the publish job so a publish push and the force-push can't race.

Supersedes #2552 (switches its single-site `deploy-pages` approach to branch accumulation). Close #2552 after this merges.

## Notes / review context

- Pages source stays "Deploy from a branch → `gh-pages` / root" (already configured, `build_type: legacy`); no settings change needed and no `pages`/`id-token` permissions added.
- The cleanup only ever touches `pr-*` / `run-*` report dirs; other Pages content (e.g. `.nojekyll`) is preserved. Nothing else in the repo publishes to `gh-pages` today.
- peaceiris pinned to the v4.0.0 commit `4f9cc66`.

## Test plan

- [ ] This PR's run publishes a report to `https://jumperexchange.github.io/jumper-exchange/pr-<this-pr>/<date>-<run>-<attempt>/` and the sticky comment links it.
- [ ] Re-running the workflow creates a NEW dir (new `<run>-<attempt>`) without overwriting the previous one.
- [ ] Published report has no `trace/` dir and no `data/*.zip` (screenshots/videos remain).
- [ ] After merge to `develop`: manually `workflow_dispatch` the cleanup workflow and confirm it prunes only aged report dirs and leaves `gh-pages` at a single commit.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * CI now generates, merges, and publishes Playwright HTML reports to the shared `gh-pages`, using stable per-run paths.
  * PR sticky comments link to the published report only when publishing succeeds.
* **Bug Fixes**
  * Reports are not published for pull requests from forks.
* **Documentation**
  * Added/expanded “CI reports” guidance: where reports appear, that trace assets are removed from the public `gh-pages` output, how the scheduled retention pruning works, and how to download/decrypt full artifacts for debugging.

### DIFF
--- a/.github/workflows/playwright-cleanup-reports.yml
+++ b/.github/workflows/playwright-cleanup-reports.yml
@@ -54,7 +54,7 @@ jobs:
           for dir in run-*; do
             [ -d "$dir" ] || continue
             age=$(age_days_from_name "${dir#run-}") || { echo "skip (undated): $dir"; continue; }
-            if [ "$age" -gt "$RETENTION_DAYS" ]; then echo "delete $dir ($age d)"; rm -rf "$dir" && deleted=$((deleted+1)); fi
+            if [ "$age" -ge "$RETENTION_DAYS" ]; then echo "delete $dir ($age d)"; rm -rf "$dir" && deleted=$((deleted+1)); fi
           done
           # PR reports: pr-<n>/<YYYYMMDD>-<run>-<attempt>
           for prdir in pr-*; do
@@ -62,7 +62,7 @@ jobs:
             for sub in "$prdir"/*; do
               [ -d "$sub" ] || continue
               age=$(age_days_from_name "$(basename "$sub")") || { echo "skip (undated): $sub"; continue; }
-              if [ "$age" -gt "$RETENTION_DAYS" ]; then echo "delete $sub ($age d)"; rm -rf "$sub" && deleted=$((deleted+1)); fi
+              if [ "$age" -ge "$RETENTION_DAYS" ]; then echo "delete $sub ($age d)"; rm -rf "$sub" && deleted=$((deleted+1)); fi
             done
             # drop the PR container once all its runs are pruned
             rmdir "$prdir" 2>/dev/null || true
--- a/.github/workflows/playwright-cleanup-reports.yml
+++ b/.github/workflows/playwright-cleanup-reports.yml
@@ -7,8 +7,6 @@ on:
   schedule:
     - cron: "0 0 * * *" # Daily at midnight UTC
   workflow_dispatch:
-permissions:
-  contents: write
 # Shared with playwright.yml's publish-report job so a publish push and this force-push
 # can never race on the gh-pages branch.
 concurrency:
@@ -20,6 +18,8 @@ jobs:
   cleanup:
     name: Delete old reports and compact history
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       - uses: actions/checkout@9f698171ed81b15d1823a05fc7211befd50c8ae0 # v6.0.3
         with:
--- a/.github/workflows/playwright-cleanup-reports.yml
+++ b/.github/workflows/playwright-cleanup-reports.yml
@@ -1,0 +1,84 @@
+name: Cleanup Playwright report directories
+# Deletes report dirs older than RETENTION_DAYS from the gh-pages branch AND squashes the
+# branch to a single orphan commit, so pruned reports leave no git-history residue and the
+# repo does not grow unbounded. Age is read from the YYYYMMDD prefix baked into each report
+# path (no reliance on git commit time, which the squash would destroy).
+on:
+  schedule:
+    - cron: "0 0 * * *" # Daily at midnight UTC
+  workflow_dispatch:
+permissions:
+  contents: write
+# Shared with playwright.yml's publish-report job so a publish push and this force-push
+# can never race on the gh-pages branch.
+concurrency:
+  group: gh-pages-publish
+  cancel-in-progress: false
+env:
+  RETENTION_DAYS: 7
+jobs:
+  cleanup:
+    name: Delete old reports and compact history
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9f698171ed81b15d1823a05fc7211befd50c8ae0 # v6.0.3
+        with:
+          ref: gh-pages
+          fetch-depth: 0
+      - name: Delete directories older than retention window
+        id: prune
+        shell: bash
+        run: |
+          set -uo pipefail
+          now=$(date -u +%s)
+          deleted=0
+          # Age in whole days, only for a name shaped EXACTLY as <YYYYMMDD>-<run>-<attempt>.
+          # Full-name regex + plausible-year bound so an 8-digit run_id from an old scheme
+          # can never be misread as a date and wrongly pruned/retained.
+          age_days_from_name() {
+            local name="$1"
+            [[ "$name" =~ ^([0-9]{4})([0-9]{2})([0-9]{2})-[0-9]+-[0-9]+$ ]] || return 1
+            local y="${BASH_REMATCH[1]}" mo="${BASH_REMATCH[2]}" da="${BASH_REMATCH[3]}"
+            [ "$y" -ge 2024 ] && [ "$y" -le 2100 ] || return 1
+            local epoch
+            epoch=$(date -u -d "$y-$mo-$da" +%s 2>/dev/null) || return 1
+            echo $(( (now - epoch) / 86400 ))
+          }
+          # workflow_dispatch reports: run-<YYYYMMDD>-<run>-<attempt>
+          for dir in run-*; do
+            [ -d "$dir" ] || continue
+            age=$(age_days_from_name "${dir#run-}") || { echo "skip (undated): $dir"; continue; }
+            if [ "$age" -gt "$RETENTION_DAYS" ]; then echo "delete $dir ($age d)"; rm -rf "$dir" && deleted=$((deleted+1)); fi
+          done
+          # PR reports: pr-<n>/<YYYYMMDD>-<run>-<attempt>
+          for prdir in pr-*; do
+            [ -d "$prdir" ] || continue
+            for sub in "$prdir"/*; do
+              [ -d "$sub" ] || continue
+              age=$(age_days_from_name "$(basename "$sub")") || { echo "skip (undated): $sub"; continue; }
+              if [ "$age" -gt "$RETENTION_DAYS" ]; then echo "delete $sub ($age d)"; rm -rf "$sub" && deleted=$((deleted+1)); fi
+            done
+            # drop the PR container once all its runs are pruned
+            rmdir "$prdir" 2>/dev/null || true
+          done
+          echo "deleted=$deleted" >> "$GITHUB_OUTPUT"
+          echo "Deleted $deleted report directories."
+      - name: Compact gh-pages history (single orphan commit)
+        shell: bash
+        run: |
+          set -euo pipefail
+          deleted="${{ steps.prune.outputs.deleted }}"; deleted="${deleted:-0}"
+          commits=$(git rev-list --count HEAD)
+          if [ "$deleted" -eq 0 ] && [ "$commits" -le 1 ]; then
+            echo "Nothing to delete and history already compact; skipping."
+            exit 0
+          fi
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git checkout --orphan compacted
+          git add -A
+          # --allow-empty guards the (unreachable while .nojekyll exists) empty-tree case
+          git commit --allow-empty -m "chore: playwright reports (compacted $(date -u +%F))"
+          git branch -D gh-pages 2>/dev/null || true # tolerate detached-HEAD checkout under set -e
+          git branch -m gh-pages
+          git push -f origin gh-pages
--- a/.github/workflows/playwright-cleanup-reports.yml
+++ b/.github/workflows/playwright-cleanup-reports.yml
@@ -21,12 +21,18 @@ jobs:
     permissions:
       contents: write
     steps:
-      - uses: actions/checkout@9f698171ed81b15d1823a05fc7211befd50c8ae0 # v6.0.3
+      # gh-pages may not exist yet (fresh repo, deleted branch); don't hard-fail the
+      # scheduled run when there's nothing to clean up.
+      - name: Checkout gh-pages
+        id: checkout
+        continue-on-error: true
+        uses: actions/checkout@9f698171ed81b15d1823a05fc7211befd50c8ae0 # v6.0.3
         with:
           ref: gh-pages
           fetch-depth: 0
       - name: Delete directories older than retention window
         id: prune
+        if: steps.checkout.outcome == 'success'
         shell: bash
         run: |
           set -uo pipefail
@@ -64,6 +70,7 @@ jobs:
           echo "deleted=$deleted" >> "$GITHUB_OUTPUT"
           echo "Deleted $deleted report directories."
       - name: Compact gh-pages history (single orphan commit)
+        if: steps.checkout.outcome == 'success'
         shell: bash
         run: |
           set -euo pipefail
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -173,10 +173,11 @@ jobs:
         run: echo "name=html-report--attempt-${{ github.run_attempt }}" >> "$GITHUB_OUTPUT"
       # The full report carries un-stripped traces (network bodies, DOM snapshots, console
       # logs), and artifact download on a public repo is open to any logged-in GitHub
-      # account — so encrypt at rest and share the key out-of-band. Plaintext is allowed
-      # ONLY for fork PRs (no secrets there, and no wallet secrets in their traces either);
-      # a same-repo run missing the key is a misconfiguration and fails loudly rather than
-      # silently uploading un-stripped traces unencrypted.
+      # account — so encrypt at rest and share the key out-of-band. AES-256 zip (not 7z,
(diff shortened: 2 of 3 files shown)


vinzenzLIFI:
@coderabbitai regarding the `gh-pages-publish` concurrency finding (`playwright.yml` 184-186): **acknowledged, won't fix for v1** — this is an intentional tradeoff.

Whole-branch serialization is required because both `publish-report` and the cleanup cron force-push `gh-pages`; the shared concurrency group is what prevents those pushes from racing and corrupting the branch. The known downside is exactly what you describe: during a burst, a newer publish can evict an older *pending* one, occasionally leaving a PR comment without a Pages link.

We're accepting this because:
- The failure mode is benign and self-healing — the report artifact is still uploaded, and re-running the workflow re-publishes and updates the sticky comment.
- The alternatives (an external publish queue, or a per-report worktree/publish path that avoids branch-level locking) are significantly more machinery than this v1 warrants.

If burst evictions turn out to be frequent in practice, we'll revisit with a proper queue. Tracking under JUM-1236.

.github/workflows/playwright-cleanup-reports.yml
```diff
@@ -0,0 +1,91 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      # gh-pages may not exist yet (fresh repo, deleted branch); don't hard-fail the
+      # scheduled run when there's nothing to clean up.
+      - name: Checkout gh-pages
+        id: checkout
+        continue-on-error: true
+        uses: actions/checkout@9f698171ed81b15d1823a05fc7211befd50c8ae0 # v6.0.3
```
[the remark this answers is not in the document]
vinzenzLIFI: @AikidoSec ignore: false positive for this step. The cleanup job checkout must keep persist-credentials enabled because the "Compact gh-pages history" step force-pushes with `git push -f origin gh-pages`, which authenticates via the checkout-persisted GITHUB_TOKEN (contents: write, scoped at job level). This matches Aikido’s own guidance to keep credentials only when the workflow performs authenticated git push operations. The publish-report checkout in playwright.yml is separately set to persist-credentials: false since peaceiris uses its own token.

.github/workflows/playwright.yml
```diff
@@ -168,7 +168,72 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@9f698171ed81b15d1823a05fc7211befd50c8ae0 # v6.0.3
+        with:
+          # peaceiris pushes with its own github_token below, so don't leave the default
+          # GITHUB_TOKEN persisted in .git/config for later steps to read.
+          persist-credentials: false
+      - name: Download HTML report
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: html-report--attempt-${{ github.run_attempt }}
```
yordanove: This downloads by exact name `html-report--attempt-${{ github.run_attempt }}`, but on "Re-run failed jobs" `run_attempt` increments while a green merge-reports doesn't re-run — so this asks for `attempt-2`, which was never uploaded, and every partial re-run fails the same way; the only recovery is re-running all jobs (full 5-shard suite, ~40 min). And a cancelled/failed publish is exactly the job you'd partially re-run — this is also what makes the accepted concurrency tradeoff's "self-healing by re-run" only hold via Re-run *all* jobs until fixed.

That's per documented re-run behavior — I haven't reproduced it live; easy to validate after merge if you want. Cleanest fix I can see: have merge-reports expose the attempt it uploaded under as a job output and download by that (outputs of non-re-run jobs survive partial re-runs), rather than binding to the current attempt.

.github/workflows/playwright.yml
```diff
@@ -168,7 +168,72 @@ jobs:
...
+          path: playwright-report
+      # Public Pages: strip the trace viewer UI and trace .zip attachments so secrets can't
+      # leak (see JUM-1235); keep only .png/.webm attachments.
+      - name: Strip trace data before publishing
+        run: |
+          echo "Report size before: $(du -sh playwright-report | cut -f1)"
+          rm -rf playwright-report/trace
+          if [ -d playwright-report/data ]; then
+            find playwright-report/data -type f ! -name '*.png' ! -name '*.webm' -delete
```
yordanove, line 266: Good fail-closed approach for the files. One coverage gap: the whitelist only governs files under `data/` — small text attachments (stdout, error-context, any `testInfo.attach` with a `body`) are embedded inline in `index.html` and publish unfiltered. I decoded the published report from this PR's own run: the inline stdout/error-context bodies are in there (clean today — selector logs and masked snapshots, I checked). But a future `testInfo.attach('api-response', { body })` walks straight past this control, and the comment "keep only .png/.webm attachments" reads as stronger than what's enforced.

Short term I'd just widen the comment + make it a convention that nothing attaches raw text bodies. Deeper fix if you ever revisit: generate the public report trace-free at merge time so the manifest is honest too — right now the published report still renders "trace" links pointing at `data/<sha>.zip` that 404 (checked live), which will confuse people clicking them.

.github/workflows/playwright.yml
```diff
@@ -202,7 +270,9 @@ jobs:
           message: |
             ${{ needs.merge-reports.outputs.summary }}
 
-            ${{ needs.create-qase-run.outputs.public_link != '' && needs.create-qase-run.outputs.public_link != null && format('📋 [View Detailed Qase Report]({0})', needs.create-qase-run.outputs.public_link) || '' }}
+            ${{ needs.publish-report.result == 'success' && needs.publish-report.outputs.report_url != '' && format('📊 [View Playwright Report]({0})', needs.publish-report.outputs.report_url) || '' }}
```
yordanove: Dropping the Qase link from the comment while keeping the Qase jobs means create-qase-run still flips every run's Qase report to **public** (`PATCH /run/WJ/<id>/public` — unauthenticated, whole-internet; that's the surface JUM-1235 lived on), and the link now surfaces nowhere except failure-only Slack. On every green run we mint a public link nobody sees.

Since Qase stays for now, I'd do one of: keep the Qase link in the comment until Qase actually goes, or drop the `create-public-link` step in this PR too. Minting it currently serves no one.

.github/workflows/playwright.yml
```diff
@@ -168,7 +168,72 @@ jobs:
...
           path: playwright-report
-          retention-days: 2
+          retention-days: 4
+
+  # Publishes each report to its own path on the gh-pages branch (keep_files accumulates
+  # across PRs/runs); the playwright-cleanup-reports.yml cron prunes + compacts it.
+  publish-report:
+    # Same-repo only: fork PRs get a read-only GITHUB_TOKEN (push would fail) and we must
+    # not host untrusted fork HTML on our public Pages domain.
+    if: ${{ !cancelled() && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) }}
```
yordanove: If merge-reports fails before its upload step (blob download flake, or its 3-min timeout), `!cancelled()` still starts publish-report, which then dies on the missing artifact — a guaranteed second red job stacked on the real failure. `needs.merge-reports.result == 'success' && <fork guard>` gives a clean skip instead, and you still publish when only tests fail (merge-reports succeeds in that case — that's the main use).

.github/workflows/playwright.yml
```diff
@@ -168,7 +168,72 @@ jobs:
...
+          # GITHUB_TOKEN persisted in .git/config for later steps to read.
+          persist-credentials: false
+      - name: Download HTML report
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: html-report--attempt-${{ github.run_attempt }}
+          path: playwright-report
+      # Public Pages: strip the trace viewer UI and trace .zip attachments so secrets can't
+      # leak (see JUM-1235); keep only .png/.webm attachments.
+      - name: Strip trace data before publishing
```
yordanove, line 261: Question / policy check, no blocker: the kept `.png` includes wallet-context failure screenshots — the published report from this PR's runs already has MetaMask account-overview and connect-popup shots on the public domain (I pulled them and looked). Nothing sensitive in them today: MetaMask masks SRP/password inputs, and there's no wallet video at all (the persistent context doesn't record). Are we OK with wallet-UI screenshots being public as standing policy? Fine either way — just want it to be a conscious yes rather than a side effect.

yordanove (commented):
Went through this in depth, read the Zeus/CodeRabbit threads first, and also poked the live gh-pages output from this PR's own runs (anonymous curl, no auth). Overall solid — fork-PR guard, `persist-credentials: false`, pinned SHAs, the shared concurrency group is the right instinct, and stripping traces before a public deploy is exactly what JUM-1235 needed. Nothing below duplicates the earlier reviews.

One thing I'd fix before merge (the re-run trap — it also weakens the accepted concurrency tradeoff, see the conversation thread), one decision to make now (the orphaned public Qase link), rest is take-or-leave.

Minor nits, no response needed: the `actions/checkout` in publish-report looks unused (nothing reads repo files — worth trying without it); the cleanup's `fetch-depth: 0` pulls the full binary-heavy history daily just to feed the `rev-list --count` guard — if that ever gets slow, note a shallow clone would truncate the count and fool the `<= 1` check, so the guard needs rethinking first; the final `branch -D` / `branch -m` dance can be one line (`git push -f origin compacted:gh-pages`); and `report_url != ''` in the comment expression is unreachable given `result == 'success'`.

.github/workflows/playwright.yml
```diff
@@ -168,7 +168,72 @@ jobs:
         with:
           name: html-report--attempt-${{ github.run_attempt }}
           path: playwright-report
-          retention-days: 2
+          retention-days: 4
```
yordanove, line 206: 2→4 days on the un-stripped artifact plus a comment link advertising it on every PR — fine by me now that #3006 keeps wallet secrets out of traces, just noting the trade consciously: traces still carry full network logs, and artifact download on a public repo is any-logged-in-GitHub-account. Maybe add "(kept 4 days)" to the comment text like the README already says, so people know why the link goes dead on quiet PRs.

.github/workflows/playwright-cleanup-reports.yml
```diff
@@ -0,0 +1,91 @@
...
+            echo $(( (now - epoch) / 86400 ))
+          }
+          # workflow_dispatch reports: run-<YYYYMMDD>-<run>-<attempt>
+          for dir in run-*; do
+            [ -d "$dir" ] || continue
+            age=$(age_days_from_name "${dir#run-}") || { echo "skip (undated): $dir"; continue; }
+            if [ "$age" -gt "$RETENTION_DAYS" ]; then echo "delete $dir ($age d)"; rm -rf "$dir" && deleted=$((deleted+1)); fi
+          done
+          # PR reports: pr-<n>/<YYYYMMDD>-<run>-<attempt>
+          for prdir in pr-*; do
```
yordanove, line 60: The prune only visits `run-*`/`pr-*`, so the leftover `2552/` dir from the #2552 experiment matches neither glob and gets re-committed into every daily squash forever. One manual delete on gh-pages after this merges (or when you close #2552) sorts it.

yordanove:
@vinzenzLIFI late addition on the accepted `gh-pages-publish` tradeoff — two data points that might change the calculus:

1. There's now a one-line middle ground: `queue: max` on the concurrency group. Default pending behavior is cancel-and-replace ("any existing pending job or workflow run in the same group is canceled and replaced"), but `queue: max` lets pending jobs queue up instead — while still keeping at most one *running* job, so the publish/cleanup force-push serialization you need is fully preserved. No external queue, no worktree machinery.

2. The self-healing premise has a catch: "Re-run failed jobs" after an evicted/failed publish can't succeed — publish downloads `html-report--attempt-${{ github.run_attempt }}`, the partial re-run bumps the attempt, merge-reports doesn't re-run, and the attempt-2 artifact doesn't exist (left a separate comment on that line). So healing currently requires re-running the full 5-shard suite.

With both fixed the tradeoff mostly disappears. Still fine as v1 if you'd rather land it — just wanted the full picture on the record.

.github/workflows/playwright.yml
vinzenzLIFI: Fixed in 2a15b66af — merge-reports now exposes the artifact name it uploaded as a job output (`html_report_name`), and publish-report downloads by that instead of recomputing `github.run_attempt`. "Re-run failed jobs" now recovers a failed publish without re-running the shard matrix.

.github/workflows/playwright.yml
vinzenzLIFI: Went with your short-term suggestion (2a15b66af): comment now states explicitly that inline-embedded attachments (stdout, error-context, `testInfo.attach` with a `body`) are NOT covered, and the no-raw-text-bodies convention is written down. The trace-free-at-merge-time deeper fix stays on the list for when Qase is gone and Pages is the only report home.

.github/workflows/playwright.yml
vinzenzLIFI: Agreed the minted link served no one — as of 2a15b66af it also no longer surfaces in the failure Slack ping, so nothing Qase-related is user-visible anywhere anymore (only in workflow run logs). The `create-public-link` step itself goes away in the immediate follow-up PR that removes Qase entirely; kept it here only to keep this PR scoped to report publishing.

.github/workflows/playwright.yml
vinzenzLIFI: Fixed in 2a15b66af — publish-report now requires `needs.merge-reports.result == 'success'` instead of `!cancelled()`, so a merge-reports failure gives a clean skip rather than a second red job. Still publishes when only tests fail, as intended.

.github/workflows/playwright.yml
vinzenzLIFI: Conscious yes for now: screenshots are failure-only, MetaMask masks SRP/password inputs, and the seed is pasted (never typed, #3006) so it can't appear mid-keystroke. Revisit if we ever add wallet flows that render balances/addresses we'd rather not publish — the private-reports-repo setup is the likely end state after Qase is decommissioned.

.github/workflows/playwright.yml
vinzenzLIFI: Added "(kept 4 days)" to the comment link text (2a15b66af). Also went one further on the any-logged-in-account concern: the full artifact is now encrypted at rest (aes-256-cbc via `PLAYWRIGHT_REPORT_ENCRYPTION_KEY` repo secret, passphrase in the 1Password Developers vault; decrypt steps in tests/README.md). Falls back to plaintext when the key is absent (fork PRs), so it's order-safe.

.github/workflows/playwright-cleanup-reports.yml
vinzenzLIFI: Correct — it matches neither glob and would ride along in every squash. Will delete `2552/` manually on gh-pages right after this merges; no workflow change needed.

vinzenzLIFI:
@lifi-qa-agent both items addressed:

1. **Retention off-by-one** — fixed in 36d578fba: `-gt` → `-ge` in both the `run-*` and `pr-*/sub` checks, so a report is deleted at exactly `RETENTION_DAYS`, matching the variable name and the "(kept 7 days)" comment text.
2. **PR title** — renamed from `feat(ci):` to `ci: publish per-PR Playwright reports to GitHub Pages - JUM-1236`.

yordanove (approved):
Re-reviewed after the updates — everything checks out. Verified this round, not just read:

- Latest run ([28945940026](https://github.com/jumperexchange/jumper-exchange/actions/runs/28945940026)) green end-to-end through the encryption path; sticky comment renders both links with retention labels.
- The artifact is genuinely WinZip AES-256 (method 99; passwordless read refused). One inherent zip-format note, informational only: entry **names** are plaintext without the key — harmless here (hashed attachment names + `index.html`), just never encode anything sensitive into an attachment filename.
- The double-click claim: reproduced the exact artifact structure locally and confirmed on macOS — Archive Utility prompts and decrypts it fine. (I doubted this one, tested it, it holds — `ditto`/CLI still can't, so the README's `7zz` fallback earns its place.)
- Re-run trap fix is correct — job outputs survive partial re-runs, so "Re-run failed jobs" now recovers a failed publish without the 40-min matrix.
- Nice catch on the implicit-`success()` transitivity in the publish gate — our suggested `result == 'success'`-only condition would have skipped publishing on test failures, which is when reports matter most.

Two things to hold post-merge (both committed to in threads): delete the legacy `2552/` dir from gh-pages, and drop `create-public-link` in the Qase-removal follow-up. `queue: max` stays on the record as the one-liner if burst evictions ever show up in practice — non-blocking now that re-runs heal them.

PR #3008, opened 2026-07-08
fix(e2e): stabilize swap-pair specs against deeplink token-resolution race

yordanove:
Closes JUM-1237

## Why

`swapActions.spec.ts` qase 23/24/26/27 (ETH/ARB swap pairs, both viewports) fail intermittently in CI: `getByText('Best Return')` times out, failure screenshots show From/To stuck on "Select…" with **zero route requests fired**.

Root cause: the widget only enables the route query once both deeplinked token **addresses** resolve into token objects against the loaded token list. Under CI shard load the list loses that race against the fixed 30s route timeout — the deeplink is parsed fine, but no routes are ever requested. Secondary: the specs dismissed the welcome overlay with a single un-retried "Get Started" click, which races React hydration.

## What

- `LandingPage.expectSwapPairResolved(pair)` — waits for the resolved swap pair before asserting routes, called after each deeplink `goto` in the four specs. Anchors on the **chain-name subheader** (renders only once chain AND token resolve), not the token symbol — the symbol drifts between environments (develop pipeline serves `USDT` where prod serves `USDT0` for the same Arbitrum token).
- `tests/e2e/utils/welcomeScreen.ts` — `seedWelcomeScreenClosed(page)` pre-seeds `welcomeScreenClosed` into the persisted settings store via `addInitScript` (merge, not overwrite), so the overlay never mounts; replaces the click in the swapActions `beforeEach`. Also ~5s faster per test.

## Validation

Two consecutive silent `workflow_dispatch` runs on this branch — qase 23/24/26/27 all green **on first attempt** (8/8, retries unused):

| | qase 23 | qase 24 | qase 26 | qase 27 |
|---|---|---|---|---|
| Baseline (recent develop-based runs) | ❌ fail | ✅ | ⚠️ flaky | ❌ fail |
| [28936639952](https://github.com/jumperexchange/jumper-exchange/actions/runs/28936639952) | ✅ 1st | ✅ 1st | ✅ 1st | ✅ 1st |
| [28937563259](https://github.com/jumperexchange/jumper-exchange/actions/runs/28937563259) | ✅ 1st | ✅ 1st | ✅ 1st | ✅ 1st |

Remaining flaky in those runs are unrelated (mainMenu qase 22/9, portfolioPage qase 52 — pre-existing, tracked separately).

## Noted during investigation, out of scope

- If platform requests (chains/tokens) **error** rather than load slowly, the widget parks on "Select…" with no retry — a failed deeplink is a dead widget for users too. Potential product finding.
- `SettingsPage` `getByText('Bridges', { exact: true })` strict-mode violation (2 elements) under load — pre-existing fragility.

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end coverage for swap flows with stronger checks that token pair details resolve correctly before validating routes.
  * Added more reliable setup for the welcome screen state to reduce test flakiness in automated runs.
  * Introduced chain ID to display-name mapping to support clearer validation of linked token/network information.

### DIFF
--- a/tests/e2e/utils/welcomeScreen.ts
+++ b/tests/e2e/utils/welcomeScreen.ts
@@ -1,0 +1,30 @@
+import type { Page } from '@playwright/test';
+
+// Mirrors the zustand persist config in src/stores/settings/createSettingsStore.tsx.
+const SETTINGS_STORE_KEY = 'jumper-store';
+const SETTINGS_STORE_VERSION = 4;
+
+/**
+ * Seeds the persisted settings store so the welcome overlay never mounts.
+ * Dismissing it by clicking races React hydration under CI load (a single
+ * missed "Get Started" click leaves the overlay up); pre-seeding removes the
+ * interaction entirely. Call before the test's first goto().
+ */
+export async function seedWelcomeScreenClosed(page: Page): Promise<void> {
+  await page.addInitScript(
+    ({ key, version }) => {
+      const existing = JSON.parse(
+        globalThis.localStorage.getItem(key) ?? 'null',
+      );
+      globalThis.localStorage.setItem(
+        key,
+        JSON.stringify({
+          version,
+          ...existing,
+          state: { ...existing?.state, welcomeScreenClosed: true },
+        }),
+      );
+    },
+    { key: SETTINGS_STORE_KEY, version: SETTINGS_STORE_VERSION },
+  );
+}
--- a/tests/e2e/data/urls.ts
+++ b/tests/e2e/data/urls.ts
@@ -38,6 +38,12 @@ export const CHAINS = {
 
 export type ChainName = (typeof CHAINS)[keyof typeof CHAINS];
 
+// Deeplink chain id → LiFi display name (the widget token card's subheader).
+export const CHAIN_NAMES_BY_ID: Record<string, string> = {
+  '1': 'Ethereum',
+  '42161': 'Arbitrum',
+};
+
 export const JUMPER_BUTTONS = {
   CONNECT: 'Connect',
   PASS: 'Pass',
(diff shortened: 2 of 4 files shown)


yordanove:
Re the review — non-gating, but for the record:

**Finding 1 (sequential waits / `Promise.all`)** — declined as-is: both cards resolve from the same token-list load, so once the list lands both assertions pass in the same render pass — the second wait is ~0 on every realistic path, and sequential keeps failure attribution unambiguous. Will revisit if a CI trace ever shows both waits independently approaching 30s.

**Finding 2 (Hyperliquid un-fixme trap)** — declined: the fixme block doesn't call `expectSwapPairResolved`, so re-enabling it as-is hits nothing; and if someone adds the wait without the mappings, `chainNameOf` fails fast with an error that names the exact fix ("add it to `CHAIN_NAMES_BY_ID` in data/urls.ts"). That's the guard working as designed.

**Spread-order note** — `version` before `...existing` is deliberate: if the app ever writes a newer store version, the seed preserves it instead of stamping it back to 4 and forcing a wrong migration path.

**`closeWelcomeScreen` migration follow-up** — accepted: tracked in [JUM-1238](https://linear.app/lifi-linear/issue/JUM-1238/e2e-migrate-remaining-welcome-screen-dismissals-to). Follow-up PR will migrate the remaining 6 specs (10 call sites) to `seedWelcomeScreenClosed` and retire the click path, except where the welcome screen itself is under test (landingPage.spec).

tests/e2e/utils/welcomeScreen.ts
```diff
@@ -0,0 +1,30 @@
+import type { Page } from '@playwright/test';
+
+// Mirrors the zustand persist config in src/stores/settings/createSettingsStore.tsx.
+const SETTINGS_STORE_KEY = 'jumper-store';
+const SETTINGS_STORE_VERSION = 4;
```
mmioana, line 5: Maybe we should have a common constant files so both the store and the test re-use the store key and version. But this can be done in a follow up

mmioana (approved):
Nice 🙏

PR #3009, opened 2026-07-08
fix(e2e): migrate welcome-screen dismissals to the persisted-store pre-seed

yordanove:
Closes JUM-1238

## Why

JUM-1237 (#3008) introduced `seedWelcomeScreenClosed()` — pre-seeding `welcomeScreenClosed` into the persisted settings store so the overlay never mounts — but only migrated `swapActions.spec.ts`. Everywhere else still dismisses the overlay with a single un-retried "Get Started" click, which races React hydration under CI load (the source of the historical `.welcome-screen-container still visible` failures).

## What

Migrated the remaining pure-setup dismissals — 6 specs, 8 call sites — to the seed (always placed before the navigation it applies to).

**Deliberately kept the real overlay + click where the welcome screen itself is under test:**
- `landingPage` qase 1 (overlay re-shows on logo click — `ClientNavbar` flips the store at runtime, so behavior against a seeded store is identical, verified)
- `mobileView` qase 5 (mobile welcome elements + "welcome can be closed")
- `mobileView` qase 4 untouched entirely — it measures the welcome view's viewport fit, so no seed at all

`closeWelcomeScreen()` therefore stays in the POM with those two users — it retires from *setup*, not from existence. Two now-dead `landingPage` locals removed.

## Validation

- Slack-silent `workflow_dispatch` [28948638532](https://github.com/jumperexchange/jumper-exchange/actions/runs/28948638532): all 5 shards green, every migrated spec **first-attempt** green (incl. settings qase 7 Mobile and both welcome-screen-under-test cases). Only flaky: earnPage qase 39 (pre-existing, unrelated). (An earlier dispatch was cancelled — shard 1 died at `webServer was not able to start`, the day's recurring upstream build blip, zero tests ran.)
- Local prod runs: migrated specs pass sequentially; the two failures that appear under a 5-file single-worker batch (`mobileView` qase 5 stats counters, `settings` qase 7 Mobile) **reproduce identically on develop without this diff** — pre-existing batch-profile artifact, A/B-exonerated.

## Follow-ups (tracked, out of scope)

- JUM-1239 — shared persist key/version constants between store and tests (Ioana's #3008 suggestion)
- JUM-1240 — wallet-framework hygiene (dead unlock path + clipboard cleanup hardening)

## Summary by CodeRabbit

* **Tests**
  * Updated end-to-end test setup to start with the welcome screen already closed.
  * Improved consistency across landing page, menu, settings, theme, mobile, and wallet connection flows.
  * Simplified test steps by removing repeated welcome-screen dismissal actions and standardizing the setup using a shared seeding helper.

### DIFF
--- a/tests/e2e/connectWallet.spec.ts
+++ b/tests/e2e/connectWallet.spec.ts
@@ -5,6 +5,7 @@ import { realWalletTest as test } from './fixtures/realWallet';
 import { ConnectWalletPage } from './pages/ConnectWalletPage';
 import { LandingPage } from './pages/LandingPage';
 import { ProfilePage } from './pages/ProfilePage';
+import { seedWelcomeScreenClosed } from './utils/welcomeScreen';
 
 test.describe('Connect/disconnect MetaMask with Jumper and open /profile', () => {
   // JUM-1116: flaky in CI (6/15) — the real-MetaMask connect + post-connect Perks tab
@@ -14,6 +15,7 @@ test.describe('Connect/disconnect MetaMask with Jumper and open /profile', () =>
     qase(36, 'Connect MetaMask wallet to Jumper'),
     async ({ wallet }) => {
       const page = await wallet.getContext().newPage();
+      await seedWelcomeScreenClosed(page);
       await page.goto('/');
       await page.waitForLoadState('domcontentloaded');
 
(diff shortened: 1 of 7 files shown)


yordanove:
N1 — good catch, and not intentional scoping: the sweep grepped `--include="*.spec.ts"`, so the fixture never matched. Fixed in 2d14743b2 — the seed now runs in the `jumperPage` fixture before its `goto`, covering the whole `connectedTest` suite; validated end-to-end with a real MetaMask connect + earn filter run. N2 — agreed, keeping the double navigation. N3 — confirmed pre-existing, unrelated to this diff.

mmioana (approved):

PR #3010, opened 2026-07-08
chore: bump widget and sdk

mmioana:

Contributes to JUM-1241

## Summary by CodeRabbit

* **Chores**
  * Updated several SDK- and widget-related dependencies to newer compatible versions.

* **Tests**
  * Improved Settings page E2E coverage to support drill-down “list option” rows, including new helpers to check visibility and selection.
  * Extended slippage settings testing with additional preset options and custom slippage validation, including warning checks.
  * Refined route-priority and slippage flows to use list-option interactions and streamlined navigation back, reducing test flakiness.


yordanove:
QA smoke on the preview — no blockers at this level, one e2e migration needed on our side.

**Verified on the preview:**
- Quote round-trip works (Base USDC→WETH: `/advanced/routes` 200, routes render, order matches payload)
- Route priority verified end-to-end through the NEW settings UI: selecting Fastest persists (`routePriority: FASTEST` in the widget store) and a fresh load sends `options.order: FASTEST` to the API; reverting to Best Return works the same way
- Wallet connect list: MetaMask appears exactly once — the 4.2.0 MetaMask-Bitcoin dedupe behaves, and no phantom BTC entry shows since the connector isn't registered (correct for this PR's scope)
- `/private` tab: private routes 200, renders fine
- Dark + MegaETH partner theme render correctly throughout (MUI 9.2 internally); light mode isn't reachable while the partner takeover pins `availableThemeModes: ["dark"]` — pre-existing behavior, not this bump
- Cards layout and limit-order mode correctly NOT active (opt-in configs we don't set)

**The shard-4 e2e failure is real, not flake** — and it's the one action item. Widget 4.2.0 moved route priority and slippage to dedicated drill-down pages: the options are list rows now, not inline buttons, so `settings.spec.ts`'s `//button[normalize-space(text())="Best Return"]` locator finds nothing (failed 3/3 deterministically). Slippage also has a new option set (Auto "Optimised per route" / 0.5% / 1% / Custom). Gas price kept the old accordion. We'll push the test migration (SettingsPage POM + the two affected steps) from QA shortly — it may also retire the JUM-1116 Desktop quarantine, since the step it quarantined no longer exists in that form.

**Post-merge:** since this bumps the execution stack (wagmi 3.7 / viem 2.54 inside the widget + all ecosystem providers), we'll run a thorough pass on develop beyond this smoke — real swap execution, the Earn widget variant, a Solana quote via Phantom, and a cross-chain route.

Minor upstream observation: on the new route-priority page, "Best Return" is subtitled "Lowest network cost" — that describes cost, not return ("Fastest / Shortest execution time" shows what the pattern intends). Widget copy from lifinance/widget#792, probably worth a nudge upstream rather than anything here.

yordanove:
Settings e2e migration for 4.2.0 is on the branch, as promised — the suite is green here now.

**What changed** (`bcf678b05`): route priority and slippage moved to drill-down pages in widget 4.2.0, so the old accordion-based steps matched nothing. The POM got drill-down row helpers (rows are `ListItemButton` divs with `role="button"`; selected state asserted via the check-icon svg — MUI v9 icons carry no testid), the slippage step was rewritten for the new option set (Auto / 0.5% / 1% / Custom with the collapsed input), and the slippage input re-anchored to its new placeholder. Gas price / bridges / exchanges / reset steps untouched — `GasPriceSettings` and `SelectEnabledToolsPage` are byte-identical between 4.1.0 and 4.2.0. Every locator was prototyped against the preview DOM before writing. Also merged develop into the branch (`1ba5619e5`) so CI runs the current test state.

**One experiment worth recording** (`428230ed6`): I removed the JUM-1116 qase 8 Desktop fixme to see whether the old quarantined stall was the replaced UI's fault. Result: the migrated steps now pass on CI-Desktop (first time ever), but the load stall persists further downstream — three runs, three *different* dying steps (bridges checkbox → slippage Custom row → back arrow; 120s click/render freezes, CI-Desktop only, Mobile + local Desktop unaffected). That's environment, not tests or the widget — so Desktop is re-quarantined with an accurate reason, and the three runs are fresh evidence for JUM-1116.

Validation: local 6/6 both viewports (retries=0); CI Mobile green through the full migrated flow in three consecutive runs; final state green on [dispatch 28961553868](https://github.com/jumperexchange/jumper-exchange/actions/runs/28961553868) and this PR's own run. Merge-ready from the QA side.

yordanove (approved):

PR #3021, opened 2026-07-09
ci(e2e): build the app once per run and share it across shards

yordanove:
## Problem

Each of the 5 Playwright shards builds the app itself inside the webServer (`pnpm run build && pnpm run start`). The build prerenders against live Strapi/backend, so a single transient upstream blip aborts the whole build and the shard dies with 0 tests (`Export encountered an error... exiting`) — 4 such incidents on 2026-07-08 alone. Five builds per run means five chances to lose that lottery, and a build failure shows up as confusing shard noise instead of a clear cause.

## Change

- New `build` job: builds once (loading `tests/.env.test` the same way Playwright's webServer does, one retry for the transient prerender-fetch case), packs `.next` minus `.next/cache` as a zstd tar (**128 MB**), uploads it as an artifact.
- Test shards `needs: build`, download + unpack (**~9 s**), and boot `next start` directly via a new `E2E_SKIP_BUILD` mode in `playwright.config.ts`. `E2E_PROD_BUILD` stays set so `next.config.mjs` resolves the same non-standalone output the build was produced with. Local run modes unchanged.
- A failed build is now a single red job: `merge-reports` skips when the shards never ran, and `post-comment` requires merge-reports success so a no-result run can't overwrite the previous sticky summary with an empty one.

Also folded in, surfaced while walking the re-run paths:
- `overwrite: true` on the blob-report upload — "Re-run failed jobs" keeps prior-attempt artifacts (merge-reports needs the passing shards' blobs), so a re-run shard must replace its own blob; without it the re-upload 409s and the run can never converge to green.
- Build artifact retention 4 days (matching the html report) — "Re-run failed jobs" doesn't re-run the succeeded build job, so a late re-run must still find the artifact.
- Dropped the test step's stale `if: !cancelled()` — it predates the artifact handover (the step could self-heal by building); now it would only stack a doomed `next start` failure on top of a real download error.

## Validation

- workflow_dispatch on this branch: https://github.com/jumperexchange/jumper-exchange/actions/runs/29019894070 — all green first attempt. Build compile 76 s; artifact upload 2 s; per-shard download+unpack 9 s (replaces the per-shard build); Qase run + Pages report publish intact.
- `E2E_SKIP_BUILD` path also verified locally against a prebuilt `.next`.

Trade-off: a green run pays a small serial build hop (~+5 min); in exchange the run rolls one retried build lottery instead of five, re-runs reuse the artifact instead of rebuilding, and build failures get clean attribution.

Linear: JUM-1242

## Summary by CodeRabbit

* **New Features**
  * Playwright end-to-end runs now reuse a shared, prebuilt application artifact across shards to speed up execution.

* **Bug Fixes**
  * Improved rerun behavior so shard report artifacts are reliably overwritten.
  * Tightened merge-report and sticky comment conditions to avoid stale/empty PR comments when shards are skipped due to build failures.

* **Improvements**
  * Added `E2E_SKIP_BUILD` support for the Playwright web server to reduce unnecessary rebuilds, including adjusted server startup timeouts.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -74,13 +74,67 @@ jobs:
 
           echo "public_link=$public_link" >> "$GITHUB_OUTPUT"
           echo "Public link: $public_link"
+  # One build shared by all shards: the build prerenders against live Strapi/backend,
+  # so every extra build is another chance for a transient upstream blip to abort a
+  # shard ("Export encountered an error"). Building here also turns a build failure
+  # into a single red job instead of every shard failing with 0 tests.
+  build:
+    name: Build App
+    runs-on: ubuntu-latest
+    # Covers the retry path: install + two full build attempts + pack/upload.
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@9f698171ed81b15d1823a05fc7211befd50c8ae0 # v6.0.3
+      - uses: pnpm/action-setup@008330803749db0355799c700092d9a85fd074e9 # v6.0.9
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 24
+          cache: "pnpm"
+      - name: Install pnpm
+        run: npm install -g pnpm
+      - name: Install dependencies
+        run: pnpm install
+      - name: Build production bundle
+        env:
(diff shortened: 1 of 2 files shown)


yordanove:
Good catch @coderabbitai — removed in 2cfe035e7, from the pre-existing test job as well so both match merge-reports (which already ran without it). With `packageManager` pinning pnpm@11.5.1, the global install could shadow the pin.

oktapodia (approved):
Looks promising!

PR #3027, opened 2026-07-13
feat: add stable data-testids for e2e selectors

yordanove:
## What

Adds the app-side `data-testid`s from JUM-924 (items 2, 3, 8, 9, 11, 13, 14 — items 1/4/5/6/7/12 are widget-side and ship via the jumper-widget fork; item 15 already exists in widget ≥4.2.0; item 10 can't be tagged here, the `/scan` rows render inside `@lifi/explorer`), plus the test-side selector swaps in the same PR so CI proves them end-to-end.

New testids:

| Testid | Where |
|---|---|
| `missions-list` | missions grid container (`MissionsPage` + skeleton) |
| `mission-verify-button` | task verify badge (`MissionTask`) |
| `mission-details` | mission detail root (`MissionDetails` + skeleton) — the list and detail previously shared only `mission-card-*`, so tests couldn't prove navigation |
| `perks-card-claimed-badge` | perk status badge, claimed state only (`PerkCard`) |
| `earn-card` | loaded earn card container (Compact/Hero/ListItem variants) |
| `earn-card-chain-name` | chain hint on loaded earn cards, via new optional `hintDataTestId` prop on `EntityStackWithBadge` (defaults to `entity-stack-hint` — all other consumers unchanged, snapshot run confirms) |
| `earn-card-tag-*` | added to ListItemEarnCard tags (parity with Compact/Hero) |
| `learn-page` | learn page root (+ skeleton) |
| `homepage-stat-{chains,bridges,dexs}-count` | welcome-screen stat counters (`ToolCard`, `tool.id` threaded) |
| `welcome-screen` | welcome overlay (via new `overlayDataTestId` prop on `WelcomeOverlayLayout`, mirrors `overlayClassName`) |

Test-side: swapped the matching weak selectors (MUI classes / xpath / text fallbacks), un-fixme'd **qase 40, 41, 42, 43, 208** (qase 40 rescoped: its cold-start flake was cured by the June earn latency fixes — verified 3x locally — and its Your-Positions-filters step moved conceptually to qase 56, which is now explicitly funded-wallet-gated: an empty wallet renders no filter selects on that tab), fixed the qase 9 oracle (now asserts `mission-details` after the card click instead of re-querying the same list selector), and corrected the fixme'd qase 203 flow (verify buttons live on the detail page; the old `name: 'Verify wallet'` locator never matched the rendered "Verify" label). Remaining `TODO(app)` markers are the widget-side items — updated to reference the ids that ship in the jumper-widget fork.

Reviving tests that hadn't executed since May also surfaced four latent test bugs, fixed here:

- `EarnPage.expectAllCardsShowChain` / `expectAllCardsHaveTag`: sync `.count()` snapshot + `nth(i)` loop raced the grid re-rendering mid-iteration (qase 42 flaked on it). Both rewritten as single retrying assertions ("zero non-matching elements").
- qase 208's "filtered count < baseline" oracle: the grid renders a fixed page, so both counts sit at the page cap (18 = 18). Dropped; the tag-exclusivity check is the real oracle. Removed the two now-unused count methods.
- `EarnPage.clearFilters()` clicked the dropdown-internal `clear-button`, which only exists while the popover is open — never reachable after `selectOptionFromDropdown` closes it. Now clicks the page-level `earn-filter-clear-filters-button`.
- qase 9's first-card click could hit a legacy `/quests/` entry whose slug doesn't resolve (develop Strapi draft data) — now scoped to canonical `/missions/` links.

## Validation

- `pnpm typecheck`, `pnpm tsc:tests`, eslint (changed files + tests) — clean
- `pnpm test:unit` — 357 passed; snapshots regenerated — diff is ONLY EarnCard/HeroEarnCard, attribute-only (tripwire for the `hintDataTestId` default held: no other .snap moved)
- Local e2e vs this branch (`--workers=1 --retries=0`): mainMenu + landingPage + mobileView + earnPage — all target tests green: qase 5, 9, 41, 42, 43, 208, plus 22 passing tests exercising the `welcome-screen` swap
- qase 22 note: it fails on a **local dev server** regardless of this PR (dev mode hard-errors on a develop-Strapi article image host missing from the `next/image` config; pre-existing `<h1>` assertion, not our swap line). CI is unaffected — the prod-build pipeline passes qase 22 (verified in dispatch run 29249692500).
- CI expectation: qase 41/42/43/208 EXECUTE (not skipped) in the shard report

Not CI-provable here (stay fixme'd on the funded-wallet prerequisite): the `perks-card-claimed-badge` swap (qase 200/201) and the qase 203 flow — verified as DOM presence on local dev only.

## Summary by CodeRabbit

- **New Features**
  - Added a configurable `data-testid` for the welcome overlay and expanded dedicated test identifiers across Learn, Missions, tool cards, and earn-card hints.

- **Bug Fixes**
  - Improved reliability of UI test targeting by standardizing card/tag test-id generation, adding missing identifiers for mission details/verification, and refining the perks “claimed” badge and hint test ids.

- **Tests**
  - Re-enabled previously skipped Earn-page filter scenarios and adjusted assertions to verify only matching results remain.
  - Updated Playwright flows to use the new test identifiers for missions navigation, wallet verification, and other page interactions.


yordanove:
@coderabbitai Adopted for the three src call sites in 032cca9e7 (`slugifyTestId` in `src/utils/`); the `EarnPage` copy stays local since `tests/` can't import from `src/` per the suite's boundary rule.

src/app/ui/missions/MissionsPage.tsx
```diff
@@ -32,7 +32,10 @@ export const MissionsPage = async () => {
     <>
       <BannerCampaign campaigns={validBannerCampaigns} />
       <MissionsSection count={totalMissions}>
-        <GridContainer gridTemplateColumns="repeat(auto-fill, minmax(min(320px, 100%), 1fr))">
+        <GridContainer
+          dataTestId="missions-list"
```
laurentsenta, line 36: To be consistent with other components like `MissionDetails` and `MissionDetailsSkeleton`: the component defines its own test id, rather that the caller, could you move this inside the `MissionsList` def?

src/app/ui/missions/MissionsPageSkeleton.tsx
```diff
@@ -23,7 +23,7 @@ export const MissionsPageSkeleton = () => {
               height: 32,
             }}
           />
-          <GridContainer>
+          <GridContainer dataTestId="missions-list">
```
laurentsenta, line 26: Same as above, for consistency I'd have each component define its own data test id

src/components/Cards/HeroEarnCard/HeroEarnCard.tsx
```diff
@@ -105,7 +106,7 @@ export const HeroEarnCard: FC<HeroEarnCardProps> = ({
               size={BadgeSize.SM}
               label={tag}
               key={tag}
-              data-testid={`earn-card-tag-${tag.toLowerCase().replace(/\s+/g, '-')}`}
+              data-testid={`earn-card-tag-${slugifyTestId(tag)}`}
```
laurentsenta, line 109: btw it's duplicating data, would it make sense to combine selectors instead?

```css
[data-testid="my-item"][label="some-tag"] { }
```

also fwiw the for you badge above doesn't get a test id

src/utils/slugifyTestId.ts
```diff
@@ -0,0 +1,2 @@
+export const slugifyTestId = (value: string): string =>
+  value.toLowerCase().replace(/\s+/g, '-');
```
laurentsenta, line 4: we have a `slugify` util I think

laurentsenta (approved):
Left a couple of comments, please double check before merging.

src/app/ui/missions/MissionsPage.tsx
yordanove: Makes sense in principle, but GridContainer is shared, so it can't hardcode the ID, and MissionsList's root is a Fragment - nothing to attach it to without adding a div that breaks the grid. Can move GridContainer inside MissionsList instead if you prefer; just say the word

src/app/ui/missions/MissionsPageSkeleton.tsx
yordanove: Same as above - plus MissionsListSkeleton doubles as the load-more loader, so if it owned the id we'd get duplicate ids mid-scroll

src/components/Cards/HeroEarnCard/HeroEarnCard.tsx
yordanove: label renders as text, not an attribute, so [label="…"] wouldn't match anything - we'd need a data-label attr, which is the same duplication in a different spot. I kept the pattern Compact/Hero already had; this PR just adds ListItem parity. The For You badge is icon-only and nothing tests it yet, so left it for whenever recommendations get e2e coverage

src/utils/slugifyTestId.ts
yordanove: I checked - urls/slugify isn't equivalent: it strips all non-alphanumerics, and ours matches what the cards already render on develop, so swapping would rename existing ids.  I'm happy to consolidate if we're ok with that; otherwise I'll add a short comment on the helper saying why it's separate

yordanove:
@laurentsenta , comments answered let me know if ok, or changes  still needed

laurentsenta:
@yordanove your feedbacks make sense, no push back on my side

PR #3031, opened 2026-07-14
chore: migrate limit orders to jumper-backend REST client

mmioana:
## Summary
- Replace the generated limit-order client/types (`gen-limit-order-api.sh`, `limitOrderClient.ts`, `jumper-limit-order.ts`) with the shared `jumper-backend` REST client and types, dropping the now-unused `NEXT_PUBLIC_LIMIT_ORDER_BACKEND_URL` env var.
- Fix a `chainId` mismatch causing `signTypedDataAsync` to throw `InvalidParamsRpcError` when cancelling an order while connected to a different chain than the order — now switches chains first if needed.
- Close (and reset) the repeat/modify order modals on wallet address change, matching the existing cancel order modal behavior.

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm build` passes
- [ ] Manually verify cancel-order signing works when connected wallet is on a different chain than the order
- [ ] Manually verify repeat/modify/cancel order modals all close on account switch


src/components/composite/CancelOrderFlow/hooks/useCancelOrder.ts
```diff
@@ -51,19 +52,21 @@ export interface CancelOrderResult {
...
+  const { account } = useAccount();
   const txFlow = useTransactionFlow();
 
   const [isAwaitingOnchainTx, setIsAwaitingOnchainTx] = useState(false);
   const [onchainChainId, setOnchainChainId] = useState<number>();
 
   const prepareAndSubmitFn = useCallback(
     async (order: Order): Promise<CancelOrderResult | undefined> => {
-      const client = makeLimitOrderClient();
+      const client = makeClient();
```
chmanie, line 65: 👍

chmanie (approved):
Nice. Can't find anything wrong with this

PR #3039, opened 2026-07-15
chore: use gRPC baseUrl for Sui client - JUM-1189

Kayanski:
Closes JUM-1189

## What

`SuiProvider` already used `SuiGrpcClient`, but still imported `getJsonRpcFullnodeUrl` from the deprecated `@mysten/sui/jsonRpc` to derive its `baseUrl`. This drops that import and hardcodes the gRPC mainnet endpoint (`https://fullnode.mainnet.sui.io:443`), completing the JSON-RPC removal on the frontend.

## Sister PR

- Backend: jumperexchange/jumper-backend#1032 (real client migration in `sui.adapter.ts`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Sui network connection to use the mainnet fullnode endpoint, improving consistency when connecting to the Sui mainnet.

### DIFF
--- a/src/providers/WalletProvider/SuiProvider.tsx
+++ b/src/providers/WalletProvider/SuiProvider.tsx
@@ -5,9 +5,8 @@ import {
   type DefaultExpectedDppKit,
 } from '@mysten/dapp-kit-react';
 import { SuiGrpcClient } from '@mysten/sui/grpc';
-import { getJsonRpcFullnodeUrl } from '@mysten/sui/jsonRpc';
-import { useHydrated } from '@/hooks/useHydrated';
 import type { FC, PropsWithChildren } from 'react';
+import { useHydrated } from '@/hooks/useHydrated';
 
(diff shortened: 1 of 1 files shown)


laurentsenta (approved):

src/providers/WalletProvider/SuiProvider.tsx
```diff
@@ -39,7 +38,7 @@ const createSuiDappKit = (options: {
     createClient: (network) =>
       new SuiGrpcClient({
         network,
-        baseUrl: getJsonRpcFullnodeUrl('mainnet'),
+        baseUrl: 'https://fullnode.mainnet.sui.io:443',
```
laurentsenta, line 41: I guess you already considered this, but just in case: if we need to change this URLs, likely we'll need to change it again, might make sense to move to an env (same on backend side)

PR #3070, opened 2026-07-20
fix: evaluate gatekeeper access for any connected wallet address

mmioana:
Move wallet-resolving helpers under accounts and allow non-EVM wallets through hasAdvanced checks without breaking the existing EVM reconnect wait.


yordanove:
Tested on the preview, works across the matrix: EVM wallet with the grant passes, reload keeps the gatekeeper waiting through the reconnect instead of redirecting, disconnected gets bounced, and the new part checks out both ways for Solana-only: without a grant it gets checked (the flags call goes out with the base58 address) and blocked, with a Wallet Access Control entry it passes. No regressions from my side.

yordanove (approved):

PR #3072, opened 2026-07-21
feat: default placeholder tokens on new widget (JUMADV-30)

mmioana:
## Summary
- Seeds default from/to placeholder pairs per surface via `resolveWidgetPlaceholderTokens` (Private / Gas stay empty):
  - Main Exchange (`default`): ETH USDC → SOL USDC
  - Advanced Swap: ETH → ETH USDC
  - Advanced Bridge: ETH USDC → SOL USDC
  - Advanced Limit: ETH USDC → ETH native
  - Private / Gas: empty
- Cold-load form seed priority per field: partner theme → props → session cache → URL → placeholders
- Empty-string tokens (e.g. from WidgetEvents clears) are treated as unset so they do not block placeholder fallback via `??`
- URL chain/token params are snapshotted once at mount so a sibling page’s `buildUrl` cannot re-pollute mid-lifecycle (`buildUrl` stays `true` on the main widget)
- Simple ↔ Advanced vertical nav: `prepareWidgetSurfaceNavigation` on click; destination Widget `consumeWidgetSurfaceNavigation` clears URL + cache before URL snapshot / form seed
- In-widget tab switches keep the widget mounted: write placeholders (or clear) with `setFieldValue`, clear session cache, and stop seeding chain/token through widget config after the first tab write so FormUpdater does not overwrite

Linear: [JUMADV-30](https://linear.app/jumper-exchange/issue/JUMADV-30)

## Test plan
- [ ] Cold load Main Exchange → ETH USDC → SOL USDC
- [ ] Cold load Advanced → Swap: ETH → ETH USDC; Bridge: ETH USDC → SOL USDC; Limit: ETH USDC → ETH native
- [ ] Private and Gas tabs stay empty (no ETH/USDC leftovers)
- [ ] Simple ↔ Advanced via vertical nav: no wipe/flicker; destination gets correct placeholders (not the other surface’s tokens)
- [ ] In-widget tab switches apply the correct pair (or clear) without remount flicker; navigation/UI state preserved
- [ ] Deep link `?fromChain=&fromToken=&toChain=&toToken=` still wins over placeholders on cold load
- [ ] Wallet balance card from-side selection still works after placeholders seed
- [ ] Announcement / partner links that set chain+token still win over placeholders
- [ ] Unit tests: `pnpm exec vitest run src/components/Widgets/hooks.spec.ts src/components/Widgets/variants/widgetConfig/utils.spec.ts --project unit`

## Summary by CodeRabbit

- **New Features**
  - Widget navigation now preserves selected chains and tokens when switching tabs.
  - Chain and token selections stay synchronized with the page URL for easier sharing and navigation.
  - Widget forms can update their values without unnecessary remounts or resets.
  - Empty states are handled consistently for specialized tabs.

- **Bug Fixes**
  - Prevented redundant navigation when selecting the already active tab.
  - Fixed empty token values being treated as valid selections.
  - Improved handling of invalid or incomplete URL parameters.

### DIFF
--- a/src/config/tokens.ts
+++ b/src/config/tokens.ts
@@ -3,6 +3,15 @@ import { ondoDenylist } from './generated/ondoDenylist';
 
 export const ARB_NATIVE_USDC = '0xaf88d065e77c8cc2239327c5edb3a432268e5831';
 
+/** Native ETH sentinel used by LI.FI / @jumperexchange/widget */
+export const ETH_NATIVE = '0x0000000000000000000000000000000000000000';
+
+/** Circle USDC on Ethereum */
+export const ETH_USDC = '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48';
+
+/** Circle USDC mint on Solana */
+export const SOL_USDC = 'EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v';
+
 export const tokens: WidgetTokens = {
   allow: [],
   deny: ondoDenylist,
--- a/src/components/Menus/VerticalMenu/useVerticalTabs.tsx
+++ b/src/components/Menus/VerticalMenu/useVerticalTabs.tsx
@@ -4,14 +4,15 @@ import { useRouter } from 'next/navigation';
 import { useTranslation } from 'react-i18next';
 import { Badge } from '@/components/Badge/Badge';
 import { BadgeSize, BadgeVariant } from '@/components/Badge/Badge.styles';
+import { prepareWidgetSurfaceNavigation } from '@/components/Widgets/variants/widgetConfig/utils';
 import { CandlestickChartIcon } from '@/components/illustrations/CandlestickChartIcon';
 import {
   TrackingAction,
   TrackingCategory,
   TrackingEventParameter,
 } from '@/const/trackingKeys';
-import { useUserTracking } from '@/hooks/userTracking/useUserTracking';
 import { useAdvancedAccess } from '@/hooks/useAdvancedAccess';
+import { useUserTracking } from '@/hooks/userTracking/useUserTracking';
 
 export const useVerticalTabs = () => {
   const { trackEvent } = useUserTracking();
@@ -21,6 +22,7 @@ export const useVerticalTabs = () => {
     useAdvancedAccess();
 
   const handleClickTab = (path: string, label: string) => () => {
+    prepareWidgetSurfaceNavigation();
     router.push(`/${path}`);
     trackEvent({
       category: TrackingCategory.Navigation,
--- a/src/components/Menus/VerticalMenu/useVerticalTabs.tsx
+++ b/src/components/Menus/VerticalMenu/useVerticalTabs.tsx
@@ -12,16 +12,29 @@ import {
   TrackingEventParameter,
 } from '@/const/trackingKeys';
 import { useAdvancedAccess } from '@/hooks/useAdvancedAccess';
+import { usePathnameWithoutLocale } from '@/hooks/routing/usePathnameWithoutLocale';
 import { useUserTracking } from '@/hooks/userTracking/useUserTracking';
+
+const normalizeVerticalTabPath = (path: string) => {
+  const withLeadingSlash = path.startsWith('/') ? path : `/${path}`;
+  return withLeadingSlash.replace(/\/+$/, '') || '/';
+};
 
 export const useVerticalTabs = () => {
   const { trackEvent } = useUserTracking();
   const router = useRouter();
+  const pathname = usePathnameWithoutLocale();
   const { t } = useTranslation();
   const { isEnabled: widgetAdvancedEnabled, isAllowed: advancedAllowed } =
     useAdvancedAccess();
 
   const handleClickTab = (path: string, label: string) => () => {
+    const targetPath = normalizeVerticalTabPath(path === '' ? '/' : path);
+    const currentPath = normalizeVerticalTabPath(pathname || '/');
+    if (currentPath === targetPath) {
+      return;
+    }
+
     prepareWidgetSurfaceNavigation();
     router.push(`/${path}`);
     trackEvent({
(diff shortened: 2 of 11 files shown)


src/config/tokens.ts
```diff
@@ -1,9 +1,87 @@
...
+    activeTabKey === 'refuel' ||
+    activeTabKey === 'private' ||
+    starterVariant === 'refuel' ||
+    starterVariant === 'private'
+  ) {
+    return undefined;
+  }
+
+  if (activeTabKey === 'swap-advanced' || activeTabKey === 'limit') {
+    return ethNativeToEthUsdc();
```
Kayanski: no, don't use ethNative as the input (fromToken) for limit orders, it's not supported by either of the limit order protocols we support, the other way around id fine

yordanove:
Tested on the preview, all the planned defaults match, two observations below.

Observation 1, intermittent: on some fresh sessions the placeholders never seed, the advanced tabs sit on Select/Select and the empty state sticks across tab switches until a reload. Hit it twice in roughly five fresh sessions (once arriving via the Gas tab toggle, once a single Limit miss after a cold /advanced load), then could not reproduce on demand, 3/3 stress runs seeded fine. Real but timing bound. Same machinery JUMEMB-18 talks about, this would be its new failure mode.

Observation 2, URL semantics changed, this is JUMEMB-17 territory: params are honored on load (tested two deeplinks, both won over the placeholders, priority order is right) but then get stripped from the address bar. A pair survives exactly one reload, a second reload falls back to placeholders, and copying the URL after load shares nothing. The tab itself now survives reload, which is the part of JUMEMB-17 this improves. Is the pair stripping intended with the session cache design? The cache did not restore the pair on the bare reload. If this is the direction, JUMEMB-17/18 probably want updating to match.

What passed: cold main Exchange USDC Ethereum to USDC Solana, Private and Gas stay empty, Advanced cold loads Swap ETH to USDC, Bridge USDC to Solana USDC, Limit ETH to USDC on direct load, and tab transitions keep each tab's own defaults with no cross-tab leaks once seeded.

Kayanski (approved):

src/config/tokens.ts
mmioana: You are right sir 😄

yordanove:
Retested on dd5383e5, the new limit pair is correct and quotes, but the fresh-session no-seed from my earlier test got worse on this head.

The issue: cold direct loads of /advanced now come up empty every time, not occasionally. Limit was Select/Select 3 out of 3, Bridge came up empty too, and Swap picked up only the buy chain with no token and stayed stuck like that. Switching tabs after load seeds correctly every time, and URL deeplinks still fill the form on a cold load, so this only hits the placeholder path on direct /advanced loads. This head also merged in the gatekeeper and widget 4.6.0, might be related.

What passes:
- Limit now seeds USDC on Ethereum selling and native ETH buying, and the pair quotes fine (1inch route, Review enabled)
- Swap ETH to USDC and Bridge USDC-Eth to USDC-Sol, both correct on tab switches
- Main tab seeds USDC-Eth to USDC-Sol on a cold load, Private and Gas stay empty
- Deeplinks still win over placeholders

yordanove:
Retested on d608394f, the cold no-seed is fixed. Every direct load that came up empty on the previous head now seeds correctly on fresh sessions: /advanced 2/2, ?tab=limit 3/3 (was 0/3), ?tab=bridge-advanced, main page 2/2. Private and Gas stay empty, arriving via the Gas tab seeds fine, tab hops keep each tab's own pair, and deeplinks win, including over an older cached pair.

Two new issues, both from this PR's code (nothing pre-existing, prod cold-loads with no defaults at all today):

**1. Switching a widget tab and then connecting a wallet empties the form.** Repro, 2 out of 2: cold load the main page with no wallet, switch Exchange to Gas and back, connect. Both fields drop to Select and the placeholders never come back. From the code: the first tab switch latches seedChainTokensFromConfig off for the rest of the visit, and connecting re-keys the feature-flags query, which remounts the widget with no seed source left (config latched off, nothing in the URL). Same family, from code reading only: a tab that disappears when its flag resolves disabled flips activeTabKey while the widget is still gated, so the placeholder write hits a null formRef but the latch still sets. Resetting the latch when the widget remounts, or skipping it while formRef.current is null, would cover both.
jumadv30-f1-latch-remount-empty-form


**2. Clicking the vertical tab you're already on wipes a picked pair.** Repro: /advanced, pick WETH as the sell token (URL shows fromToken=0xC02a…), click Advanced in the vertical nav again. Form and URL flip back to the ETH placeholder. prepareWidgetSurfaceNavigation runs on every click with no same-tab check (useVerticalTabs.tsx:25). One caveat: I fired that click programmatically, a physical click on the center of the active tab gets intercepted, so real-mouse reach may be narrower. The wipe itself is real.

Two small ones:
- /advanced?fromChain=abc&fromToken=xyz leaves Send on "Select": junk params block the placeholder fallback instead of being ignored (NaN passes the null checks in hooks.ts:75-80).
- On Limit, the Market Price panel still says "Select a token to see its market price" with the seeded pair, since it reads URL params and placeholders aren't written there. The panel being URL-driven predates this PR, the visible mismatch is just new. Flagging in case it matters.

URL behavior is nicer than before: params no longer get stripped, picked tokens are written to the URL on both Simple and Advanced, so a picked pair survives reloads and a copied URL shares it. Bare loads always give placeholders. If that's the intended design, JUMEMB-17/18 should be updated to match.

yordanove:
Retested on 1b9efede, both bugs are fixed and the smaller items too.

- Tab switch then connect (bug 1): 2/2 clean with the exact repro (fresh session, Exchange→Gas→back, connect Rabby). The pair stays seeded after the connect remount, where yesterday it dropped to Select/Select permanently.
- Same-tab vertical click (bug 2): clicking the already-active Advanced tab twice with a deeplinked pair on screen changes nothing, URL and pick both intact. Real switches still navigate fine (main→Advanced checked).
- Junk params: `?fromChain=abc&fromToken=xyz` now seeds the defaults and the URL gets rewritten to the seeded pair.
- Market Price on Limit now follows the seeded pair (chart live on a cold `?tab=limit` load, no more "Select a token").

Cold seeds still good on this head: main 3/3, /advanced swap, limit 2/2, deeplinks win, and the new URL mirror means the pair survives reloads and is shareable, which closes yesterday's one-reload oddity.

One observation, pre-existing rather than from this PR: a deeplinked pick still doesn't survive a widget tab round-trip (swap→limit→swap comes back with placeholders). Develop does the same today except it comes back empty, so this PR actually improves the landing state. JUMEMB-17/18 territory, no new ticket from me.

LGTM from QA.

yordanove (approved):
Retested on 1b9efede, both reported bugs verified fixed (details in my comment). Approving.

